PR #1, opened 2024-12-21
Develop target_features 1.1 prototype in main

raphlinus:
This branch contains a prototype of ideas that allow reasonably ergonomic, reasonably safe use of SIMD, which will work on stable as soon as target_feature 1.1 is stabilized (hopefully soon). There are other proposals, hopefully at least one better than this one. A major purpose in pushing this is so that these proposals can be compared empirically.

In the meantime, the tradeoffs and compromises made here are fairly well suited to developing SIMD color conversion (either in color or in a sibling crate), fast geometry, and cpu sparse strip rendering.

In the current state, it's not very usable - it only has a fraction of the types implemented, support for the different levels is spotty. Developing in main will let us do more experiments, and also open the door to potential collaboration.


examples/fast_detect.rs
```diff
@@ -0,0 +1,28 @@
...
+    use simd::f32_4::*;
+    let x = inp.into();
+    mul_add(x, x, splat(1.0)).into()
+}
+
+fn main() {
+    // Safety: doing this before launching any threads that might use SIMD.
+    unsafe {
+        fearless_simd::init_simd_detect();
+    }
```
lqd, line 25: worst case you wrap the feature detection in another entry point proc macro, à la 

```rust
#[fearless_simd::main]
fn main() {
    let inp = [0.0, 1.0, 2.0, 3.0];
    println!("{:?}", foo(inp));
}
```

raphlinus: That opens a potential safety hole, nothing stops you from spawning a thread and applying that macro to a function run by the thread.

raphlinus:
I don't think this is the way forward, #2 is better, so I'm closing.

PR #2, opened 2024-12-27
Experiment with token-based API

raphlinus:
This is another exploration, largely to see what the API would look like. In this one, the vector types look like `f32x4<S: Simd>`, where `Simd` is a trait impl'd by zero size tokens that witness the CPU capabilities. The vector types are 2-element struct with the corresponding `[f32; 4]` array and the token. Like the original fearless_simd, because they witness feature detection, std::ops can be implemented safely on them.

This design is actually quite a bit more similar to the original fearless_simd idea, and also borrows lots from pulp, including pretty much the entire safe delegation of intrinsics. As far as I can tell, this basically removes the need for target_feature 1.1.

I think one thing that works well is the `as_neon()` method on the level enum, which can also be gotten from the token. This is a way to get downcasting, so you can write code completely generic on the Simd level (this is the main motor for multiversioning) while also 

One downside of the vector-with-witness approach is that you can't do safe transmute on it (ie they can't implement the bytemuck traits), otherwise that would let you synthesize a token not actually supported by the CPU.

Detection is safe. For performance, the preferred pattern is to mint a level enum at a high level (for example, when instantiating a renderer), then pass it around. When inlined, the dispatch should go away. When not inlined, it should be a cheap conditional branch. This also solves the question of how to take the address of a multiversioned function - take the address of the non-generic version that take the level enum, rather than one that's polymorphic in `<S: Simd>`.

As of the first draft, very few types and operations are implemented, not much more than needed to run the srgb example.


raphlinus:
I marked this as ready for review, as I'm inclined to merge it. I think it's enough to serve as a discussion point for what SIMD should look like, and I'm not likely to do significant more work at the moment.

README.md
```diff
@@ -1,85 +1,37 @@
...
-Errors (including in judgment for going down this path) are my own, but I've benefitted from discussions with many people, including with James McCartney, Andrew Gallant ([burntsushi](https://github.com/BurntSushi)), [talchas](https://github.com/talchas), Colin Rofls, and Alex Crichton.
-
-[faster]: https://github.com/AdamNiederer/faster
-
-[packed_simd]: https://github.com/rust-lang-nursery/packed_simd
+[pulp]: https://crates.io/crates/pulp
+[target_feature 1.1]: https://github.com/rust-lang/rfcs/pull/2396
+[Towards fearless SIMD]: https://raphlinus.github.io/rust/simd/2018/10/19/fearless-simd.html
+[fearless_simd 0.1.1]: https://crates.io/crates/fearless_simd/0.1.1
+[half]: https://crates.io/crates/pulp
```
tomcur, line 33:
```suggestion
[half]: https://crates.io/crates/half
```

src/half_assed.rs
```diff
@@ -0,0 +1,140 @@
...
+        f16(bits)
+    }
+
+    /// Constructs a 16-bit floating point value from a 32-bit floating point value.
+    ///
+    /// This function is identical to [`from_f32`][Self::from_f32] except it never uses hardware
+    /// intrinsics, which allows it to be `const`. [`from_f32`][Self::from_f32] should be preferred
+    /// in any non-`const` context.
+    ///
+    /// This operation is lossy. If the 32-bit value is to large to fit in 16-bits, ±∞ will result.
```
tomcur:
```suggestion
    /// This operation is lossy. If the 32-bit value is too large to fit in 16-bits, ±∞ will result.
```

README.md
tomcur:
```suggestion
[half]: https://crates.io/crates/half
```

src/fallback.rs
```diff
@@ -1,74 +1,259 @@
-//! Fallback implementations of the SIMD traits.
+// Copyright 2024 the Fearless_SIMD Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use std::ptr;
+//! Fallback implementations of SIMD methods.
 
-use traits::{SimdF32, SimdMask32};
+// Discussion question: what's the natural width? Most other implementations
+// implementations default to 1, but perhaps we want to express the idea
```
tomcur:
```suggestion
// Discussion question: what's the natural width? Most other implementations
// default to 1, but perhaps we want to express the idea
```

src/x86_64/avx2.rs
```diff
@@ -0,0 +1,178 @@
...
+};
+use core::arch::x86_64::*;
+
+use super::Level;
+
+/// The SIMD token for the "avx2" level.
+///
+/// This is short for the "x86-64-v3 microarchitecture level". In this level, the
+/// following target_features are enabled: "avx2", "bmi2", "f16c", "fma", "lzcnt".
+
```
tomcur: The docs are here twice.

```suggestion
```

src/x86_64/avx2.rs
```diff
@@ -0,0 +1,178 @@
...
+/// The SIMD token for the "avx2" level.
+///
+/// This is short for the "x86-64-v3 microarchitecture level". In this level, the
+/// following target_features are enabled: "avx2", "bmi2", "f16c", "fma", "lzcnt".
```
tomcur: Is authoritative documentation available somewhere of CPU feature hierarchies?

tomcur (commented):
I've taken a quick glance; I like the token-based approach in general (as you know). 

Something that may or may not be nice (I haven't worked out the ergonomics yet), is that it will be possible to encode the features at the type-level once `adt_const_params` lands – I believe it is close to stabilization. The token type could then be const-generic over a struct with boolean fields, one for each feature, with specializations for specific blessed levels.

src/x86_64/avx2.rs
raphlinus: That's a *good* question, and when you look at other implementations you see immense confusion and potential bugs that will impact fringe x86 clones (I linked at least one such bug from my [irlo thread](https://internals.rust-lang.org/t/better-codegen-for-cpu-feature-detection/22083)).

I'm generally going off [Wikipedia](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) for the v3 etc microarchitecture levels. In terms of the graph of implied features, my main source is [rustc](https://github.com/rust-lang/rust/blob/52890e82153cd8716d97a96f47fb6ac99dec65be/compiler/rustc_target/src/target_features.rs#L207).

Probably this should be captured in a wiki; there's a ton of lore and I'm not aware of any great authoritative sources.

Ralith:
Is this blocked on further review? I'm excited to see new work in this space, given how `std::simd` is bogged down and other safe abstractions seem very narrowly scoped.

taj-p:
Hey @Ralith - see [#rust > A plan for SIMD @ 💬](https://xi.zulipchat.com/#narrow/channel/255911-rust/topic/A.20plan.20for.20SIMD/near/520769933)

Cargo.toml
```diff
@@ -1,11 +1,14 @@
 [package]
 name = "fearless_simd"
-version = "0.1.1"
+version = "0.2.0"
 license = "MIT/Apache-2.0"
+edition = "2021"
```
ajakubowicz-canva, line 5: Nit: Edition 2024? If it breaks stuff can punt and I can update in a followup.

Cargo.toml
```diff
@@ -1,11 +1,14 @@
...
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 keywords = ["simd"]
 categories = ["hardware-support"]
 description = "Safer and easier SIMD"
 readme = "README.md"
+rust-version = "1.83"
 
 [dependencies]
+half = { version = "2.4.1", optional = true }
```
ajakubowicz-canva: Nit: Could maybe update to `2.6`

README.md
```diff
@@ -1,85 +1,37 @@
 # Towards "fearless SIMD"
 
-This crate provides safe wrappers to make it easier to write SIMD code. It doesn't yet deliver on the promise of "fearless SIMD", but shows a potential path towards it.
+This crate proposes an experimental way to use SIMD intrinsics reasonably safely, using the new [target_feature 1.1] feature in Rust. It is currently nightly-only until that feature is stabilized, but that is expected to happen soon.
```
ajakubowicz-canva: Rust 1.86 stabilized target_feature: https://blog.rust-lang.org/2025/04/03/Rust-1.86.0/#allow-safe-functions-to-be-marked-with-the-target-feature-attribute

README.md
```diff
@@ -1,85 +1,37 @@
...
+Some code has been cut and pasted from the [half] crate, which is released under identical license. That crate is also an optional dependency, for more full f16 support.
 
-* No `unsafe` required to use.
+## SIMD types
 
-* Access to advanced SIMD primitives such as rounding and approximate reciprocal.
+The SIMD types in this crate are a thin newtype around the corresponding array, for example `f32x4` is a newtype for `[f32; 4]`. These types are in the crate root and have a number of methods, including loading and storing, that do not require intrinsics. They are not aligned, as modern CPUs support unaligned access efficiently.
 
-* Works with stable Rust.
+There are
```
ajakubowicz-canva: Nit: Incomplete sentence.

ajakubowicz-canva (approved):
LGTM as we've discussed the direction of this PR and are happy to merge and move forward with this work.

Cargo.toml
raphlinus: It forces a big diff, as there's a lot of `unsafe unsafe` stuttering in 2024. I think it should be done in a followup either as part of the autogen PR or shortly after.

ajakubowicz-canva: I am happy with whatever is easiest. I'm happy to volunteer to do the migration once it won't create merge conflicts or churn.

PR #3, opened 2025-06-06
Move toward autogeneration of types and methods

raphlinus:
This moves forward with the plan for SIMD, autogenerating the bulk of the types and the implementation.

There are some limitations:

* No documentation
* Neon only
* No fp16 yet (there are choices how best to do this)
* Missing mul_add
* Many other missing SIMD methods


fearless_simd_gen/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+            Module::SimdTrait => mk_simd_trait::mk_simd_trait(),
+            Module::Ops => mk_ops::mk_ops(),
+            Module::Neon => mk_neon::mk_neon_impl(mk_neon::Level::Neon),
+        }
+    }
+
+    fn generate_string(self) -> String {
+        let code = self.generate_code();
+        let parsed = syn::parse_file(&code.to_string()).expect("error parsing {name}");
+        let code_str = prettyplease::unparse(&parsed);
```
Ralith, line 45: Would it be simpler/more robust to delegate to the real `rustfmt`? That approach has worked well for us [in ash](https://github.com/ash-rs/ash/blob/55dc692946f52f0dff326bd13b1722d98f52fd2b/generator/src/lib.rs#L3516-L3526), and will likely produce more idiomatic results. That should also enable dropping `syn` entirely for faster iteration.

raphlinus: I'm completely open to this, and it sounds like ash is well worth studying as an example.

PR #6, opened 2025-06-16
More improvements for `fearless_simd`

LaurenzV:
This PR combines three changes (each separated by a single commit):
- Adds the `min`, `max` (as well as precise methods for those), `madd` and `floor` methods. This is enough to create a SIMD version of strip rendering.
- Adds support for a `Fallback` backend, using scalar operations.
- Reformats `fearless_simd_gen` (we probably should add this to CI sooner than later).


fearless_simd/src/core_arch/fallback.rs
```diff
@@ -0,0 +1,13 @@
+/// A token for fallback SIMD.
```
ajakubowicz-canva, line 4:
```suggestion
// Copyright 2024 the Fearless_SIMD Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT

/// A token for fallback SIMD.
```

fearless_simd_gen/src/arch/fallback.rs
```diff
@@ -0,0 +1,64 @@
+use crate::arch::Arch;
```
ajakubowicz-canva, line 4:
```suggestion
// Copyright 2024 the Fearless_SIMD Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT

use crate::arch::Arch;
```

fearless_simd_gen/src/arch/neon.rs
```diff
@@ -1,23 +1,15 @@
-// Copyright 2025 the Fearless_SIMD Authors
-// SPDX-License-Identifier: Apache-2.0 OR MIT
-
-use proc_macro2::{Span, TokenStream};
-use quote::quote;
-use syn::Ident;
-
+use crate::arch::Arch;
```
ajakubowicz-canva, line 4:
```suggestion
// Copyright 2024 the Fearless_SIMD Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT

use crate::arch::Arch;
```

fearless_simd_gen/src/mk_fallback.rs
```diff
@@ -0,0 +1,305 @@
+use crate::arch::fallback::Fallback;
```
ajakubowicz-canva, line 4:
```suggestion
// Copyright 2024 the Fearless_SIMD Authors
// SPDX-License-Identifier: Apache-2.0 OR MIT

use crate::arch::fallback::Fallback;
```

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -38,6 +38,8 @@ pub fn mk_simd_types() -> TokenStream {
             quote! {}
         };
         let impl_block = simd_impl(ty);
+        let simd_from_items =make_list( (0..ty.len).map(|idx| quote! { val[#idx] })
```
ajakubowicz-canva, line 41: nit: run a `fmt`.

fearless_simd_gen/src/arch/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+pub(crate) mod fallback;
+pub(crate) mod neon;
+
+use proc_macro2::TokenStream;
+
+use crate::types::VecType;
+
+pub trait Arch {
```
ajakubowicz-canva: Is this trait intended to be public?

Nit: `pub(crate)`

fearless_simd_gen/src/mk_fallback.rs
```diff
@@ -0,0 +1,305 @@
...
+
+                    if method == "madd" {
+                        // TODO: This is has slightly different semantics than a fused multiply-add,
+                        // since we are not actually fusing it, should this be documented?
+                        quote! {
+                            #[inline(always)]
+                            fn #method_ident(self, a: #ty<Self>, b: #ty<Self>, c: #ty<Self>) -> #ret_ty {
+                               a.add(b.mul(c))
+                            }
+                        }
```
ajakubowicz-canva, line 137: There's probably a nicer pattern for these methods that need special handling rather than inline `if` statements. But happy to leave as-is since we're very early days.

fearless_simd_gen/src/mk_fallback.rs
```diff
@@ -0,0 +1,305 @@
...
+                            #items.simd_into(self)
+                        }
+                    }
+                }
+                OpSig::Ternary => {
+                    let args = [
+                        quote! { a.into() },
+                        quote! { b.into() },
+                        quote! { c.into() },
+                    ];
```
ajakubowicz-canva: Nit: args could be moved into the `else` block to be closer to usage.

fearless_simd/src/generated/fallback.rs
```diff
@@ -0,0 +1,3484 @@
...
+                b[0usize * 2 + 1],
+                a[1usize * 2 + 1],
+                b[1usize * 2 + 1],
+                a[2usize * 2 + 1],
+                b[2usize * 2 + 1],
+                a[3usize * 2 + 1],
+                b[3usize * 2 + 1],
+            ]
+                .simd_into(self),
+        )
```
ajakubowicz-canva: It'll be excellent when we have tests, since this PR is a little too big for me to review thoroughly.

I think this implementation is wrong and other `unzip` implementations may suffer a similar issue.

Using `neon` as a source of truth (copied into this comment for ease):

```rs
   #[inline(always)]
    fn unzip_i16x8(self, a: i16x8<Self>, b: i16x8<Self>) -> (i16x8<Self>, i16x8<Self>) {
        let x = a.into();
        let y = b.into();
        unsafe { (vuzp1q_s16(x, y).simd_into(self), vuzp2q_s16(x, y).simd_into(self)) }
    }
```

If we just look at the lower half command `vuzp1q_s16` ([docs](https://developer.arm.com/architectures/instruction-sets/intrinsics/vuzp1q_s16)), this command:

> This instruction reads corresponding even-numbered vector elements from the two source SIMD&FP registers, starting at zero, places the result from the first source register into consecutive elements in the lower half of a vector, and the result from the second source register into consecutive elements in the upper half of a vector

So I believe the lower half should be `[a0, a2, a4, a6, b0, b2, b4, b6]` instead of `a` and `b` being interleaved.

ajakubowicz-canva (commented):
This is looking quite good! I still have a bit to review but posting initial comments.

fearless_simd_gen/src/mk_simd_types.rs
LaurenzV: I didn’t do this yet because the previous version wasn’t formatted either (so there would be more changes), and the generated version also has different formatting than rustfmt. So maybe better to leave this for a separate PR that also adds formatting to CI?

fearless_simd_gen/src/mk_fallback.rs
LaurenzV: Indeed, and I just copied this from the current neon implementation.

fearless_simd/src/generated/fallback.rs
LaurenzV: Thanks, will take a look. And yes, I agree that tests should be a priority soon.

fearless_simd_gen/src/arch/mod.rs
LaurenzV: I think not? It was public before as well that's why I didn't touch it, but you are probably right that it shouldn't be public. Will change it for this instance, but there are probably many other instances which I think we should leave for later, since the focus for now is on functionality.

fearless_simd/src/generated/fallback.rs
LaurenzV: I manually tested it now and it seems to yield the same results as the neon implementation, so hopefully it should be good now.

fearless_simd_gen/src/mk_simd_types.rs
ajakubowicz-canva: Ah yeah great point!

fearless_simd_gen/src/mk_fallback.rs
ajakubowicz-canva: omg. I used 2024 instead of 2025. hahaha! I'm still living in the past!

PR #7, opened 2025-06-17
Make `fearless_simd` no_std compatible and add basic CI

LaurenzV:


LaurenzV:
I've also consolidated everything into a single `Level` struct instead of having a different one for each target, which will hopefully also make it easier to add WASM without having to duplicate code.

fearless_simd/src/lib.rs
```diff
@@ -18,21 +19,82 @@ pub use traits::*;
...
+
+    #[cfg(all(feature = "std", target_arch = "x86_64"))]
+    pub fn new() -> Self {
+        Self::fallback()
+    }
+    
+    #[cfg(any(not(feature = "std"), not(any(target_arch = "aarch64", target_arch = "x86_64"))))]
+    pub fn new() -> Self {
+        Self::fallback()
+    }
```
ajakubowicz-canva, line 63: Can these be simplified to be the inverse of `#[cfg(all(feature = "std", target_arch = "aarch64"))]`?
E.g.

```suggestion
    #[cfg(not(all(feature = "std", target_arch = "aarch64")))]
    pub fn new() -> Self {
        Self::fallback()
    }
```

fearless_simd/src/lib.rs
```diff
@@ -18,21 +19,82 @@ pub use traits::*;
...
+
+    #[cfg(all(feature = "std", target_arch = "aarch64"))]
+    #[inline]
+    pub fn as_neon(self) -> Option<Neon> {
+        match self {
+            Level::Neon(neon) => Some(neon),
+            _ => None,
+        }
+    }
```
ajakubowicz-canva, line 72: I'm interested in your thoughts. My rough impression was that `fearless_simd` APIs such as `as_neon` act as a witness and thus are a helper for consumer libraries. E.g. in the `examples/play.rs`:

```rs
fn do_something_on_neon(_level: Level) -> f32 {
    #[cfg(target_arch = "aarch64")]
    if let Some(neon) = _level.as_neon() {
        return neon.vectorize(
            #[inline(always)]
            || {
                let v = neon.neon.vdupq_n_f32(42.0);
                neon.neon.vgetq_lane_f32::<0>(v)
            },
        );
    }
    0.0
}
```

Thus, my main question is, is it a better API for the method `as_neon` to completely vanish given `cfg`, or, have the method always present, but make it unconditionally return `None` if `not(all(feature = "std", target_arch = "aarch64"))`, maybe something like:

```suggestion
    #[inline]
    pub fn as_neon(self) -> Option<Neon> {
        match self {
            #[cfg(all(feature = "std", target_arch = "aarch64"))]
            Level::Neon(neon) => Some(neon),
            _ => None,
        }
    }
````

File: .github/workflows/ci.yml [hunk 3]
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          toolchain: ${{ env.RUST_STABLE_VER }}
+
+      - name: restore cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name != 'merge_group' }}
+
+      - name: cargo test
+        run: cargo test run --workspace --locked --all-features --no-fail-fast
```
ajakubowicz-canva, line 51:
```suggestion
        run: cargo test --workspace --locked --all-features --no-fail-fast
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+    
+  build-no-std:
+    name: build (no_std)
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
ajakubowicz-canva, line 57: This can be removed as `fail-fast` applies to a `matrix`. ([ref](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#handling-failures))

LaurenzV:
(Will address the above comments tomorrow, merging now so that Andrew can merge his WASM implementation now.)

.github/workflows/ci.yml
ajakubowicz-canva: Resolved in my PR.

fearless_simd/src/lib.rs
ajakubowicz-canva: I've changed my mind.

fearless_simd/src/lib.rs
ajakubowicz-canva: I've simplified in the wasm architecture PR.

PR #8, opened 2025-06-18
Add initial WASM SIMD128 support

ajakubowicz-canva:
> [!NOTE]
> The measurements in this PR were done very rough via a quick Chrome profile. Because the numbers are small, until we run a rigorous benchmark it'll be hard to validate the impact. Especially across browser engines, and with JS engines optimising WASM and fusing operators.

## Overview

This PR adds initial WASM SIMD support to `fearless_simd`, implementing enough operations to enable WASM SIMD in https://github.com/linebender/vello/pull/1053. Rather than implementing all operations in one large PR, this focuses on the essential subset needed for Vello and breaks up an otherwise huge change. There's also some tricky operations to implement using the small amount of WASM instructions 😅 .

## Performance Impact

Tested with Ghost Tiger rendering in Vello:

### Without fast kurbo (baseline):
- **Without SIMD**: ~42.31ms per frame
- **With SIMD**: ~39.91ms per frame  
- **Improvement**: ~5% faster

### With fast kurbo (https://github.com/linebender/kurbo/pull/427):
- **Without SIMD**: ~6ms per frame
- **With SIMD**: ~4ms per frame
- **Improvement**: ~30% faster (Take this with a huge grain of salt)

*Test methodology: I linked `vello` locally to `fearless_simd` via `path` reference, and modified `vello_hybrid` to use the WASM SIMD level.*

## Changes

- **New architecture**: Added WASM SIMD128 support
- **Operations implemented**: Core subset including:
  - Binary ops: `add`, `sub`, `mul`, `min`, `max`
  - Comparison ops: `simd_eq`, `simd_ne`, `simd_lt`, etc.
  - Math ops: `sqrt`, `madd`
- **Testing**: Added parity tests ensuring Fallback and WASM SIMD produce identical results
- **Bug fix**: Fixed incorrect mask generation in Fallback comparison operations (was returning `0/1` instead of `0/-1`)

## Test Plan

Added `test_wasm_simd_parity!` macro that verifies operations produce identical results across Fallback and WASM SIMD implementations.

I only tested a small subset. Maybe in the future we code-gen the tests as well?

## Next Steps

Future PRs will add more operations to achieve full WASM SIMD coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,8 +53,6 @@ jobs:
   build-no-std:
     name: build (no_std)
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,33 @@ jobs:
 
       - name: cargo test
         run: cargo test run --workspace --locked --all-features --no-fail-fast
-    
+
+  test-stable-wasm:
+    name: test (wasm32)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: install stable toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_STABLE_VER }}
+          targets: wasm32-unknown-unknown
+
+      - name: restore cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name != 'merge_group' }}
+
+      - name: install wasm-pack
+        uses: taiki-e/install-action@v2
+        with:
+          tool: wasm-pack
+
+      - name: Run fearless_simd_tests on Chrome
+        run: wasm-pack test --headless --chrome
+        working-directory: fearless_simd_tests
+  
   build-no-std:
     name: build (no_std)
     runs-on: ubuntu-latest
--- a/fearless_simd_gen/src/arch/wasm.rs
+++ b/fearless_simd_gen/src/arch/wasm.rs
@@ -1,0 +1,93 @@
+// Copyright 2025 the Fearless_SIMD Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::arch::Arch;
+use crate::types::{ScalarType, VecType};
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::quote;
+
+pub struct Wasm;
+
+fn translate_op(op: &str) -> Option<&'static str> {
+    Some(match op {
+        "abs" => "abs",
+        "neg" => "neg",
+        "floor" => "floor",
+        "sqrt" => "sqrt",
+        "add" => "add", // There is also a add_sat
+        "sub" => "sub", // There is also a sub_sat
+        "mul" => "mul",
+        "div" => "div",
+        "simd_eq" => "eq",
+        "simd_lt" => "lt",
+        "simd_le" => "le",
+        "simd_ge" => "ge",
+        "simd_gt" => "gt",
+        "not" => "not",
+        "and" => "and",
+        "or" => "or",
+        "xor" => "xor",
+        // TODO: With target-feature "relaxed-simd" this could be "relaxed_max".
+        "max" => "max",
+        // TODO: With target-feature "relaxed-simd" this could be "relaxed_min".
+        "min" => "min",
+        "max_precise" => "max",
+        "min_precise" => "min",
+        "splat" => "splat",
+        // TODO: Only target-feature "relaxed-simd" has "relaxed_madd".
+        _ => return None,
+    })
+}
+
+fn simple_intrinsic(name: &str, ty: &VecType) -> TokenStream {
+    let ty_prefix = Wasm.arch_ty(ty);
+    let ident = Ident::new(name, Span::call_site());
+    let combined_ident = Ident::new(
+        &format!("{}_{}", ty_prefix.to_string(), ident.to_string()),
+        Span::call_site(),
+    );
+    quote! { #combined_ident }
+}
+
+fn v128_intrinsic(name: &str) -> TokenStream {
+    let ty_prefix = Ident::new("v128", Span::call_site());
+    let ident = Ident::new(name, Span::call_site());
+    let combined_ident = Ident::new(
+        &format!("{}_{}", ty_prefix.to_string(), ident.to_string()),
+        Span::call_site(),
+    );
+    quote! { #combined_ident }
+}
+
+impl Arch for Wasm {
+    fn arch_ty(&self, ty: &VecType) -> TokenStream {
+        let scalar = match ty.scalar {
+            ScalarType::Float => "f",
+            ScalarType::Unsigned => "u",
+            ScalarType::Int | ScalarType::Mask => "i",
+        };
+        let name = format!("{}{}x{}", scalar, ty.scalar_bits, ty.len);
+        let ident = Ident::new(&name, Span::call_site());
+        quote! { #ident }
+    }
+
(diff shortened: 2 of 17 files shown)


fearless_simd_tests/tests/wasm.rs
```diff
@@ -0,0 +1,226 @@
...
+            let wasm_result = dispatch(Level::WasmSimd128(wasm32::WasmSimd128::new_unchecked()));
+
+            assert_eq!(fallback_result, wasm_result,
+                concat!(stringify!($test_name), ": Expected fallback and WASM SIMD results to match."));
+        }
+    };
+}
+
+test_wasm_simd_parity! {
+    fn test_add_f32x4() {
```
DJMcNab: Nit: `test_` prefix on tests is not necessary/idiomatic, at least outside of `wasm_bindgen_test`

ajakubowicz-canva: Thank you! Fixed in https://github.com/raphlinus/fearless_simd/pull/8/commits/230270f1092530922f7931860b6dd9851e3d9da5

fearless_simd_tests/tests/wasm.rs
LaurenzV, line 1: I agree that, in the long term, we will probably want to autogenerate the tests, so that you write a test once with the expected output, and then compare it across all implementations. But, happy to land this as an intermediate version so we have at least some test coverage.

.github/workflows/ci.yml
LaurenzV, line 1: Will the WASM tests you added run in CI (I think not)? If it's on purpose feel free to ignore, just wanted to point it out! ^^

fearless_simd_tests/tests/wasm.rs
```diff
@@ -0,0 +1,226 @@
+// Copyright 2025 the Fearless_SIMD Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#[cfg(target_arch = "wasm32")]
+use fearless_simd::*;
+#[cfg(target_arch = "wasm32")]
+use wasm_bindgen_test::*;
+
+/// `test_wasm_simd_parity` enforces that the fallback level and +simd128 levels output the same
```
LaurenzV, line 9: Ideally the reference result should be provided manually (since it's possible the fallback is wrong as well), but that's for the future when we implement a proper test suite

fearless_simd_gen/src/arch/wasm.rs
```diff
@@ -0,0 +1,93 @@
...
+        "or" => "or",
+        "xor" => "xor",
+        // TODO: With target-feature "relaxed-simd" this could be "relaxed_max".
+        "max" => "max",
+        // TODO: With target-feature "relaxed-simd" this could be "relaxed_min".
+        "min" => "min",
+        "max_precise" => "max",
+        "min_precise" => "min",
+        "splat" => "splat",
+        // TODO: Only target-feature "relaxed-simd" has "relaxed_madd".
```
LaurenzV, line 37: Would be interesting to see if there are performance improvements if you use relaxed_madd. For me, that was the main thing that made a difference between the autovectorized version and the manual SIMD implementation, the speedup was like 15% there.

fearless_simd_gen/src/arch/wasm.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct Wasm;
+
+fn translate_op(op: &str) -> Option<&'static str> {
+    Some(match op {
+        "abs" => "abs",
+        "neg" => "neg",
+        "floor" => "floor",
+        "sqrt" => "sqrt",
+        "add" => "add", // There is also a add_sat
+        "sub" => "sub", // There is also a sub_sat
```
LaurenzV: `add_sat` and `sub_sat` will be separate instructions

fearless_simd_gen/src/mk_fallback.rs
```diff
@@ -173,11 +173,11 @@ fn mk_simd_impl() -> TokenStream {
                     let mask_type = VecType::new(ScalarType::Mask, vec_ty.scalar_bits, vec_ty.len);
                     let items = make_list(
                         (0..vec_ty.len)
-                            .map(|idx| {
+                            .map(|idx: usize| {
                                 let args = [quote! { &a[#idx] }, quote! { &b[#idx] }];
                                 let expr = Fallback.expr(method, vec_ty, &args);
                                 let mask_ty = mask_type.scalar.rust(scalar_bits);
-                                quote! { #expr as #mask_ty }
+                                quote! { -(#expr as #mask_ty) }
```
LaurenzV, line 180: Oh I see now, this is because in neon, the mask for true is all bits set to 1? I assumed it doesn't matter what the representation for true is as long as `false` is, but I guess it makes sense to keep it consistent. Do you know if all SIMD variants are guaranteed to have that representation?

fearless_simd_gen/src/mk_wasm.rs
```diff
@@ -0,0 +1,310 @@
...
+                                == (&VecType {
+                                    scalar: ScalarType::Int,
+                                    scalar_bits: 8,
+                                    len: 16,
+                                }))
+                    {
+                        quote! {
+                            #[inline(always)]
+                            fn #method_ident(self, a: #ty<Self>, b: #ty<Self>) -> #ret_ty {
+                                // TODO: WASM doesn't have `i8x16_mul` or `u8x16_mul`.
```
LaurenzV, line 122: So how do you multiply 8-bit integers? 🤔

LaurenzV (approved):

.github/workflows/ci.yml
ajakubowicz-canva: Great point – I think I've added the action correctly in https://github.com/raphlinus/fearless_simd/pull/8/commits/732454e275f7c0f444561a5c2971db9022cb2aea

fearless_simd_tests/tests/wasm.rs
ajakubowicz-canva: Thank you!

I tried a couple different macro approaches, but it ended up being extremely confusing. Maybe there is a way to express the tests as a nice macro as well. I am also not very confident with macros.

ajakubowicz-canva: I also don't yet have a good idea of how to test `neon`. Maybe the CI mac already supports it?

fearless_simd_gen/src/arch/wasm.rs
ajakubowicz-canva: That's super interesting. I didn't get enough time to explore yet but it's definitely something I'm also curious about.

fearless_simd_gen/src/arch/wasm.rs
ajakubowicz-canva: Ah excellent. I'll remove the comments. https://github.com/raphlinus/fearless_simd/pull/8/commits/96eed42712980a0f0a1913f49d26e007d7560621

fearless_simd_gen/src/mk_fallback.rs
ajakubowicz-canva: From a quick look, SSE/AVX (x86) and neon both set all bits to 1 for true, and all to 0 for false. So this change will be consistent with them.

E.g. Neon, x86, fallback, and Wasm should all be identical.

fearless_simd_gen/src/mk_wasm.rs
ajakubowicz-canva: I thought the same thing, and was confused why those types don't exist! 

Luckily, [Highway has an implementation](https://github.com/google/highway/blob/9d1fa11ad921b97357aed0cb277ded70ad9170f8/hwy/ops/wasm_128-inl.h#L950-L958) that can be used as a reference.

fearless_simd_tests/tests/wasm.rs
DJMcNab: The CI Macs should support this, because they're physical M1 machines. In theory, you should be running the same code with different level enum values, I think.

PR #12, opened 2025-06-20
[WASM] implement `mul_u8x16` and `mul_i8x16`

ajakubowicz-canva:
`mul_u8x16` and `mul_i8x16` are implemented as truncating multiplication, matching `neon`.
While unit testing this change I noticed that the Fallback level panics on overflow. Opened: https://github.com/raphlinus/fearless_simd/issues/11

A source of inspiration for this implementation was Highway's [MulHigh](https://github.com/google/highway/blob/9d1fa11ad921b97357aed0cb277ded70ad9170f8/hwy/ops/wasm_128-inl.h#L952-L957).

### DIFF
--- a/fearless_simd_tests/tests/wasm.rs
+++ b/fearless_simd_tests/tests/wasm.rs
@@ -61,6 +61,26 @@ test_wasm_simd_parity! {
 }
 
 test_wasm_simd_parity! {
+    fn mul_u8x16() {
+        |s| -> [u8; 16] {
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):
Haven't looked into how `shuffle` works, but seems fine!

ajakubowicz-canva:
I've added back tests, but kept the tests within bounds as to not exercise overflow logic.
I won't land this until the larger #14 has landed.

PR #13, opened 2025-06-20
[fearless_simd_gen]: Fix rustfmt style_edition defaulting to `2015`

ajakubowicz-canva:
### Context

Generated code is formatted with `rustfmt`. However, by default rustfmt [uses the 2015 edition](https://github.com/rust-lang/rustfmt?tab=readme-ov-file#rusts-editions). This PR adds `rustfmt.toml` such that generating the code and `cargo fmt` match.

I also snuck in a fix to an `expect` message.

### Test plan

When I run `cargo run --bin fearless_simd_gen` I now get the same result as `cargo fmt`. Sorry for not catching this in the initial PR. I could get around it by using the generator then running `cargo fmt` 🤦


fearless_simd_gen/src/main.rs
ajakubowicz-canva, line 1: This file contains an improvement to an expect message that would otherwise literally panic with the message: `"error parsing {name}"`. Now it provides the name.

LaurenzV (approved):

PR #14, opened 2025-06-22
Add more features

LaurenzV:
This PR adds a bunch of additional features which already allow us to implement
- Basic alpha compositing of fills/strips.
- Rendering of blurred rectangles.
- Rendering of gradients.

in vello_cpu, using exclusively fearless_simd. The changes include:

- Widening and narrowing instructions for u8 and u16.
- 512-bit vector types
- I removed zip/unzip and instead added a zip1/zip2 instruction for zipping the upper/lower part of the vectors.
- The `fract` and `msub` methods
- The ability to reinterpret integer types as u8

Test cases pass with both, fallback and NEON.


fearless_simd/src/generated/wasm.rs
```diff
@@ -138,7 +138,7 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn fract_f32x4(self, a: f32x4<Self>) -> f32x4<Self> {
-        todo!().simd_into(self)
```
ajakubowicz-canva: whoops! Good fix thank you!

fearless_simd/src/generated/wasm.rs
```diff
@@ -138,7 +138,7 @@ impl Simd for WasmSimd128 {
...
+        todo!()
```
ajakubowicz-canva: Thank you so much!

fearless_simd/src/generated/wasm.rs
```diff
@@ -192,7 +192,7 @@ impl Simd for WasmSimd128 {
         todo!()
     }
     #[inline(always)]
-    fn shr_i8x16(self, a: i8x16<Self>, b: i8x16<Self>) -> i8x16<Self> {
+    fn shr_i8x16(self, a: i8x16<Self>, shift: u32) -> i8x16<Self> {
```
ajakubowicz-canva: Was I running something incorrectly? I am surprised this compiled, especially given there are other WASM SIMD tests.
In any case, thank you very much for fixing this.

Edit: Disregard, I reviewed this incrementally and saw this was a change.

fearless_simd_gen/src/ops.rs
```diff
@@ -36,14 +40,16 @@ pub const FLOAT_OPS: &[(&str, OpSig)] = &[
...
-    ("zip", OpSig::Zip),
-    ("unzip", OpSig::Zip),
+    ("zip1", OpSig::Zip(true)),
+    ("zip2", OpSig::Zip(false)),
     ("max", OpSig::Binary),
     ("max_precise", OpSig::Binary),
     ("min", OpSig::Binary),
     ("min_precise", OpSig::Binary),
     ("madd", OpSig::Ternary),
+    ("msub", OpSig::Ternary),
```
ajakubowicz-canva, line 50: I don't think we need to add `msub` as this should be something that LLVM can optimize into.

Excerpt from a DM to Raph:
 > As a guiding principle, I feel like anything that can be accessed through llvm optimizations doesn't need an intrinsic API method [...] and in some cases the right thing to do might be to file issues against llvm to get more optimizations. ~@raphlinus (via DM)

fearless_simd_gen/src/ops.rs
```diff
@@ -57,14 +63,17 @@ pub const INT_OPS: &[(&str, OpSig)] = &[
...
+    ("shr", OpSig::Shift),
     ("simd_eq", OpSig::Compare),
     ("simd_lt", OpSig::Compare),
     ("simd_le", OpSig::Compare),
     ("simd_ge", OpSig::Compare),
     ("simd_gt", OpSig::Compare),
-    ("zip", OpSig::Zip),
-    ("unzip", OpSig::Zip),
+    ("zip1", OpSig::Zip(true)),
+    ("zip2", OpSig::Zip(false)),
```
ajakubowicz-canva: I prefer Highway's naming of [`ZipLower` and `ZipHigher`](https://github.com/google/highway/blob/9d1fa11ad921b97357aed0cb277ded70ad9170f8/hwy/ops/wasm_128-inl.h#L3258) vs the extremely terse name `zip1` and `zip2`. What do you think about a more readable `zip_low` and `zip_high`?

ajakubowicz-canva (commented):
So many great improvements in this PR!!! 🎉  I haven't gotten through everything but from my first pass I think the comments I've left are the major points. Anything else will probably be a nit or can be done in a followup.

fearless_simd_gen/src/ops.rs
LaurenzV: How so? Isn't this the same as for fused multiply-add, where the semantics are not exactly the same as for add + mul, and thus the compiler cannot make this optimization by default?

fearless_simd/src/generated/wasm.rs
LaurenzV: I changed this because the way shifting works between fallback/neon is slightly different, so the easiest for now was to just take one parameter for the shift that is applied to all lanes, instead of allowing different shift values for each lane. Obviously would also be a nice feature to have, but not a priority for us.

fearless_simd_gen/src/ops.rs
LaurenzV: That seems reasonable to me!

fearless_simd_gen/src/ops.rs
AndrewJakubowicz: I'll need Raph to chime in to answer your question because I'm also unclear about why madd isn't a similar consideration. I won't block on this.

AndrewJakubowicz:
> Test cases pass with both, fallback and NEON.

Just clarifying that this relates to the Vello test suite right?
Thank you!

LaurenzV:
> Just clarifying that this relates to the Vello test suite right?

Yep.

ajakubowicz-canva:
Just needs a fmt to pass CI.

fearless_simd_gen/src/ops.rs
ajakubowicz-canva: We discussed in office hours. I am going to try my best to summarize the discussion. Essentially, we already have a fused multiply-add, and I believe the intent is that a consumer of fearless_simd can express a fused multiply-sub by negating the fused multiply-add. Then LLVM should be able to optimize the negated fused multiply-add into the right underlying SIMD intrinsic.

This is preferred as we can keep our API surface smaller. It's also preferred because there's a complexity of which underling implementations to use on different architectures, so it's much simpler to leverage LLVM in these cases.

Finally we discussed removing msub and replacing usages with negeted fused multiply-adds.

@LaurenzV Does that summary capture the office hours discussion? Also, than you for bearing with me on this one 😄

PR #16, opened 2025-06-25
[WASM] fix unconditional SIMD inclusion in WASM binary

ajakubowicz-canva:
### Context

I made a mistake using `#[target_feature(enable = "simd128")]` because it means the enclosed code is compiled using the simd128 feature flag. This can result in binaries containing SIMD. Even if the SIMD is never executed, older browsers may fail to validate the WASM binary if it includes SIMD instructions.

Thus, this PR removes all usages of target_feature for WASM, instead opting into optional compilation.

### Test plan

This really needs an automated test, but I haven't found a good way to test this yet.

I manually tested by running the following command:

```
cargo build --example srgb --target=wasm32-unknown-unknown && wasm2wat ./target/wasm32-unknown-unknown/debug/examples/srgb.wasm | grep -q v128 && echo "SIMD FOUND" || echo "NO SIMD"
```

Now returns `NO SIMD` , when previously it returned `SIMD FOUND`.

```
RUSTFLAGS=-Ctarget-feature=+simd128 cargo build --example srgb --target=wasm32-unknown-unknown && wasm2wat ./target/wasm32-unknown-unknown/debug/examples/srgb.wasm | grep -q v128 && echo "SIMD FOUND" || echo "NO SIMD"
```

Returns `SIMD FOUND` because the `+simd128` feature was requested.

### DIFF
--- a/fearless_simd_gen/src/mk_wasm.rs
+++ b/fearless_simd_gen/src/mk_wasm.rs
@@ -257,7 +257,6 @@ fn mk_simd_impl(level: Level) -> TokenStream {
 
             #[inline]
             fn vectorize<F: FnOnce() -> R, R>(self, f: F) -> R {
-                #[target_feature(enable = "simd128")]
                 #[inline]
                 // unsafe not needed here with tf11, but can be justified
                 unsafe fn vectorize_simd128<F: FnOnce() -> R, R>(f: F) -> R {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
           tool: wasm-pack
 
       - name: Run fearless_simd_tests on Chrome
-        run: wasm-pack test --headless --chrome
+        run: RUSTFLAGS=-Ctarget-feature=+simd128 wasm-pack test --headless --chrome
         working-directory: fearless_simd_tests
   
   build-no-std:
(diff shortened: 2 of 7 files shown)


ajakubowicz-canva:
cc/ @taj-p

fearless_simd_gen/src/mk_wasm.rs
```diff
@@ -257,7 +257,6 @@ fn mk_simd_impl(level: Level) -> TokenStream {
 
             #[inline]
             fn vectorize<F: FnOnce() -> R, R>(self, f: F) -> R {
-                #[target_feature(enable = "simd128")]
```
LaurenzV, line 260: Don't we need the `target_arch` here?

ajakubowicz-canva: I'm going to have to think about this a bit more in the WASM case.

Although I agree that the `vectorize` method exists in order to _force_ the usage of the `simd128` target feature, hence allowing vectorization, it's not actually possible in the WASM case to even call this without the target feature already being set - since WASM chooses whether SIMD is allowed at compile time.

I.e. this `vectorize` method ends up within `generated/wasm.rs`, which is gated behind a `#[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]`.

I think this API doesn't make sense for WASM, as in WASM you'd reach for `target_feature` if you wanted vectorization. I'm leaning towards doing nothing at the moment. Maybe I could add a comment?

LaurenzV: Sure, just was wondering!

ajakubowicz-canva: It's a really good questions :D
Thank you for reviewing!

PR #17, opened 2025-06-25
Add `trunc` and interleaved loading in neon and fallback

LaurenzV:
This adds a couple of more features required for the SIMD version of `vello_cpu`, in particular:

- Support for the `trunc` instruction.
- Support for the `wrapping_sub` instruction.
- Supported for interleaved loading in 128-bit blocks.


ajakubowicz-canva (commented):
Some initial comments. I am happy with `trunc` and interleaved loading. However, I think `wrapping_sub` may be something we don't need to explicitly add. I assume that a followup PR would then need to implement `wrapping_*` variants of the other math operations.

Maybe it may be useful for @raphlinus to weigh in. Using Highway's design philosophy, they do not separate `sub` from `wrapping_sub`. excerpt:

> 9. Ensuring signed integer overflow has well-defined semantics (wraparound).
> [Link](https://github.com/google/highway/blob/master/g3doc/design_philosophy.md#:~:text=Ensuring%20signed%20integer%20overflow%20has%20well%2Ddefined%20semantics%20(wraparound).)

My question is - do we want to provide explicitly different API's for wrapping and non wrapping operations?

Great work!

Edit: I also want to add a caveat that I'm not even close to a reliably source here, hence why I defer to Raph. I won't block review though. I'll try get a chance to review tomorrow.

LaurenzV:
I'm just not sure that using `wrapping_sub` in all instances where you have a simple subtraction for fallback might lead to performance issues on some platforms where there isn't an explicit instruction for that.

Other than that, for me it would be fine to merge the two.

ajakubowicz-canva (commented):
A quick note that I've reviewed `trunc` and `wrapping_sub` and that work LGTM.

I just have interleaved to review but ran out of time. If there's urgency, you could split out those instructions and land incrementally. Otherwise I'll complete review after weekend!

fearless_simd_gen/src/arch/wasm.rs
```diff
@@ -15,7 +15,8 @@ fn translate_op(op: &str) -> Option<&'static str> {
         "floor" => "floor",
         "sqrt" => "sqrt",
         "add" => "add",
-        "sub" => "sub",
+        // TODO: Is wrapping sub same on WASM?
```
ajakubowicz-canva, line 18: In WASM:
 - `sub` is wrapping. ([ref](https://github.com/WebAssembly/simd/blob/main/proposals/simd/SIMD.md#integer-subtraction))
 - `sub_sat` is saturating. ([ref](https://github.com/WebAssembly/simd/blob/main/proposals/simd/SIMD.md#saturating-integer-subtraction))

I am a little concerned that `wrapping_sub` was added because the **fallback** architecture doesn't wrap by default but panics.

Highway defines `sub` as wrapping by default, and then provides an additional API for [saturating sub](https://github.com/google/highway/blob/9d1fa11ad921b97357aed0cb277ded70ad9170f8/hwy/ops/scalar-inl.h#L587).

Both WASM SIMD and Neon implement sub as wrapping by default.
Similarly, once we add Intel Intrinsics (x86) like `_mm_sub_epi8` for subtraction, those also wrap by default.


The major architectures all wrap by default, then offer an alternative of saturating.

My thoughts about what we should do. We should only do one of these proposals, I am leaning towards proposal 1):

### Proposal 1

We follow Highway, and define `sub` as wrapping subtraction. We remove `wrapping_sub` and instead modify the fallback implementation of `sub` to wrap.

Then in the future we can add [`saturating_sub`](https://github.com/google/highway/blob/9d1fa11ad921b97357aed0cb277ded70ad9170f8/hwy/ops/scalar-inl.h#L582).

In total we'll end up with two variants of subtraction:
 - `sub`: always wrapping.
 - `saturating_sub`: always saturating.

### Proposal 2

We make it explicit that `sub` (and other arithmetic operations) are "implementation specific" out-of-bounds. To make it explicit we could add some prefix like `relaxed` (inspired by WASM [relaxed-simd proposal](https://github.com/WebAssembly/relaxed-simd/blob/main/proposals/relaxed-simd/Overview.md)). Thus, we rename `sub` to `relaxed_sub`.

This would mean we'd end up with 3 variants eventually:
 - `relaxed_sub`
 - `wrapping_sub`
 - `saturating_sub`

### My thoughts

I think we should start with proposal 1, as I think all we need to do is change the fallback implementation to always wrap. Then in the future we can consider a relaxed variant.

What do you think?

fearless_simd/src/generated/wasm.rs
```diff
@@ -2412,6 +2522,10 @@ impl Simd for WasmSimd128 {
         (b0.simd_into(self), b1.simd_into(self))
     }
     #[inline(always)]
+    fn load_interleaved_128_u8x64(self, src: &[u8; 64usize]) -> u8x64<Self> {
+        todo!()
+    }
```
ajakubowicz-canva, line 2431: [non actionable thinking out loud] Since WASM doesn't support interleaved instructions at all, I wonder if there's a pattern that we can do where the implementation is basically:

 - transmute/cast v128 to slice
 - run the fallback logic
 - transmute/cast back to v128

ajakubowicz-canva (commented):
Sorry for the slow review.

I'm happy with `trunc` and `interleaved` but on further thought I don't think we should have `sub` and `wrapping_sub`. I think there's a high chance if we add `wrapping_sub` in this PR, we'll want to remove it in the future (or add more explicit subtraction variants).

fearless_simd_gen/src/arch/wasm.rs
raphlinus: I'm strongly in favor of wrapping semantics for `sub`, and feel like this is the semantics of all SIMD intrinsics already, we should as a philosophical principle be using that as a source of truth and make fallback match those, as opposed to the autovectorization philosophy where scalar is the source of truth and we try to make vectors match.

fearless_simd/src/generated/wasm.rs
LaurenzV: That's probably possible, yeah. But can't you load the elements normally and then do a shuffle?

ajakubowicz-canva: Good point!

PR #19, opened 2025-06-26
Add clippy to CI (as non blocking warnings)

ajakubowicz-canva:
### Context

This PR adds the Linebender clippy warnings to CI. Because CI doesn't block us it's ok to land. Ideally I'd like all the clippy lints to become errors and for us to burn down the clippy warnings.

I also deleted some embarrassing `println!` debug statements that I accidentally merged into `main` 🤦 

Finally, I increased the scope of the `fmt` check to all packages.

### Goal

In the longer term I think it's important we conform to clippy, if only because it'll catch issues.

As a random note, I couldn't use `fearless_simd` with `vello` using `bytemuck` version `1.23.1`, due to conflicting dependencies in `vello`. Using `1.23.0` matches the version of [bytemuck in vello](https://github.com/linebender/vello/blob/main/Cargo.toml#L100).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,11 +66,6 @@ jobs:
       - name: cargo clippy
         run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features std
 
-      # TODO: Add back the suffix ` -- -D warnings` onto this command to make warnings errors.
-      # Currently there are too many warnings to turn on this lint.
-      - name: cargo clippy (auxiliary)
-        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features std --tests --benches --examples
-
 
   clippy-stable-wasm:
     name: cargo clippy (wasm32)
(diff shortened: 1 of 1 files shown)


fearless_simd/src/macros.rs
ajakubowicz-canva, line 1: Sorry for landing this!

.github/workflows/ci.yml
```diff
@@ -23,9 +23,87 @@ jobs:
           toolchain: ${{ env.RUST_STABLE_VER }}
           components: rustfmt
 
-      # For now, we only check `fearless_simd_gen`, since `fearless_simd` uses `prettyplease`.
       - name: cargo fmt
-        run: cd fearless_simd_gen && cargo fmt  --check
+        run: cargo fmt --check
```
LaurenzV: Do we need to pass `--all` here or does it work just like that?

Cargo.toml
```diff
@@ -6,5 +6,54 @@ members = ["fearless_simd", "fearless_simd_gen", "fearless_simd_tests"]
 license = "MIT/Apache-2.0"
 edition = "2024"
 
+
+[workspace.lints]
+
+# LINEBENDER LINT SET - Cargo.toml - v5
```
LaurenzV, line 12: I'm not sure whether we really want _all_ of those lints as some might become a bit annoying to deal with and unnecessary (see also currently failing CI) since we are dealing with code generation, but I'm okay with leaving it as is for now.

PR #21, opened 2025-06-26
Implement SIMD boolean operations for WASM

no-materials:
This PR implements the previously TODO-marked SIMD boolean operations for WASM targets.

### Changes Made

- The reinterpretation logic for these operations was already implemented in the codebase. This PR removes the special conditions that were preventing these operations from being generated through the normal code flow.

-  Implemented tests for each boolean operation, currently covering the `i8x16` vector type to verify correctness of the implementation.

### Operations Implemented

The following SIMD boolean operations are now available for WASM:
- `and`, `or`, `xor`, `not`

### Testing

Added test coverage for each operation using the `i8x16` vector type. 

Should I extend the test suite to cover additional vector types, or is the current `i8x16` coverage sufficient?

### DIFF
--- a/fearless_simd_tests/tests/wasm.rs
+++ b/fearless_simd_tests/tests/wasm.rs
@@ -283,3 +283,81 @@ test_wasm_simd_parity! {
         }
     }
 }
+
+test_wasm_simd_parity! {
+    fn and_u8x16() {
+        |s| -> [u8; 16] {
+            let a = u8x16::from_slice(s, &[1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0]);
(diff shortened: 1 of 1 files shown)


LaurenzV:
Thanks! Will let @ajakubowicz-canva do the review because he is more familiar with WASM.

> Should I extend the test suite to cover additional vector types, or is the current i8x16 coverage sufficient?

I think that's up to you, in the future we will obviously want to have tests for "everything", but we don't really have a proper test suite in place. This will happen eventually, but right now our focus is just to add all operations we need to vello.

ajakubowicz-canva (approved):
Thank you so much! Changes look great. I'll merge once CI passes.

PR #23, opened 2025-06-30
Fix WASM `pmax` and `pmin` such that vello_cpu tests pass

ajakubowicz-canva:
> [!Note]
> Min/Max corner cases are target-specific and may change. If either argument is qNaN, x86 SIMD returns the second argument, Armv7 Neon returns NaN, Wasm is supposed to return NaN but does not always, but other targets actually uphold IEEE 754-2019 minimumNumber: returning the other argument if exactly one is qNaN, and NaN if both are.
> ~[Highway docs for Min/Max](https://github.com/google/highway/blob/0e759dd39119644c902eb0dcda2cfe03483c40ac/g3doc/quick_reference.md?plain=1#L702-L706)

Addresses issue https://github.com/raphlinus/fearless_simd/issues/20

### Before this change

vello_cpu with WASM SIMD tests had diffs like so:

Screenshot 2025-06-27 at 4 13 47 pm
Screenshot 2025-06-27 at 4 13 02 pm

### After this change

All vello_cpu WASM SIMD tests pass. (Note, I needed to manually test this as we don't automatically test vello_cpu on WASM SIMD. (Although I think we should consider it in the near term).

### Why?

This is basically explained by the comment https://github.com/raphlinus/fearless_simd/issues/20#issuecomment-3017811768

In vello_cpu we have the following code:

https://github.com/linebender/vello/blob/main/sparse_strips/vello_common/src/strip.rs#L327-L330

Inlined:

```rs
let line_px_left_y = line_top_y
  .madd(px_left_x - line_top_x, y_slope)
  .max_precise(ymin)
  .min_precise(ymax);
```

This code can multiply `f32::INFINITY` with `0` and result in `NaN`. The `max_precise` is used to massage `NaN` to `ymin`. In order to accomplish this we need to match Highway's semantics described in https://github.com/raphlinus/fearless_simd/issues/20#issuecomment-3017811768
.

Note, the three (neon, x86, WASM) implementations are not exactly the same in all situations, but for vello_common strip generation, they should all behave the same after this PR.

### Test plan

I used a `path` config in Cargo.toml from `vello` in order to test the `vello_cpu` wasm tests against this change. They went from all failing to all passing with `RUSTFLAGS=-Ctarget-feature=+simd128 wasm-pack test --chrome`.

I also added two very specific tests that enforce the NaN handling behavior of WASM in `max_precise` and `min_precise`.

### DIFF
--- a/fearless_simd_gen/src/mk_wasm.rs
+++ b/fearless_simd_gen/src/mk_wasm.rs
@@ -108,6 +108,19 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                                 }
                             }
                         }
+                        "max_precise" | "min_precise" => {
+                            // For `max_precise` and `min_precise` the arguments are switched such
+                            // that `max(NaN, x)` and `min(NaN, x)` result in `x`. This matches
+                            // `_mm_max_ps` and `_mm_min_ps` semantics on x86.
+                            let swapped_args = [quote! { b.into() }, quote! { a.into() }];
+                            let expr: TokenStream = Wasm.expr(method, vec_ty, &swapped_args);
+                            quote! {
+                                #[inline(always)]
+                                fn #method_ident(self, a: #ty<Self>, b: #ty<Self>) -> #ret_ty {
+                                    #expr.simd_into(self)
+                                }
+                            }
+                        }
                         _ => {
                             let expr = Wasm.expr(method, vec_ty, &args);
                             quote! {
--- a/fearless_simd_tests/tests/wasm.rs
+++ b/fearless_simd_tests/tests/wasm.rs
@@ -1,7 +1,7 @@
 // Copyright 2025 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-#![cfg(target_arch = "wasm32")]
+#![cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
 
 use fearless_simd::*;
 use wasm_bindgen_test::*;
--- a/fearless_simd_tests/tests/wasm.rs
+++ b/fearless_simd_tests/tests/wasm.rs
@@ -1,9 +1,9 @@
 // Copyright 2025 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-#[cfg(target_arch = "wasm32")]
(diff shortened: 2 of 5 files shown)


fearless_simd_gen/src/mk_wasm.rs
```diff
@@ -108,6 +108,19 @@ fn mk_simd_impl(level: Level) -> TokenStream {
...
+                            // For `max_precise` and `min_precise` the arguments are switched such
+                            // that `max(NaN, x)` and `min(NaN, x)` result in `x`. This matches
+                            // `_mm_max_ps` and `_mm_min_ps` semantics on x86.
+                            let swapped_args = [quote! { b.into() }, quote! { a.into() }];
+                            let expr: TokenStream = Wasm.expr(method, vec_ty, &swapped_args);
+                            quote! {
+                                #[inline(always)]
+                                fn #method_ident(self, a: #ty<Self>, b: #ty<Self>) -> #ret_ty {
+                                    #expr.simd_into(self)
+                                }
```
LaurenzV, line 121: So this assumes that the user will always put the value that can be `NaN` as the first argument to achieve that specific behavior?

fearless_simd_tests/tests/wasm.rs
```diff
@@ -1,9 +1,9 @@
 // Copyright 2025 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-#[cfg(target_arch = "wasm32")]
+#![cfg(target_arch = "wasm32")]
```
LaurenzV: Shouldn't there be a flag for the target feature as well? Otherwise, what happens when you run the test on a WASM target that doesn't support SIMD?

fearless_simd_gen/src/mk_wasm.rs
ajakubowicz-canva: That's right. Which should match x86 behavior when we get there.

fearless_simd_tests/tests/wasm.rs
ajakubowicz-canva: Yes! Otherwise the whole test suite crashes. Fixed in https://github.com/raphlinus/fearless_simd/pull/23/commits/0a9fddf96dd1c1f15006c30dd10da6613d6f5c8f

PR #24, opened 2025-07-02
[wasm] add `load_interleaved`

ajakubowicz-canva:
### Context

Added interleaving to WASM. [Used Highway as a reference](https://github.com/google/highway/blob/9d1fa11ad921b97357aed0cb277ded70ad9170f8/hwy/ops/wasm_128-inl.h#L3091C1-L3256C2).

### Test plan

Added three unit tests which cover the three new interleave methods:
 - `load_interleaved_128_u32x16`
 - `load_interleaved_128_u16x32`
 - `load_interleaved_128_u8x64`


LaurenzV:
Ouch... I really hope that this won't tank performance, otherwise we have a problem... But luckily this one is only used for packing, so if performance is really bad then maybe we can fall back to the scalar implementation for WASM.

Interleaved storing will be used for gradients and blending, though.

fearless_simd/src/generated/wasm.rs
```diff
@@ -2390,7 +2390,36 @@ impl Simd for WasmSimd128 {
...
+        );
+        let out3 = u8x16_shuffle::<8, 24, 9, 25, 10, 26, 11, 27, 12, 28, 13, 29, 14, 30, 15, 31>(
+            v02_upper, v13_upper,
+        );
+        let mut result = [0; 64usize];
+        unsafe { v128_store(result[0..].as_mut_ptr() as *mut v128, out0) };
+        unsafe { v128_store(result[16usize..].as_mut_ptr() as *mut v128, out1) };
+        unsafe { v128_store(result[16usize * 2..].as_mut_ptr() as *mut v128, out2) };
+        unsafe { v128_store(result[16usize * 3..].as_mut_ptr() as *mut v128, out3) };
+        result.simd_into(self)
```
LaurenzV: Shouldn't it be possible to use `simd.combine_u8x16/simd.combine_u8x32`? Though it probably doesn't make a difference, but at least we could get rid of some unsafe code.

LaurenzV: (Same also for the other occurences.

ajakubowicz-canva: Done: https://github.com/raphlinus/fearless_simd/pull/24/commits/e5a8d46e1500baa3cf2dd2d3cf06060742daad6e

Thanks for a really excellent suggestion.

PR #25, opened 2025-07-02
Add more features

LaurenzV:
This PR adds the following methods:
- Convert from u32 to f32.
- Interleaved loading for f32.
- Interleaved storing for various types.
- Fixes a bug where for neon, we used `vcvtnq_u32_f32`, which does conversion while rounding to nearest, instead of `vcvtq_u32_f32`, which has the same truncating as scalar. I guess the question here is which one should be the default, but for `vello_cpu` we at least need the truncating behavior.

### DIFF
--- a/fearless_simd/src/generated/neon.rs
+++ b/fearless_simd/src/generated/neon.rs
@@ -169,7 +169,7 @@ impl Simd for Neon {
     }
     #[inline(always)]
     fn cvt_u32_f32x4(self, a: f32x4<Self>) -> u32x4<Self> {
-        unsafe { vcvtnq_u32_f32(a.into()).simd_into(self) }
+        unsafe { vcvtq_u32_f32(a.into()).simd_into(self) }
     }
     #[inline(always)]
     fn splat_i8x16(self, val: i8) -> i8x16<Self> {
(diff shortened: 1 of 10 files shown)


fearless_simd/src/generated/neon.rs
```diff
@@ -169,7 +169,7 @@ impl Simd for Neon {
     }
     #[inline(always)]
     fn cvt_u32_f32x4(self, a: f32x4<Self>) -> u32x4<Self> {
-        unsafe { vcvtnq_u32_f32(a.into()).simd_into(self) }
+        unsafe { vcvtq_u32_f32(a.into()).simd_into(self) }
```
ajakubowicz-canva, line 172: I think it makes sense for conversion to be truncating.

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -160,6 +163,27 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                         }
                     }
                 }
+                OpSig::StoreInterleaved(block_size, count) => {
+                    let intrinsic = {
+                        // The function expects 64-bit or 128-bit
```
ajakubowicz-canva, line 168: Should this be asserted?

LaurenzV: It's already asserted in the `opt_q` method:

```rs
pub fn opt_q(ty: &VecType) -> &'static str {
    match ty.n_bits() {
        64 => "",
        128 => "q",
        _ => panic!("unsupported simd width"),
    }
}
```

PR #26, opened 2025-07-04
[wasm] Add `fract`, `trunc`, `zip_low` and `zip_high`

ajakubowicz-canva:
### Context

This PR adds support for `fract`, `trunc`, `zip_low`, and `zip_high`.

### Test plan

Added unit tests.

### DIFF
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -101,11 +101,11 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn zip_low_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        todo!()
+        u32x4_shuffle::<0, 4, 1, 5>(a.into(), b.into()).simd_into(self)
     }
     #[inline(always)]
     fn zip_high_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        todo!()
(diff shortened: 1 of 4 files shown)


fearless_simd/src/generated/wasm.rs
```diff
@@ -101,11 +101,11 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn zip_low_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        todo!()
+        u32x4_shuffle::<0, 4, 1, 5>(a.into(), b.into()).simd_into(self)
```
LaurenzV, line 104: Is there no shuffle for f32x4?

LaurenzV (approved):

fearless_simd/src/generated/wasm.rs
ajakubowicz-canva: I wondered a similar thing and fun fact, there is only the single shuffle instruction available in WASM: `i8x16.shuffle(a: v128, b: v128, imm: ImmLaneIdx32[16]) -> v128` ([ref](https://github.com/WebAssembly/spec/blob/main/proposals/simd/SIMD.md#shuffling-using-immediate-indices)).

All the various shuffles defined in [Rust](https://doc.rust-lang.org/beta/core/arch/wasm32/index.html) are just abstractions over that single wasm SIMD shuffle. And I think that's why there is no f32x4 abstraction. (Although I am guessing).

PR #27, opened 2025-07-07
[wasm] Add copysign, select, shr, reinterpret, and cvt

ajakubowicz-canva:
### Context

Add a bunch of operations to wasm32 architecture. This is actually enough to run the vello_cpu SIMD demo.

### Test plan

I unit tested, but I also tested against https://github.com/linebender/vello/pull/1078

### Results

TODO

### DIFF
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -76,8 +76,10 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn copysign_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        /// TODO: copysign
-        todo!()
+        let sign_mask = f32x4_splat(-0.0_f32);
+        let sign_bits = v128_and(b.into(), sign_mask.into());
+        let magnitude = v128_andnot(a.into(), sign_mask.into());
+        v128_or(magnitude, sign_bits).simd_into(self)
     }
     #[inline(always)]
     fn simd_eq_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> mask32x4<Self> {
@@ -145,7 +147,7 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn select_f32x4(self, a: mask32x4<Self>, b: f32x4<Self>, c: f32x4<Self>) -> f32x4<Self> {
-        todo!()
+        v128_bitselect(b.into(), c.into(), a.into()).simd_into(self)
     }
     #[inline(always)]
     fn combine_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x8<Self> {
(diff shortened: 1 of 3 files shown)


ajakubowicz-canva:
I can take a screenshare video - but I get wildly variable results between runs and performance tests. I need to hook up my benchmark tool and take a rigorous benchmark.

ajakubowicz-canva:
Also – I think the reason the measurements are different is that the `devicePixelRatio` of the screen impacts the numbers a lot. So I'd just look at the magnitude difference at the moment because I take them side by side. It would also be interesting to see what you get.

I run the cpu example, go to the cowboy emoji, then zoom in the screen until the eyes are touching the sides – that gets me the largest "frame lag" that I measure. Definitely need a more scientific approach in the future though :P

ajakubowicz-canva:
@LaurenzV I've done the following:
 1. I finished adding operations so now _all_ WASM methods are implemented
 2. I screen-recorded my back-of-the-napkin performance profiles. (See repro steps in description and the video).

It's ready for review :D

LaurenzV:
I tried it and it indeed seems to be faster! Perhaps there is even more speedup to be gained if we can SIMDify the `f32_to_u8` mehod.

fearless_simd_gen/src/mk_wasm.rs
```diff
@@ -243,22 +233,87 @@ fn mk_simd_impl(level: Level) -> TokenStream {
...
+                                    let low_masked = v128_and(low.into(), mask);
+                                    let high_masked = v128_and(high.into(), mask);
+                                    let result = u8x16_narrow_i16x8(low_masked, high_masked);
+                                    result.simd_into(self)
+                                }
+                            }
+                        }
+                        _ => unimplemented!(),
+                    }
+                    // let to_ty = &VecType::new(scalar, scalar_bits, vec_ty.len);
```
LaurenzV: Is there a reason to keep this?

fearless_simd/src/generated/wasm.rs
```diff
@@ -2159,11 +2168,45 @@ impl Simd for WasmSimd128 {
...
+        let out0 = u32x4_shuffle::<0, 4, 1, 5>(v02_lower, v13_lower);
+        let out1 = u32x4_shuffle::<2, 6, 3, 7>(v02_lower, v13_lower);
+        let out2 = u32x4_shuffle::<0, 4, 1, 5>(v02_upper, v13_upper);
+        let out3 = u32x4_shuffle::<2, 6, 3, 7>(v02_upper, v13_upper);
+        unsafe {
+            v128_store(dest[0 * 4usize..].as_mut_ptr() as *mut v128, out0);
+            v128_store(dest[1 * 4usize..].as_mut_ptr() as *mut v128, out1);
+            v128_store(dest[2 * 4usize..].as_mut_ptr() as *mut v128, out2);
+            v128_store(dest[3 * 4usize..].as_mut_ptr() as *mut v128, out3);
+        }
```
LaurenzV, line 2209: Since we know that dest is of length 16 it might not be critical, but maybe we should also add a [..4] so that it's really clear that each element is 4 in length? Same for the other occurences. But not sure.

fearless_simd_gen/src/mk_wasm.rs
ajakubowicz-canva: Good catch! Whoops!

fearless_simd/src/generated/wasm.rs
ajakubowicz-canva: Will investigate in a followup if you don't mind. Will merge so that your PR can point to this new commit.

PR #31, opened 2025-07-10
Introduce semantic conversion traits

Ralith:
This proposes a generic API for conversions between vectors of the same element size and count, usable on generic native-width vectors. If I get positive feedback I'll fill out some more cases before merging.


Ralith:
Open question: Should this supplant `Simd::cvt_*`? If not, it could probably be blanket-implemented based on those methods rather than separately implemented for each backend as in this PR.

Ralith:
Tweaked naming, factored out common generation logic, and expanded to a superset of existing conversion operations. I think this is ready now, though still happy to brainstorm on better names.

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -364,6 +352,28 @@ fn mk_simd_impl(level: Level) -> TokenStream {
...
+    let cvts = crate::ops::conversions().map(|cvt| {
+        let src_prefix = scalar_prefix(cvt.src_scalar);
+        let dst_prefix = scalar_prefix(cvt.dst_scalar);
+        let bits = cvt.scalar_bits;
+        let op = match cvt.len {
+            2 => "vcvt",
+            4 => "vcvtq",
+            _ => unimplemented!(),
+        };
+        // FIXME: Saturation?
```
Ralith: I guess I should handle this with extra logic, and we can introduce unsafe unchecked truncation later if needed?

fearless_simd/src/generated/simd_trait.rs
```diff
@@ -667,6 +669,10 @@ pub trait SimdFloat<Element: SimdElement, S: Simd>:
     + core::ops::Div<Output = Self>
     + core::ops::Div<Element, Output = Self>
 {
+    #[inline(always)]
+    fn to_int<T: SimdCvtTruncate<Self>>(self) -> T {
```
Ralith, line 874: Naming bikeshed thread. Ideas: `to_int_truncate`, `truncate_to_int`, one of those but with `trunc`...

fearless_simd/src/traits.rs
```diff
@@ -105,3 +105,13 @@ impl SimdElement for u32 {
...
 }
+
+/// Construction of integer vectors from floats by truncation
+pub trait SimdCvtTruncate<T> {
+    fn truncate_from(x: T) -> Self;
+}
+
+/// Construction of floating point vectors from integers
+pub trait SimdCvtFloat<T> {
+    fn float_from(x: T) -> Self;
```
Ralith, line 124: Naming bikeshed: `from_int`, `convert_from_int`, ...

fearless_simd_gen/src/mk_wasm.rs
ajakubowicz-canva, line 1: This change would make an excellent standalone change (separate from the overall proposal).

Ralith: Split into https://github.com/linebender/fearless_simd/pull/38. Github sucks at PR chains, so it'll be in both PRs for now as it's a merge conflict otherwise, but if that one can be merged quicker than all should be well.

Ralith:
Rebased and regenerated.

Ralith:
Whoops, needs some tweaks to handle i64.

Ralith:
We don't actually *have* u/i64 yet, so I've just masked those cases off for now.

LaurenzV:
Thanks!

PR #32, opened 2025-07-10
Drop syn and prettyplease dependencies from generator

Ralith:
Faster generator build and less risk of formatting drift.


fearless_simd_gen/src/main.rs
```diff
@@ -56,11 +56,21 @@ impl Module {
...
+    fn generate_string(self) -> String {
         let code = self.generate_code();
-        let parsed = syn::parse_file(&code.to_string()).expect(&format!("error parsing {name}"));
-        let code_str = prettyplease::unparse(&parsed);
-        format!("// This file is autogenerated by fearless_simd_gen\n\n{code_str}")
+        let mut child = std::process::Command::new("rustfmt")
+            .stdin(std::process::Stdio::piped())
+            .stdout(std::process::Stdio::piped())
+            .spawn()
+            .expect("Failed to spawn `rustfmt`");
```
ajakubowicz-canva: Extreme nit: Per [Common Message Styles](https://doc.rust-lang.org/std/error/index.html#common-message-styles) prefer: "describe the reason we expect the Result should be Ok"

```suggestion
            .expect("`rustfmt` should spawn");
```

ajakubowicz-canva (approved):
Thank you for an amazing contribution. Looks great to me (with a super small nit)

fearless_simd_gen/src/main.rs
Ralith: Done.

ajakubowicz-canva:
Are you able to merge or do you need me to merge? Thanks for a great contribution!

Ralith:
I don't have the commit bit, no.

PR #33, opened 2025-07-10
Add another set of features

LaurenzV:
This PR adds the following features:
- Vector types for f64 with a limited set of operations for now to keep things simple.
- The `unzip` method.
- A couple more reinterpret operations.

WASM SIMD is still missing, @ajakubowicz-canva maybe you can take care of it once this is landed? These features are needed for SIMD flattening.

### DIFF
--- a/fearless_simd_gen/src/arch/neon.rs
+++ b/fearless_simd_gen/src/arch/neon.rs
@@ -60,10 +60,10 @@ impl Arch for Neon {
     // expects args and return value in arch dialect
     fn expr(&self, op: &str, ty: &VecType, args: &[TokenStream]) -> TokenStream {
         // There is no logical NOT for 64-bit, so we need this workaround.
-        if op == "not" && ty.scalar_bits == 64 && ty.scalar == ScalarType::Mask  {
-            return quote! { vreinterpretq_s64_s32(vmvnq_s32(vreinterpretq_s32_s64(a.into()))) }
+        if op == "not" && ty.scalar_bits == 64 && ty.scalar == ScalarType::Mask {
+            return quote! { vreinterpretq_s64_s32(vmvnq_s32(vreinterpretq_s32_s64(a.into()))) };
         }
-        
+
         if let Some(xlat) = translate_op(op) {
             let intrinsic = simple_intrinsic(xlat, ty);
             return quote! { #intrinsic ( #( #args ),* ) };


fearless_simd_gen/src/mk_neon.rs
```diff
@@ -331,6 +337,20 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                         }
                     }
                 }
+                OpSig::Unzip(zip1) => {
```
ajakubowicz-canva: Nit: Maybe `zip1` can be replaced with something like `select_even`?

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -131,7 +131,13 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                 }
                 OpSig::Unary => {
                     let args = [quote! { a.into() }];
-                    let expr = Neon.expr(method, vec_ty, &args);
+
+                    let expr = if vec_ty.scalar == ScalarType::Mask && vec_ty.scalar_bits == 64 {
+                        quote! { vreinterpretq_s64_s32(vmvnq_s32(vreinterpretq_s32_s64(a.into()))) }
```
ajakubowicz-canva: Shouldn't this check for a specific `method` instead of checking the `vec_ty` fields?

ajakubowicz-canva: I think it might be good to move this into `Neon.expr`?

fearless_simd/src/generated/ops.rs
```diff
@@ -2799,3 +3038,122 @@ impl<S: Simd> core::ops::Not for mask32x16<S> {
...
+    }
+}
+impl<S: Simd> core::ops::BitOr for mask64x8<S> {
+    type Output = Self;
+    #[inline(always)]
+    fn bitor(self, rhs: Self) -> Self::Output {
+        self.simd.or_mask64x8(self, rhs)
+    }
+}
+impl<S: Simd> core::ops::BitXor for mask64x8<S> {
```
ajakubowicz-canva, line 3146: Non blocking and probably very naive - but is it really advantageous to have 512 bit lanes? (I am biased to the 128bit lanes that WASM supports :P )

fearless_simd/src/generated/simd_types.rs
```diff
@@ -4037,211 +4253,1535 @@ impl<S: Simd> Bytes for f32x16<S> {
...
+    }
+}
+impl<S: Simd> From<mask32x8<S>> for [i32; 8] {
+    #[inline(always)]
+    fn from(value: mask32x8<S>) -> Self {
+        value.val
+    }
+}
+impl<S: Simd> core::ops::Deref for mask32x8<S> {
+    type Target = [i32; 8];
```
ajakubowicz-canva, line 4434: Another naive question - unrelated to this PR specifically, but shouldn't the underlying type for mask be `[bool; 8]` instead of i32?

I am just curious what are valid values for a Mask?

ajakubowicz-canva (approved):
Nice nice nice!

Merge this when ready, and I'll handle the WASM implementation separately.
I read over lots of the generated code and it looks good! Only small nits and some clarifications.

There is just one piece of code that I'm confused about, and it's the `vreinterpretq_s64_s32(vmvnq_s32(vreinterpretq_s32_s64(a.into())))` [here](https://github.com/linebender/fearless_simd/pull/33/files#diff-6b0e0ac9ded3898eeb48a564d7f7852b8218f097d4940cdd4d4d25acdb25194aR136)?

fearless_simd_gen/src/mk_neon.rs
LaurenzV: Yep you are right whoops, I didn't notice because the only unary expression for masks is NOT.

ajakubowicz-canva: Awesome! Adding explicit check for the `not` method then PR LGTM :D

fearless_simd/src/generated/ops.rs
LaurenzV: No worries, the SIMD flattening is written with mainly 128/256-bit in mind: https://github.com/linebender/vello/blob/b0c969063a40cc6ce1ceba11fed82781835c168a/sparse_strips/vello_common/src/flatten_simd.rs :)

I don't even think I'm using any of the 512-bit types explicitly, iirc it just was some of the code generation stuff that required me to introduce those types as well.

fearless_simd/src/generated/simd_types.rs
LaurenzV: I'm also not 100% sure, maybe to make it more compatible with the SIMD types which usually use integers (all 0 or all 1) as the main representation? Raph would now better.

PR #34, opened 2025-07-11
Test float-to-int conversion saturation

Ralith:
I wasn't sure if Neon behaved as intended here. CI says it does, but the test seems nice to have for clarity of intent and future backends.

### DIFF
--- a/fearless_simd_tests/tests/mod.rs
+++ b/fearless_simd_tests/tests/mod.rs
@@ -1,8 +1,36 @@
 // Copyright 2025 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+use fearless_simd::*;
+
 #[cfg(target_arch = "wasm32")]
 wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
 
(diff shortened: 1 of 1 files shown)


ajakubowicz-canva:
I think the `CI.yaml` file may need updating.

The GitHub actions log: https://github.com/linebender/fearless_simd/actions/runs/16209298613/job/45766237902?pr=34 shows 0 tests run.

Ralith:
cargo target selection logic is a complete mystery to me, but this seems to have helped. Not sure if there's a way to get this test to run in wasm too.

ajakubowicz-canva (approved):
> Not sure if there's a way to get this test to run in wasm too.

This PR doesn't need to block on WASM unless you want to try add it. It might be as simple as expanding the `#[test]` into:

```rs
#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
#[cfg_attr(not(target_arch = "wasm32"), test)]
fn saturate_float_to_int() {
  ...
}
``` 

Awesome work!

Ralith:
I'll leave that for future work, since we'll probably want something more concise as the number of tests grows and there's probably a helper out there somewhere.

PR #35, opened 2025-07-13
Linebender standards README improvement

ajakubowicz-canva:
Small documentation only PR to move the README.md towards Linebender standards.

Reference thread: https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Linebender.20Standards/with/527927335

I've added a large warning that this is a very experimental crate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,40 @@
-# Towards "fearless SIMD"
+<!--
+
+This repo-level readme needs restructuring, pending some Linebender templating decisions.
+https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Bikeshedding.20badges/with/452312397
+
+For now, prefer updating the package-level readmes, e.g. fearless_simd/README.md.
+
+-->
+
+<div align="center">
+
+# Fearless SIMD
+
+**Safer and easier SIMD**
+
+[![Linebender Zulip](https://img.shields.io/badge/Linebender-%23vello-blue?logo=Zulip)](https://xi.zulipchat.com/#narrow/channel/197075-vello)
+[![dependency status](https://deps.rs/repo/github/linebender/fearless_simd/status.svg)](https://deps.rs/repo/github/linebender/fearless_simd)
+[![Apache 2.0 or MIT license.](https://img.shields.io/badge/license-Apache--2.0_OR_MIT-blue.svg)](#license)
+
+[![Crates.io](https://img.shields.io/crates/v/fearless_simd.svg)](https://crates.io/crates/fearless_simd)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,40 @@
...
+
+-->
+
+<div align="center">
+
+# Fearless SIMD
+
+**Safer and easier SIMD**
+
+[![Linebender Zulip](https://img.shields.io/badge/Linebender-%23vello-blue?logo=Zulip)](https://xi.zulipchat.com/#narrow/channel/197075-vello)
```
DJMcNab: This should be `#simd`.

We seem to have standardised on using `blue` for all badges; we do have a couple of repos which use e.g. orange, but I think consistently using Zulip's the same colour is valuable

README.md
```diff
@@ -1,6 +1,40 @@
...
+[![dependency status](https://deps.rs/repo/github/linebender/fearless_simd/status.svg)](https://deps.rs/repo/github/linebender/fearless_simd)
+[![Apache 2.0 or MIT license.](https://img.shields.io/badge/license-Apache--2.0_OR_MIT-blue.svg)](#license)
+
+[![Crates.io](https://img.shields.io/crates/v/fearless_simd.svg)](https://crates.io/crates/fearless_simd)
+[![Docs](https://docs.rs/vello/badge.svg)](https://docs.rs/vello)
+[![Build status](https://github.com/linebender/fearless_simd/workflows/CI/badge.svg)](https://github.com/linebender/fearless_simd/actions)
```
DJMcNab: Nit: For the alt texts here, we should probably follow the pattern from https://github.com/linebender/vello/pull/601

(See also the `{badge1}\[newline]{badge2}` style from that PR, which we should probably also emulate).

But also, we know that we're going to change this some more, so it's fine to land as-is.

DJMcNab (approved):
See also https://github.com/linebender/color/ for the other essential things in a top-level README (i.e. MSRV, licensing, contribution). But those should probably come as a follow-up, to avoid muddling up reviews.

Thanks!

README.md
ajakubowicz-canva: Fixed the badge text `#simd` and the link (such that it navigates to the #simd channel).

Commit: https://github.com/linebender/fearless_simd/pull/35/commits/8c4620239c3d716dfc52d657381db022f9256656

New badge:
Screenshot 2025-07-15 at 1 20 08 pm

README.md
ajakubowicz-canva: TIL!

I've followed the pattern of the alts, and also applied the `\` pattern.
I've also re-ordered the badges to match the order in the linked PR.

PR #36, opened 2025-07-18
[ci] Add check to ensure code generation remains in sync

ajakubowicz-canva:
### Context

Add CI action "check generated code" which enforces that the code generator does not produce a `git diff` after running in CI.

### Test plan

This PR is test only.

### Additional changes

There are some other cleanup changes, that were informed by `cargo clippy -p fearless_simd_gen`.

### DIFF
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -1174,7 +1174,6 @@ impl Simd for WasmSimd128 {
         let (b0, _) = self.split_f32x8(b);
         self.combine_f32x4(self.zip_low_f32x4(a0, b0), self.zip_high_f32x4(a0, b0))
     }
-    // DO NOT COMMIT - test CI fails
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


fearless_simd_gen/src/main.rs
```diff
@@ -73,6 +76,32 @@ impl Module {
...
+            .spawn()
+            .expect("`rustfmt` should spawn");
+
+        let mut stdin = child.stdin.take().unwrap();
+        stdin
+            .write_all(
+                format!("// This file is autogenerated by fearless_simd_gen\n\n{code}").as_bytes(),
+            )
+            .unwrap();
+        drop(stdin);
```
LaurenzV: I'm not sure I understand why we need this?

LaurenzV: Is this not already generated as part of the auto-generation process?

LaurenzV (approved):
Lovely!

PR #37, opened 2025-07-18
[WASM] Add f64x2 operations and unzip

ajakubowicz-canva:
### Context

This is a WASM architecture followup to https://github.com/linebender/fearless_simd/pull/33. Implementing this PR removes all `todo!()` calls from the generated WASM file.

### How

 - Add `unzip`.
 - Add `f64x2` implementations.

I also cleaned up some other implementations to use generic methods instead of type-specific methods. For example we can express `fract` generically as `a.sub(a.trunc())` which then works on both `f32x4` and `f64x2`.

### Test plan

Added unit tests for full coverage of new methods.


LaurenzV:
Thanks!!

PR #38, opened 2025-07-18
Generalize wasm convert intrinsic lookup

Ralith:
No functional change, but it's a little more concise and should save some churn in the future.


ajakubowicz-canva (approved):
Nice! Thank you 😄

PR #39, opened 2025-07-23
Improve CI so it better matches Linebender standards

ajakubowicz-canva:
### Context

This was raised in: [#simd > Linebender Standards @ 💬](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Linebender.20Standards/near/529829126)

This PR uses [`Kurbo`'s](https://github.com/linebender/kurbo/blob/main/.github/workflows/ci.yml) and [Vello's](https://github.com/linebender/vello/blob/main/.github/workflows/ci.yml) `.github/ci.yml` files as templates.

This PR also adds the `copyright.sh` file. Note, I've modified this file such that the generated code is exempt from needing copyright headers (as currently implemented). If we want generated code to generate a copyright header we should definitely discuss that.

### Changes

Additionally I've raised https://github.com/linebender/fearless_simd/issues/40 to track burning down the Clippy `expect` attributes. These allow us to add blocking clippy CI and then burn them down.

### Test plan

Running CI 😄


fearless_simd/examples/unpremul.rs
ajakubowicz-canva, line 1: The changes in this file were required for tests to pass. I've commented them out – but could be convinced to delete it entirely (thanks to version control).

I think leaving this code here in a commented form is motivating for implementing f16 and seeing a usage.

DJMcNab:
(FYI, I'm not reviewing this yet - beyond the initial comments on Zulip - because it's still in draft. If you want me to have a look anyway, just let me know)

.github/workflows/ci.yml
```diff
@@ -1,11 +1,62 @@
...
+  # We aim to always test with the latest stable Rust toolchain, however we pin to a specific
+  # version like 1.70. Note that we only specify MAJOR.MINOR and not PATCH so that bugfixes still
+  # come automatically. If the version specified here is no longer the latest stable version,
+  # then please feel free to submit a PR that adjusts it along with the potential clippy fixes.
+  RUST_STABLE_VER: "1.88" # In quotes because otherwise (e.g.) 1.70 would be interpreted as 1.7
+  # The purpose of checking with the minimum supported Rust toolchain is to detect its staleness.
+  # If the compilation fails, then the version specified here needs to be bumped up to reality.
+  # Be sure to also update the rust-version property in the workspace Cargo.toml file,
+  # plus all the README.md files of the affected packages.
+  RUST_MIN_VER: "1.85"
```
DJMcNab, line 11: Not for this PR, but we should document this in the READMEs (we also do probably need crate-level READMEs...)

.github/workflows/ci.yml
```diff
@@ -26,6 +77,22 @@ jobs:
       - name: cargo fmt
         run: cargo fmt --all --check
 
+      - name: Install Taplo
+        uses: uncenter/setup-taplo@09968a8ae38d66ddd3d23802c44bf6122d7aa991 # v1
+        with:
+          version: "0.9.3"
+
+      - name: Run taplo fmt
+        run: taplo fmt --check --diff
```
DJMcNab, line 88: Nice!

.github/workflows/ci.yml
```diff
@@ -93,13 +155,15 @@ jobs:
...
       - name: cargo clippy (no_std)
-        run: cargo hack clippy -p fearless_simd --locked --target wasm32-unknown-unknown --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features std,default,half
+        run: cargo hack clippy -p fearless_simd --locked --target wasm32-unknown-unknown --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }}
 
       # TODO: Add back the suffix ` -- -D warnings` onto this command to make warnings errors.
       # Currently there are too many warnings to turn on this lint.
       - name: cargo clippy
         run: cargo hack clippy -p fearless_simd --locked --target wasm32-unknown-unknown --optional-deps --each-feature --ignore-unknown-features --features std
 
+      # TODO: Add (auxiliary) step.
```
DJMcNab: We probably need to document why we've not done that here. This could be just in the form of a reply here, but I'd prefer something in the code.

.github/workflows/ci.yml
```diff
@@ -93,13 +155,15 @@ jobs:
       # TODO: Add back the suffix ` -- -D warnings` onto this command to make warnings errors.
       # Currently there are too many warnings to turn on this lint.
       - name: cargo clippy (no_std)
-        run: cargo hack clippy -p fearless_simd --locked --target wasm32-unknown-unknown --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features std,default,half
+        run: cargo hack clippy -p fearless_simd --locked --target wasm32-unknown-unknown --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }}
```
DJMcNab: This should *probably* use the `NO_WASM_PACKAGES` technique used in other crates. That ensures that merges of future versions of the CI would be clean.

This *could* be deferred.

.github/workflows/ci.yml
```diff
@@ -125,10 +189,9 @@ jobs:
         run: git diff --exit-code
 
   test-stable:
-    name: test
+    name: cargo test
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false
```
DJMcNab, line 131: Fwiw, our concurrency story is something we need to address. I agree with removing this here for now. See also [#linebender > GitHub concurrency control](https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/GitHub.20concurrency.20control/with/493170477)

.github/workflows/ci.yml
```diff
@@ -174,23 +244,91 @@ jobs:
         run: RUSTFLAGS=-Ctarget-feature=+simd128 wasm-pack test --headless --chrome
         working-directory: fearless_simd_tests
```
DJMcNab, line 244: I agree with not changing things here; this probably does need to be hardcoded

.github/workflows/ci.yml
DJMcNab, line 1: We probably should also apply https://github.com/linebender/xilem/pull/775 here (and in Vello).
This however is also something which can be deferred. Ultimately, centralising where our CI lives is the long-term plan, which would fix a lot of these niggles

.github/copyright.sh
```diff
@@ -0,0 +1,23 @@
...
+	echo $output
+	echo -e "\n\nPlease add the following header:\n"
+	echo "// Copyright $(date +%Y) the Fearless_SIMD Authors"
+	echo "// SPDX-License-Identifier: Apache-2.0 OR MIT"
+	echo -e "\n... rest of the file ...\n"
+	exit 1
+fi
+
+echo "All files have correct copyright headers."
+exit 0
```
DJMcNab: Missing final newline

fearless_simd/examples/play.rs
DJMcNab, line 1: In theory, we should instead be adding the "safe_wrapper"s feature as a required feature for this example. But this can be deferred. The proposed solution is reasonable enough

fearless_simd/examples/unpremul.rs
DJMcNab: Maybe we should use `#[cfg(false)]` here instead (or well, the pre-1.88 MSRV-compatible `#[cfg(any())]`).

Either way, that could be a follow-up

fearless_simd/src/core_arch/x86_64/avx.rs
```diff
@@ -1,6 +1,11 @@
 // Copyright 2024 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+#![expect(
+    clippy::too_many_arguments,
+    reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
+)]
+
```
DJMcNab, line 8: This is surprising to me. `clippy::too_many_arguments` is already `allow` in the lint set...

Oh dear, this is really bad! The workspace crates don't actually have the lint set enabled - see e.g. https://github.com/linebender/fearless_simd/blob/main/fearless_simd/Cargo.toml; there's no
```
[lints]
workspace=true
```
(!)

We have to make a decision here. I think deferring that would be fine-ish, but we should do it as an immediate follow-up.

fearless_simd/src/core_arch/mod.rs
```diff
@@ -3,6 +3,22 @@
...
+        reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
+    )
+)]
+#![cfg_attr(
+    feature = "std",
+    expect(
+        clippy::missing_safety_doc,
+        reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
+    )
+)]
```
DJMcNab: Since there's no unique lint under `feature="std"`, we should instead do:

```suggestion
#![cfg_attr(
    not(feature = "std"),
    expect(
        clippy::too_many_arguments,
        reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
    )
)]
#![expect(
    clippy::missing_safety_doc,
    reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
)]
```
(with whatever formatting rustfmt does here).

Although missing safety doc is one which we should fix pretty immediately, ideally!

fearless_simd/src/half_assed.rs
```diff
@@ -1,8 +1,15 @@
...
-// Much of the code below is copy-pasted from half-rs by Kathryn Long, version 2.4.1.
+#![cfg_attr(
+    feature = "std",
+    expect(
+        clippy::missing_safety_doc,
+        reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
+    )
+)]
 
+// Much of the code below is copy-pasted from half-rs by Kathryn Long, version 2.4.1.
```
DJMcNab, line 12: It might be worth linking to e.g. https://github.com/VoidStarKat/half-rs/releases/tag/v2.4.1 here.
This is non-blocking, and doesn't need an issue.

DJMcNab (approved):
I agree that this is an improvement on the current state. The only thing I consider blocking is the TODO comment about auxiliary jobs on WASM.

This has however brought up a pretty massive follow-up job, which is that the Linebender lint set not actually being applied here(!).

Let me know when you're ready to merge this, and I'll do so. (And if you want another review of any changes you make, let me know).

The pending jobs are due to branch protection, and can be safely ignored[^0]

[^0]: The current "interim" version of branch protection being used assumes that the old CI setup (i.e. the one being removed in this PR) is active. Just before we merge this PR, whoever is merging it will swap the branch protection over. That does mean that one of the admins will need to do the change.

fearless_simd/src/core_arch/x86_64/avx.rs
ajakubowicz-canva: Amazing catch! I think this can be an immediate followup.

.github/copyright.sh
ajakubowicz-canva: Fixed: https://github.com/linebender/fearless_simd/pull/39/commits/3d9fac2246aeb756cfd28bc2f420c9fd8686b231#diff-fb1b9974afa7fe944b143b3f53fbc6b56f4f89f84516898da058ac95e9970215R23

.github/workflows/ci.yml
ajakubowicz-canva: Filed: https://github.com/linebender/fearless_simd/issues/44

.github/workflows/ci.yml
ajakubowicz-canva: Applied the `NO_WASM_PACKAGES` pattern: https://github.com/linebender/fearless_simd/pull/39/commits/3d9fac2246aeb756cfd28bc2f420c9fd8686b231#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR18

.github/workflows/ci.yml
ajakubowicz-canva: Fixed: https://github.com/linebender/fearless_simd/pull/39/commits/3d9fac2246aeb756cfd28bc2f420c9fd8686b231#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL165

fearless_simd/examples/unpremul.rs
ajakubowicz-canva: Went with `any()`: https://github.com/linebender/fearless_simd/pull/39/commits/3d9fac2246aeb756cfd28bc2f420c9fd8686b231#diff-1fad5e6bdde2245e8c56fd16899ab5fa0be8b757b7f83a04833cae9c5070ddc9R6

fearless_simd/src/half_assed.rs
ajakubowicz-canva: Done: https://github.com/linebender/fearless_simd/pull/39/commits/3d9fac2246aeb756cfd28bc2f420c9fd8686b231#diff-6eabc8e60c6c7bee214f2f7b5d51bac541b11fa9a23427cba4584e43b4b6f931R13

.github/workflows/ci.yml
ajakubowicz-canva: Tracking: https://github.com/linebender/fearless_simd/issues/45

ajakubowicz-canva:
@DJMcNab I noticed that not only had I missed the auxiliary step in Wasm, I had not made the Wasm clippy checks blocking 😬 . I've now fixed that and propagated the changes.

If you are happy to lead, I am happy for you to land this PR now. Once this PR lands I'll immediately followup with the workspace lints fix. Thank you thank you!

fearless_simd/examples/unpremul.rs
```diff
@@ -1,12 +1,12 @@
 // Copyright 2025 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use fearless_simd::{Level, Simd};
+use fearless_simd::Level;
```
DJMcNab, line 5: This needs a TODO comment of some kind. I'd also probably prefer these lines were of the form:
```rust
#[cfg(target_arch= ...)]
 // TODO: Our f16 support doesn't actually exist yet, but this is what it will look like when it does.
#[cfg(any())]
```

However, because this PR is special, I'll leave this unaddressed for now.

DJMcNab (approved):
To unblock future work, I'm going to merge this now despite my comment (as it's late here)

fearless_simd/examples/unpremul.rs
ajakubowicz-canva: 🙏 
Tracked in: https://github.com/linebender/fearless_simd/issues/40

fearless_simd/src/core_arch/x86_64/avx.rs
ajakubowicz-canva: Addressing in https://github.com/linebender/fearless_simd/pull/47

PR #42, opened 2025-07-26
SSE 4.2 support

Ralith:
Depends on #41.

Implementation is still incomplete, the output doesn't build, and there's probably some logic errors, so this is just opened for visibility for now. Feedback of course still welcome if I'm way off track somewhere.


Ralith:
Subsumed by https://github.com/linebender/fearless_simd/pull/50.

PR #43, opened 2025-07-28
Update README.md with current leevel support

sagudev:
I was confused by lying readme: https://github.com/linebender/vello/issues/1130#issuecomment-3127950405


ajakubowicz-canva (approved):
Thank you

PR #47, opened 2025-07-30
Add workspace lints

ajakubowicz-canva:
This PR follows the great observation that we were not applying lints in crates.
Here: https://github.com/linebender/fearless_simd/pull/39#discussion_r2241836153

This PR follows the same philosophy as https://github.com/linebender/fearless_simd/pull/39 which is to plug all lint errors with `expect` calls and TODO reasons.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,8 +3,9 @@ resolver = "2"
 members = ["fearless_simd", "fearless_simd_gen", "fearless_simd_tests"]
 
 [workspace.package]
-license = "MIT/Apache-2.0"
 edition = "2024"
+license = "Apache-2.0 OR MIT"
+repository = "https://github.com/linebender/vello"
 
 
 [workspace.lints]
--- a/fearless_simd/Cargo.toml
+++ b/fearless_simd/Cargo.toml
@@ -3,6 +3,7 @@ name = "fearless_simd"
 version = "0.2.0"
 license.workspace = true
 edition.workspace = true
+repository.workspace = true
 authors = ["Raph Levien <raph.levien@gmail.com>"]
 keywords = ["simd"]
 categories = ["hardware-support"]
@@ -15,6 +16,9 @@ safe_wrappers = []
 default = ["std"]
 std = []
 libm = ["dep:libm"]
+
+[lints]
+workspace=true
 
 [dependencies]
 bytemuck = "1.23.0"
--- a/fearless_simd/src/core_arch/aarch64/fp16.rs
+++ b/fearless_simd/src/core_arch/aarch64/fp16.rs
@@ -3,6 +3,14 @@
 
 //! Access to fp16 intrinsics on aarch64.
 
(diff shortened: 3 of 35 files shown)


fearless_simd/examples/sigmoid.rs
```diff
@@ -1,6 +1,12 @@
 // Copyright 2024 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+#![expect(
+    missing_docs,
+    clippy::shadow_unrelated,
```
DJMcNab: Apologies I didn't express this directly; we should also update the lint set to version 6.1 (https://linebender.org/wiki/canonical-lints/). The only relevant difference in that version is that it removes `shadow_unrelated` (see https://github.com/linebender/linebender.github.io/pull/90)

Luckily, we're already on v5, so this update only involves removing the expects for shadow_unrelated.

fearless_simd/examples/play.rs
```diff
@@ -1,6 +1,11 @@
 // Copyright 2024 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+#![expect(
+    missing_docs,
```
DJMcNab, line 5: FYI, for examples, `missing_docs` "should" only be complaining about the crate level docs being missing (i.e. a short explanation of the example). That's not to say those should be added in this PR, but more to make fixing these feel more tractable.

Cargo.toml
```diff
@@ -3,8 +3,9 @@ resolver = "2"
 members = ["fearless_simd", "fearless_simd_gen", "fearless_simd_tests"]
 
 [workspace.package]
-license = "MIT/Apache-2.0"
 edition = "2024"
+license = "Apache-2.0 OR MIT"
```
DJMcNab, line 12: Good catch!

Cargo.toml
```diff
@@ -3,8 +3,9 @@ resolver = "2"
...
+repository = "https://github.com/linebender/vello"
```
DJMcNab: Wrong repo!

DJMcNab (approved):
Looks good to me - a few small things to address.

Cargo.toml
ajakubowicz-canva: 🤦 🙃

ajakubowicz-canva: This is now fixed: https://github.com/linebender/fearless_simd/pull/47/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R13

fearless_simd/examples/sigmoid.rs
ajakubowicz-canva: Updated! Thank you :D
https://github.com/linebender/fearless_simd/pull/47/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R18

PR #48, opened 2025-07-31
Add a better test harness for `fearless_simd`

LaurenzV:


fearless_simd_dev_macros/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+        quote! {}
+    };
+
+    let wasm_snippet = if include_wasm {
+        quote! {
+            #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
+            #[wasm_bindgen_test::wasm_bindgen_test]
+            fn #wasm_name() {
+                let neon = unsafe { fearless_simd::wasm32::WasmSimd128::new_unchecked() };
+                #input_fn_name(neon);
```
ajakubowicz-canva: Super tiny nit: `neon` -> `wasm`

LaurenzV: Ah oops, yep. But overall, the design looks good to you? If so I can keep porting the remaining tests.

fearless_simd_dev_macros/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+#[allow(unused_variables, reason = "on purpose.")]
+fn exclude_fallback(name: &str) -> bool {
+    false
+}
+
+#[allow(dead_code, reason = "on purpose.")]
+#[allow(unused_variables, reason = "on purpose.")]
+fn exclude_wasm(name: &str) -> bool {
+    false
+}
```
ajakubowicz-canva: I'm fine either way, but it may be nicer to have the ignore be right on the test (passed via the attribute).

I am happy for it to remain like this, but maybe something like this?

`#[simd_test(skip_wasm = true, skip_neon = true)]` where the default omitted value is `false`? Definitely open to alternative.

ajakubowicz-canva (commented):
I love this direction so much! Thank you so much for actually doing the work to migrate tests over to something more idiomatic and something that has far better test coverage!

fearless_simd_dev_macros/src/lib.rs
LaurenzV: I had that thought too, but I felt like in the beginning it's easier to keep it that way, because otherwise when starting to develop a new backend (e.g. SSE/AVX), you have to put that attribute on every test method manually. While if you have this function, you can just do something like `if name contains u8x16`.

ajakubowicz-canva: That's a really great point.

fearless_simd_dev_macros/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+fn exclude_fallback(name: &str) -> bool {
+    false
+}
+
+#[allow(dead_code, reason = "on purpose.")]
+#[allow(unused_variables, reason = "on purpose.")]
+fn exclude_wasm(name: &str) -> bool {
+    matches!(
+        name,
+        "min_precise_f32x4_with_nan" | "max_precise_f32x4_with_nan"
```
ajakubowicz-canva, line 105: 😢

fearless_simd_tests/tests/harness.rs
```diff
@@ -0,0 +1,1099 @@
...
+#[simd_test]
+fn max_precise_f32x4_with_nan<S: Simd>(simd: S) {
+    let a = f32x4::from_slice(simd, &[f32::NAN, -3.0, f32::INFINITY, 0.5]);
+    let b = f32x4::from_slice(simd, &[1.0, f32::NAN, 7.0, f32::NEG_INFINITY]);
+    let result = a.max_precise(b).val;
+
+    assert_eq!(result[0], 1.0);
+    assert_eq!(result[1], -3.0);
+    assert_eq!(result[2], f32::INFINITY);
+    assert_eq!(result[3], 0.5);
```
ajakubowicz-canva, line 107: Doesn't need action, but is there a reason this is split up over 4 asserts?

fearless_simd_tests/tests/harness.rs
```diff
@@ -0,0 +1,1099 @@
...
+#[simd_test]
+fn floor_f64x2<S: Simd>(simd: S) {
+    let a = f64x2::from_slice(simd, &[1.7, -2.3]);
+    assert_eq!(a.floor().val, [1.0, -3.0]);
+}
+
+#[simd_test]
+fn fract_f64x2<S: Simd>(simd: S) {
+    let a = f64x2::from_slice(simd, &[1.7, -2.3]);
+    assert_eq!(a.fract().val, [0.7, -0.2999999999999998]);
```
ajakubowicz-canva, line 1092: Is there some floating point precision tolerance that we may need to introduce across architectures?

Non blocking - just curious.

ajakubowicz-canva (approved):
Look at all the [fallback and neon tests](https://github.com/linebender/fearless_simd/actions/runs/16653308583/job/47131611063?pr=48#step:6:209) being tested!
This is a HUGE improvement.

Thank you for fixing `load_interleaved`!

fearless_simd_tests/tests/harness.rs
LaurenzV: Ah, this was because before one of the numbers was `NaN`, so I used a `x.is_nan()` in one line instead. Using the trick of doing a `mem::transmute` didn't work because there were differences between the scalar and WASM version. But in this case, I can probably revert it. Good catch.

fearless_simd_tests/tests/harness.rs
LaurenzV: I guess we might soon find out. :D

PR #49, opened 2025-08-01
Add more tests for zip and unzip

LaurenzV:
Since those are a bit tricky to implement.


ajakubowicz-canva (approved):
Awesome! Approving although I assume you will end up closing this PR as this change has been mirrored in https://github.com/linebender/fearless_simd/pull/50

LaurenzV:
Eh, why not merge separately 😄

PR #50, opened 2025-08-01
Add some initial SSE4.2 support

LaurenzV:
This PR is based on #42 but contains a couple more function implementation / fixes. With this patch, we are able to completely render strips using SIMD (flattening works as well, thought there are some problems with precision that I still need to debug). And while there are still many functions missing, I think this is a good chunk of implementation that can be merged first before doing further implementation work,


fearless_simd_dev_macros/src/lib.rs
```diff
@@ -59,6 +65,19 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
         quote! {}
     };
 
+    let sse4_snippet = if include_sse4 {
+        quote! {
+            #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+            #[test]
+            fn #sse4_name() {
+                let neon = unsafe { fearless_simd::x86::Sse4_2::new_unchecked() };
+                #input_fn_name(neon);
```
ajakubowicz-canva: Nit (naming): neon

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -0,0 +1,4680 @@
...
+
+#![allow(unused_variables)]
+use crate::{Level, Simd, SimdFrom, SimdInto, seal::Seal};
+use crate::{
+    f32x4, f32x8, f32x16, f64x2, f64x4, f64x8, i8x16, i8x32, i8x64, i16x8, i16x16, i16x32, i32x4,
+    i32x8, i32x16, mask8x16, mask8x32, mask8x64, mask16x8, mask16x16, mask16x32, mask32x4,
+    mask32x8, mask32x16, mask64x2, mask64x4, mask64x8, u8x16, u8x32, u8x64, u16x8, u16x16, u16x32,
+    u32x4, u32x8, u32x16,
+};
+use core::arch::x86_64::*;
```
ajakubowicz-canva, line 14: Non blocking, but we may want to track in an issue and TODO?

This module is being used under both `x86` and `x86_64` checks. However, the generated code is only importing from `x86_64`. Should we be compile checking here and choosing between the two modules?

See this documentation: https://doc.rust-lang.org/core/arch/index.html#static-cpu-feature-detection

Cherry picking one example, [`_mm_set1_ps`](https://doc.rust-lang.org/core/arch/x86_64/fn._mm_set1_ps.html). It looks like the API works on _both_ `x86` and `x86_64` and is exported by both. I'm hoping that the only change might be conditionally importing:

```rs
#[cfg(target_arch = "x86")]
use std::arch::x86::*;
use std::arch::x86_64::*;
```

Then we get compile time guarantees that the APIs work for both target architectures. Maybe this is an easy change?

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -0,0 +1,570 @@
...
+}
+
+pub fn mk_sse4_2_impl() -> TokenStream {
+    let imports = type_imports();
+    let simd_impl = mk_simd_impl();
+    let ty_impl = mk_type_impl();
+
+    quote! {
+        // Until we have implemented all functions.
+        #![allow(unused_variables)]
```
ajakubowicz-canva: Prefer `expect`.

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -0,0 +1,570 @@
...
+                            fn #method_ident(self, a: #ty<Self>, b: #ty<Self>, c: #ty<Self>) -> #ret_ty {
+                                a + b * c
+                            }
+                        }
+                    }
+                    "msub" => {
+                        quote! {
+                            #[inline(always)]
+                            fn #method_ident(self, a: #ty<Self>, b: #ty<Self>, c: #ty<Self>) -> #ret_ty {
+                                a + b * c
```
ajakubowicz-canva: `a + b`?

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -0,0 +1,570 @@
...
+                        expr = quote! { #ident(#expr) }
+                    }
+
+                    quote! {
+                        #[inline(always)]
+                        fn #method_ident(self, a: #ty<Self>, b: #ty<Self>) -> #ret_ty {
+                            unsafe { #expr.simd_into(self) }
+                        }
+                    }
+                }
```
ajakubowicz-canva: For "less than or equal" or "greater than or equal" there are dedicated intrinsics like `_mm_cmple_ps`. Should those be used instead of `or`-ing together a `less than` and an `equals`?

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -0,0 +1,570 @@
...
+                            }
+                            .rust_name()
+                        );
+                        quote! {
+                            #[inline(always)]
+                            fn #method_ident(self, a: #ty<Self>) -> #ret_ty {
+                                unsafe {
+                                    let raw = a.into();
+                                    let high = #extend(raw).simd_into(self);
+                                    let low = #extend(_mm_slli_si128(raw, 8)).simd_into(self);
```
ajakubowicz-canva, line 195: Nit: May be worth documenting magic number `8`.

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -0,0 +1,570 @@
...
+                        _ => unreachable!(),
+                    };
+                    if scalar_bits == 8 {
+                        quote! {
+                            #[inline(always)]
+                            fn #method_ident(self, a: #ty<Self>, b: u32) -> #ret_ty {
+                                todo!()
+                            }
+                        }
+                        // let extend = extend_intrinsic(vec_ty.scalar, 8, 16);
```
ajakubowicz-canva, line 249: Could a "why is this commented out" be added? How is it wrong, or what needs to be done? Otherwise it can just be deleted.

ajakubowicz-canva (approved):
As is explicitly known from the PR, there are some known issues, but this is a huge step forward in implementing x86 in fearless_simd.

Unfortunately I don't have a device that can actually test this so I'm leaning on the passing tests and assuming that this was also tested a little with vello.

None of my comments need to be blocking as we're literally in the earliest stages of development except it does seem like the comparison operations could be made more efficient by reducing the `or`-ing of `<` and `=` and using a single `less-than-and-equals` intrinsic. I am not opposed that being done in a followup.

LaurenzV:
> except it does seem like the comparison operations could be made more efficient by reducing the or-ing of < and = and using a single less-than-and-equals intrinsic.

From what I can tell those actually don't exist on SSE4.2 for integer numbers, only for floats, that's the reason why I had to resort to that in the first place: https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#expand=1749&ig_expand=2066,986,4842,4826,4197,3079&text=_mm_cmpl&techs=SSE_ALL,AVX_ALL

fearless_simd/src/generated/sse4_2.rs
LaurenzV: I think I will open an issue for this, so that we can also add appropriate CI checks for this.

LaurenzV: Tracked in #51.

fearless_simd_gen/src/mk_sse4_2.rs
LaurenzV: true, we can use thos for floats, but for integers for some reaoson only lt/gt exists.

fearless_simd_gen/src/mk_sse4_2.rs
LaurenzV: I'm not a 100% sure because I didnt write that part, Iäll add a TODO.

fearless_simd_gen/src/mk_sse4_2.rs
LaurenzV: I'm also not sure here because I didn't write that. Let's keep it in there and ask Benjamin later.

PR #52, opened 2025-08-04
Add more SSE4.2 impls

LaurenzV:
This is enough to make the whole test suite pass with SSE4.2! There is still something wrong because performance is absolute trash on some benchmarks (I presume some functions are not being inlined properly), but at least the tests pass!

This also fixes an issue where we didn't use wrapping semantics in the scalar code (this also applies to some other arithmetic operations, but for now I've only fixed it for `mul`).

### DIFF
--- a/fearless_simd/src/generated/sse4_2.rs
+++ b/fearless_simd/src/generated/sse4_2.rs
@@ -3151,7 +3151,7 @@ impl Simd for Sse4_2 {
     #[inline(always)]
     fn store_interleaved_128_f32x16(self, a: f32x16<Self>, dest: &mut [f32; 16usize]) -> () {
         let fb = crate::Fallback::new();
-        fb.store_interleaved_128_f32x16(a.val.simd_into(fb), dest)
+        fb.store_interleaved_128_f32x16(a.val.simd_into(fb), dest);
     }
     #[inline(always)]
     fn reinterpret_u8_f32x16(self, a: f32x16<Self>) -> u8x64<Self> {
@@ -3463,7 +3463,7 @@ impl Simd for Sse4_2 {
     #[inline(always)]
     fn store_interleaved_128_u8x64(self, a: u8x64<Self>, dest: &mut [u8; 64usize]) -> () {
         let fb = crate::Fallback::new();
-        fb.store_interleaved_128_u8x64(a.val.simd_into(fb), dest)
+        fb.store_interleaved_128_u8x64(a.val.simd_into(fb), dest);
     }
     #[inline(always)]
     fn reinterpret_u32_u8x64(self, a: u8x64<Self>) -> u32x16<Self> {
@@ -3832,7 +3832,7 @@ impl Simd for Sse4_2 {
     #[inline(always)]
     fn store_interleaved_128_u16x32(self, a: u16x32<Self>, dest: &mut [u16; 32usize]) -> () {
         let fb = crate::Fallback::new();
-        fb.store_interleaved_128_u16x32(a.val.simd_into(fb), dest)
+        fb.store_interleaved_128_u16x32(a.val.simd_into(fb), dest);
     }
     #[inline(always)]
     fn narrow_u16x32(self, a: u16x32<Self>) -> u8x32<Self> {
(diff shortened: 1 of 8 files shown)


fearless_simd/src/generated/sse4_2.rs
```diff
@@ -210,11 +210,15 @@ impl Simd for Sse4_2 {
     }
     #[inline(always)]
     fn cvt_u32_f32x4(self, a: f32x4<Self>) -> u32x4<Self> {
-        unsafe { _mm_cvtps_epi32(_mm_floor_ps(a.into())).simd_into(self) }
+        unsafe {
+            _mm_cvtps_epi32(_mm_max_ps(_mm_floor_ps(a.into()), _mm_set1_ps(0.0))).simd_into(self)
+        }
```
LaurenzV, line 215: This is needed so that negative numbers get clamped to 0. This obviously still doesnt fix the issue for numbers > i32::MAX

fearless_simd_dev_macros/src/lib.rs
ajakubowicz-canva, line 1: Tests are really nice. I can't stress how happy and thankful I am that you implemented the test suite so well!

fearless_simd_tests/tests/harness.rs
ajakubowicz-canva, line 1: Could a `mul_f32x4` test be added? (I can't find one in the harness).

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -210,11 +210,15 @@ impl Simd for Sse4_2 {
...
     }
     #[inline(always)]
     fn cvt_i32_f32x4(self, a: f32x4<Self>) -> i32x4<Self> {
-        unsafe { _mm_cvtps_epi32(_mm_floor_ps(a.into())).simd_into(self) }
+        unsafe {
+            _mm_cvtps_epi32(_mm_max_ps(_mm_floor_ps(a.into()), _mm_set1_ps(0.0))).simd_into(self)
```
ajakubowicz-canva: Shouldn't `i32` be able to be converted to a negative float? E.g. `-1_i32` should remain `-1_f32`. It looks like this would instead convert `-1` to `0`. Could a test be added as well? Thank you!

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -489,10 +496,45 @@ fn mk_simd_impl() -> TokenStream {
...
+                                let out3 = _mm_unpackhi_epi64(tmp1, tmp3); // [3,7,11,15]
+
+                                self.combine_u32x8(
+                                    self.combine_u32x4(out0.simd_into(self), out1.simd_into(self)),
+                                    self.combine_u32x4(out2.simd_into(self), out3.simd_into(self)),
+                                )
+                            }
+                        }
+                    } else {
+                        quote! { crate::Fallback::new().#method_ident(src).val.simd_into(self) }
```
ajakubowicz-canva, line 568: Could a comment be added here that this is a stop-gap, and that we don't want to lean on fallback for sse4.2, unless for some reason we need to lean into scalar code?

This feedback applies to later usages of `crate::Fallback`. I think it's totally fine to use - but I'd like it to be marked as a stop-gap explicitly.

ajakubowicz-canva: Also - is there a perf hit because calling `crate::Fallback::new()` needs to construct a struct to then effectively call a pure function?
Not a blocker - just curious.

ajakubowicz-canva (approved):
A couple additional tests would be valuable for some of the changes. Otherwise another excellent change.


Is the reason for using `todo` in some places and `Fallback` in others so we get vello_cpu tests passing? We could use `Fallback` in all places so we technically have functionality, and then we have motivation to replace those locations with more optimal intrinsic usage.

Alternatively - it is also good to have slow paths fail loudly so we prioritise fixing them.

As always, great work!

fearless_simd_gen/src/mk_sse4_2.rs
Ralith: It's a ZST, so short of catastrophic inliner failure it'll get erased completely.

LaurenzV:
> We could use Fallback in all places so we technically have functionality, and then we have motivation to replace those locations with more optimal intrinsic usage.

Yeah for sure, I just really wanted to prioritize getting everything for vello_cpu to work.

fearless_simd/src/generated/sse4_2.rs
LaurenzV: You are right. 🤦This should only happen for u32, will add a test.

fearless_simd_gen/src/mk_sse4_2.rs
ajakubowicz-canva: Ah yes! Assuming ZST means "zero-sized type". Thank you for the clarification! :D

PR #53, opened 2025-08-06
Add x86 support on top of x86_64 and test it in CI

### LINKED ISSUE #51: Make sure that x86 SIMD also works on 32-bit
See: https://github.com/linebender/fearless_simd/pull/50#discussion_r2250184572

sagudev:
Fixes #51.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,10 +189,15 @@ jobs:
 
   test-stable:
     name: cargo test
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.platform.os }}
     strategy:
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        platform:
+          - { target: aarch64-apple-darwin, os: macos-latest }
+          - { target: x86_64-apple-darwin, os: macos-13 }
+          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
+          - { target: x86_64-pc-windows-msvc, os: windows-latest }
+          - { target: i686-pc-windows-msvc, os: windows-latest }
     steps:
       - uses: actions/checkout@v4
 
@@ -200,6 +205,7 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_STABLE_VER }}
+          targets: ${{ matrix.platform.target }}
 
       - name: install cargo-nextest
         uses: taiki-e/install-action@v2
@@ -212,10 +218,10 @@ jobs:
           save-if: ${{ github.event_name != 'merge_group' }}
 
       - name: cargo nextest
-        run: cargo nextest run --workspace --locked --all-features --no-fail-fast
+        run: cargo nextest run --workspace --locked --all-features --no-fail-fast --target ${{ matrix.platform.target }}
 
       - name: cargo test --doc
-        run: cargo test --doc --workspace --locked --all-features --no-fail-fast
+        run: cargo test --doc --workspace --locked --all-features --no-fail-fast --target ${{ matrix.platform.target }}
 
   test-stable-wasm:
     name: cargo test (wasm32)
--- a/fearless_simd/src/core_arch/x86/avx.rs
+++ b/fearless_simd/src/core_arch/x86/avx.rs
@@ -9,9 +9,14 @@
 //! Access to AVX intrinsics.
 
 use crate::impl_macros::delegate;
-use core::arch::x86_64::*;
+#[cfg(target_arch = "x86")]
+use core::arch::x86 as arch;
+#[cfg(target_arch = "x86_64")]
+use core::arch::x86_64 as arch;
 
-/// A token for AVX intrinsics on `x86_64`.
+use arch::*;
+
+/// A token for AVX intrinsics on `x86` and `x86_64`.
 #[derive(Clone, Copy, Debug)]
 pub struct Avx {
     _private: (),
@@ -28,7 +33,7 @@ impl Avx {
         Self { _private: () }
     }
 
-    delegate! { core::arch::x86_64:
+    delegate! { arch:
         fn _mm256_add_pd(a: __m256d, b: __m256d) -> __m256d;
         fn _mm256_add_ps(a: __m256, b: __m256) -> __m256;
         fn _mm256_and_pd(a: __m256d, b: __m256d) -> __m256d;
--- a/fearless_simd/src/generated/sse4_2.rs
+++ b/fearless_simd/src/generated/sse4_2.rs
@@ -15,12 +15,15 @@ use crate::{
     mask32x8, mask32x16, mask64x2, mask64x4, mask64x8, u8x16, u8x32, u8x64, u16x8, u16x16, u16x32,
     u32x4, u32x8, u32x16,
 };
+#[cfg(target_arch = "x86")]
+use core::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
 use core::arch::x86_64::*;
 use core::ops::*;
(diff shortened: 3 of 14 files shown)


.github/workflows/ci.yml
sagudev, line 1: That's how we do multitarget stuff in servo. Only windows has preinstalled stuff for compiling to x86.

fearless_simd/src/core_arch/x86/avx.rs
```diff
@@ -0,0 +1,336 @@
...
+    ///
+    /// # Safety
+    ///
+    /// The required CPU features must be available.
+    #[inline]
+    pub unsafe fn new_unchecked() -> Self {
+        Self { _private: () }
+    }
+
+    delegate! { core::arch::x86:
```
sagudev: This macro is the reason why we have separate x86 and x86_64 modules. The content of both modules is same, except for `x86` <-> `x86_64`.

sagudev (commented):
Some commentary. 

In servo we use result (sink) job, that depends on all previous jobs for required status check. This allows easier adding/removing/renaming jobs.

.github/workflows/ci.yml
ajakubowicz-canva, line 1: cc @DJMcNab  - CI changes for your review. I'll review the other changes.

LaurenzV:
I think we don't need to (presumably that's what happened) copy-paste all of the files in core_arch/x86, as we really only need the SIMD token from that, no? Or is there a reason why we have to copy the whole code again? My understanding was that the things in core_arch are basically deprecated.

sagudev:
> I think we don't need to (presumably that's what happened) copy-paste all of the files in core_arch/x86, as we really only need the SIMD token from that, no? Or is there a reason why we have to copy the whole code again? My understanding was that the things in core_arch are basically deprecated.

Indeed, I just copy pasted and fixed the code to make it compile. We need to duplicate it for delegate macro as noted in https://github.com/linebender/fearless_simd/pull/53#discussion_r2255988388. Not sure if we actually use it's implementations I am new to fearless simd.

fearless_simd/src/core_arch/x86/avx.rs
```diff
@@ -0,0 +1,336 @@
+// Copyright 2024 the Fearless_SIMD Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![expect(
+    clippy::too_many_arguments,
+    reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
+)]
```
DJMcNab: I'm surprised that this expect is required. We've removed too_many_arguments from the lint set, and so all expects of it should have been removed at the same time

DJMcNab (commented):
CI changes seem reasonable enough, as a special case for this foundational library. My one question is what is this in service of?
I'm not keen for us to drag legacy x86 support up through our full stack in service of a small handful of users. Maybe I'm underestimating how many x86 users there are?

fearless_simd/src/core_arch/x86/avx.rs
sagudev: IDK, that is copy-paste from x86_64, so this is something that needs to be resolved outside of this PR.

sagudev:
> My one question is what is this in service of?
> I'm not keen for us to drag legacy x86 support up through our full stack in service of a small handful of users.

This is foundational library and support for x86 is cheap when you already have x86_64.

.github/workflows/ci.yml
```diff
@@ -189,17 +189,23 @@ jobs:
 
   test-stable:
     name: cargo test
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.platform.os }}
     strategy:
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        platform:
+          - { target: aarch64-apple-darwin, os: macos-14 }
```
DJMcNab: Why not macos-15 or macos-latest?

DJMcNab:
I've also prepared the new branch protection this will need.

.github/workflows/ci.yml
sagudev: Because I copypasted this form servo, but will fix. btw macos-latest == macos-14.

DJMcNab: Sure, but not for too much longer.

sagudev: > Sure, but not for too much longer.

Indeed, but in servo we prefer to hardcode runner images, because updates usually broke something, at least in more complex (ahm C++) libraries we have.

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -15,11 +15,17 @@ use crate::{
...
+#[cfg(target_arch = "x86_64")]
 use core::arch::x86_64::*;
 use core::ops::*;
 #[doc = r#" The SIMD token for the "SSE 4.2" level."#]
 #[derive(Clone, Copy, Debug)]
 pub struct Sse4_2 {
+    #[cfg(target_arch = "x86")]
+    pub sse4_2: crate::core_arch::x86::Sse4_2,
+    #[cfg(target_arch = "x86_64")]
     pub sse4_2: crate::core_arch::x86_64::Sse4_2,
```
ajakubowicz-canva: Thank you for an awesome PR!

Instead of duplicating the code between `x86` and `x86_64`, what do you think about handling the difference between the two architectures within a single universal `x86` module?

For example, instead of duplicating `sse4_2` under both `x86` and `x86_64`, is it possible to instead do something like the following (to pass different architectures to the delegate macro):

```rust
// Conditionally import the right architecture aliasing import to `arch`.

#[cfg(target_arch = "x86")]
use core::arch::x86 as arch;
use core::arch::x86_64 as arch;

use arch::*;

// ... truncated code ...
impl Sse4_2 {
    // ...

    // Call delegate on `arch`. Goal of this suggestion is reducing duplication.
    delegate! { arch:
        fn _mm_cmpistrm<const IMM8: i32>(a: __m128i, b: __m128i) -> __m128i;
    // ...
}
```


Is this possible? Are there any advantages to the code duplication and separation by architecture? It does look like there may be some differences in API between 32bit and 64 bit, but we could probably conditionally compile those differences as well if required.

sagudev: This is great idea and now done https://github.com/linebender/fearless_simd/pull/53/commits/564a5253a0e049157487a2d1fe487299eff2459b.

ajakubowicz-canva (approved):
This is so good! Thank you!

ajakubowicz-canva:
Just checking since I am not up to speed on the newest processes of fearless_simd – are you able to merge this or do you need me to click "merge". Thank you!

sagudev:
I can only enable auto-merge, because of CI changes we need different branch rules.

DJMcNab:
Should be fixed now. I think you might need to remove and re-add

.github/workflows/ci.yml
```diff
@@ -189,17 +189,23 @@ jobs:
...
   test-stable:
     name: cargo test
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.platform.os }}
     strategy:
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
+        platform:
+          - { target: aarch64-apple-darwin, os: macos-latest }
+          - { target: x86_64-apple-darwin, os: macos-13 }
```
DJMcNab, line 197: What's your plan in ~a month's time when this runner is removed? https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down

sagudev: My response: https://github.com/linebender/fearless_simd/pull/60

PR #58, opened 2025-08-07
Add cursory docs

DJMcNab:
Includes some standards improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ For now, prefer updating the package-level readmes, e.g. fearless_simd/README.md
 [![Apache 2.0 or MIT license.](https://img.shields.io/badge/license-Apache--2.0_OR_MIT-blue.svg)](#license)
 \
 [![Linebender Zulip, #simd channel.](https://img.shields.io/badge/Linebender-%23simd-blue?logo=Zulip)](https://xi.zulipchat.com/#narrow/channel/514230-simd)
-[![GitHub Actions CI status.](https://github.com/linebender/fearless_simd/workflows/CI/badge.svg)](https://github.com/linebender/fearless_simd/actions)
+[![GitHub Actions CI status.](https://img.shields.io/github/actions/workflow/status/linebender/fearless_simd/ci.yml?logo=github&label=CI)](https://github.com/linebender/fearless_simd/actions)
 [![Dependency staleness status.](https://deps.rs/repo/github/linebender/fearless_simd/status.svg)](https://deps.rs/repo/github/linebender/fearless_simd)
 
 </div>
--- a/README.md
+++ b/README.md
@@ -31,8 +31,7 @@ For now, prefer updating the package-level readmes, e.g. fearless_simd/README.md
 
 ## Motivation
 
-<!-- TODO: I don't think we use Target Feature 1.1 - our MSRV is 1.85, when TF1.1 stabilised in 1.86 -->
-This crate proposes an experimental way to use SIMD intrinsics reasonably safely, using the new [target_feature 1.1] feature in Rust, recently stabilized.
+This crate proposes an experimental way to use SIMD intrinsics reasonably safely.
 The blog post [A plan for SIMD] contains the high level motivations, goal, and summary for Fearless SIMD.
 
 ## History
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ For now, prefer updating the package-level readmes, e.g. fearless_simd/README.md
 **Safer and easier SIMD**
 
 [![Latest published version.](https://img.shields.io/crates/v/fearless_simd.svg)](https://crates.io/crates/fearless_simd)
-[![Documentation build status.](https://docs.rs/fearless_simd/badge.svg)](https://docs.rs/fearless_simd)
+[![Documentation build status.](https://img.shields.io/docsrs/fearless_simd.svg)](https://docs.rs/fearless_simd)
 [![Apache 2.0 or MIT license.](https://img.shields.io/badge/license-Apache--2.0_OR_MIT-blue.svg)](#license)
 \
 [![Linebender Zulip, #simd channel.](https://img.shields.io/badge/Linebender-%23simd-blue?logo=Zulip)](https://xi.zulipchat.com/#narrow/channel/514230-simd)
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ For now, prefer updating the package-level readmes, e.g. fearless_simd/README.md
 
 ## Motivation
 
+<!-- TODO: I don't think we use Target Feature 1.1 - our MSRV is 1.85, when TF1.1 stabilised in 1.86 -->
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -31,6 +31,7 @@ For now, prefer updating the package-level readmes, e.g. fearless_simd/README.md
 
 ## Motivation
 
+<!-- TODO: I don't think we use Target Feature 1.1 - our MSRV is 1.85, when TF1.1 stabilised in 1.86 -->
```
raphlinus: Yes, this is right. I'm not sure how that got in there.

fearless_simd/Cargo.toml
```diff
@@ -11,16 +11,33 @@ description = "Safer and easier SIMD"
...
 default = ["std"]
+# Get floating point functions from the standard library (likely using your targets libc)
 std = []
+# Use floating point implementations from libm
 libm = ["dep:libm"]
 
+# Include safe wrappers for (some) target feature specific intrinsics,
+# beyond the basic SIMD operations abstracted on all platforms
+safe_wrappers = []
+# Enable `f16` (16 bit floating point) support. Only supported on aarch64, as other
```
raphlinus, line 31: The `f16` support that's here is vestigial, and is based on a userland type (either a dependency on the half crate or our own). We've decided to wait for proper `f16` support to land in Rust, which I think will be fairly soon. Probably best to rip all the f16 stuff out, which maybe should be done in a separate PR.

ajakubowicz-canva:
[quick update] I have not forgot this but got a bit busy. If no one else reviews I'll have it done by end-of-workday tomorrow (AEST).

fearless_simd/src/lib.rs
```diff
@@ -96,6 +231,14 @@ impl Level {
...
         }
     }
+
+    /// If this is a proof that SSE4.2 (or better) is available, access that instruction set.
+    ///
+    /// This method should be preferred over matching against the `Sse4_2` variant of self,
+    /// because if Fearless SIMD gets support for an instruction set which is a superset of SSE4.2,
+    /// this method will return a value even if that "better" instruction set is available.
+    ///
+    /// This is primarily intended to be used if the `safe_wrappers` feature is enabled.
```
raphlinus: I'm wondering if this is overly specific. It absolutely makes sense to downcast to a SIMD level and use intrinsics unsafely or semi-safely (by which I basically mean unsafe transition to target_feature then using the combination of tf 1.1 and safe intrinsics under tf).

I'd probably write this, "this can be used in combination with the `safe_wrappers` feature to gain checked access to the level-specific SIMD capabilities."

fearless_simd/src/lib.rs
```diff
@@ -105,11 +248,30 @@ impl Level {
...
+    /// explicitly vectorise a function. That has a better developer experience
+    /// than an implementation of `WithSimd`, and is less likely to miss a vectorisation
+    /// opportunity.
+    ///
+    /// This has two use cases:
+    /// 1) To call a manually written implementation of [`WithSimd`].
+    /// 2) To ask the compiler to autovectorise scalar code.
+    ///
+    /// For the second case to work, the provided function *must* be attributed with `#[inline(always)]`.
+    /// Note also that any calls that function makes to other functions will likely not be autovectoised,
```
raphlinus: "autovectorised". There may be other instances.

fearless_simd/README.md
```diff
@@ -21,10 +21,101 @@ Full documentation at https://github.com/orium/cargo-rdme -->
...
 
+Fearless SIMD exposes safe SIMD with ergonomic multi-versioning in Rust.
+
+Fearless SIMD uses "marker values" which serve as proofs of which target features are available on the current CPU.
+These each implement the [`Simd`] trait, which exposes a core set of SIMD operations which are implemented as
+efficiently as possible on each target platform.
+
+Additionally, there are types for packed vectors of a specific width and element type (such as [`f32x4`]).
+Fearless SIMD does not currently support vectors of less than 128 bits, due to there only being limited hardware
+with SIMD support but not support for 128 bit wide vectors. <!-- TODO: confirm -->
```
raphlinus: I would cut the "due to" clause. Support for 64 bit is pretty widespread, and use cases do pop up (for example, f32 -> f16 conversion on neon naturally produces a 64 bit value). We don't implement every possible case because there are just so many of them.

fearless_simd/README.md
```diff
@@ -21,10 +21,101 @@ Full documentation at https://github.com/orium/cargo-rdme -->
...
+# Kernels vs not kernels
+
+TODO: Talk about writing versions of functions which can be called in other `S: Simd` functions.
+I think this pattern can also have a macro.
+-->
+
+# Webassembly
+
+WASM SIMD doesn't have feature detection, and so you need to compile two versions of your bundle for WASM, one with SIMD and one without,
+then select the appropriate one for your user's browser.P
```
raphlinus: There's a stray "P" here.

fearless_simd/README.md
```diff
@@ -21,10 +21,101 @@ Full documentation at https://github.com/orium/cargo-rdme -->
...
+TODO: Expand on this.
+
+# Feature Flags
+
+The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
+
+- `std` (enabled by default): Get floating point functions from the standard library (likely using your targets libc).
```
raphlinus: "target's"

fearless_simd/README.md
```diff
@@ -21,10 +21,101 @@ Full documentation at https://github.com/orium/cargo-rdme -->
...
+- `libm`: Use floating point implementations from [libm].
+- `safe_wrappers`: Include safe wrappers for (some) target feature specific intrinsics,
+  beyond the basic SIMD operations abstracted on all platforms.
+- `half`: Use `f16` (16 bit floating point) support from the [half] crate.
```
raphlinus, line 111: Delete all the f16 related stuff for now.

raphlinus (approved):
Looks good, thanks! There's a bunch of minor stuff inline.

fearless_simd/README.md
DJMcNab: Removed - but should we have a tracking issue for 64 bit wide vectors?

fearless_simd/Cargo.toml
DJMcNab: I'll leave it for another PR

fearless_simd/README.md
DJMcNab: As in https://github.com/linebender/fearless_simd/pull/58#discussion_r2269224670, I'll leave it for another PR. It's best to keep these docs up-to-date with the state of the code, even though we plan to immediately remove this.

fearless_simd/src/lib.rs
DJMcNab: I would love if could have a safe (macro-assisted) transition to `target_feature` based on these types. I don't think it's possible without a macro, but the macro should be relatively trivial. We do need to be careful not to make it possible to call a true unsafe function in safe code.

PR #59, opened 2025-08-07
Add basic structure for AVX2

LaurenzV:
Currently doesn't actually use any AVX2 intrsincs, except for FMA (in all other cases it falls back to SSE4.2), but this seems like a good start to then incrementally extend it.


fearless_simd_gen/src/arch/x86_common.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright 2024 the Fearless_SIMD Authors
```
ajakubowicz-canva:
```suggestion
// Copyright 2025 the Fearless_SIMD Authors
```

fearless_simd_gen/src/arch/avx2.rs
```diff
@@ -0,0 +1,18 @@
...
+pub(crate) struct Avx2;
+
+impl Arch for Avx2 {
+    fn arch_ty(&self, ty: &VecType) -> TokenStream {
+        x86_common::arch_ty(ty)
+    }
+
+    fn expr(&self, op: &str, ty: &VecType, args: &[TokenStream]) -> TokenStream {
+        x86_common::expr(op, ty, args)
+    }
```
ajakubowicz-canva, line 17: Is it worth adding a TODO that these are falling back to sse4.2 instructions which must be present if Avx2 is present, but we want to adopt those Avx2 instructions where it makes sense?

fearless_simd_gen/src/mk_avx2.rs
```diff
@@ -0,0 +1,229 @@
...
+            type mask16s = mask16x8<Self>;
+            type mask32s = mask32x4<Self>;
+            #[inline(always)]
+            fn level(self) -> Level {
+                Level::#level_tok(self)
+            }
+
+            #[inline]
+            fn vectorize<F: FnOnce() -> R, R>(self, f: F) -> R {
+                #[target_feature(enable = "avx2,fma")]
```
ajakubowicz-canva, line 125: Out of interest, as `avx2` and `fma` should only be present if `sse4.2` exists (to my understanding), does this mean that the presence of `avx2,fma` implicitly assumes `sse4.2` is enabled?

fearless_simd_gen/src/x86_common.rs
```diff
@@ -0,0 +1,107 @@
+// Copyright 2024 the Fearless_SIMD Authors
```
ajakubowicz-canva:
```suggestion
// Copyright 2025 the Fearless_SIMD Authors
```

fearless_simd_gen/src/arch/sse4_2.rs
```diff
@@ -1,173 +1,18 @@
...
-            }
-        }
-    }
-}
-
-pub(crate) fn op_suffix(mut ty: ScalarType, bits: usize, sign_aware: bool) -> &'static str {
-    use ScalarType::*;
-    if !sign_aware && ty == Unsigned {
-        ty = Int;
-    }
```
ajakubowicz-canva, line 111: This was copied over, and as I was reviewing it I realized it reads a bit confusing. It sort of reads like: "if not sign aware then convert unsigned to signed".

No change required.

ajakubowicz-canva (approved):
Lets go! Good stuff!

ajakubowicz-canva:
Thank you for setting up the foundation.

fearless_simd_gen/src/mk_avx2.rs
LaurenzV: Yes, I'm pretty sure it does.

fearless_simd_gen/src/arch/avx2.rs
LaurenzV: I think this makes more sense to add in `mk_simd`then in this file, because x86_common should atcually be generic over the SIMD width.

PR #62, opened 2025-08-18
Fix match statement for `no_std` build

LaurenzV:
In a previous PR the `Level` enum was marked as non-exhaustive, but we forgot to add the corresponding branch in one of the `simd_dispatch` macros. We probably should configure CI so this is caught in the future.


DJMcNab (approved):
The change seems fine, but I would like to understand why this didn't get caught by CI.
If this is blocking someone, of course we can land before it's fixed/understood. But if not, I'd consider leaving this open as a reminder.

Separately, this std/no_std difference seems a bit odd. Like yes, we can't detect the CPU features on no_std, but someone can still manually work out what features are available (e.g. by knowing what embedded CPU they are using). This isn't a priority, but might be worth an issue.

LaurenzV:
I've opened #65 and #66 to track both issues.

PR #68, opened 2025-08-21
Address the periodic lints, in preparation for v0.2.0

DJMcNab:
See https://linebender.org/wiki/canonical-lints/

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,10 @@
+# LINEBENDER LINT SET - .clippy.toml - v1
+# See https://linebender.org/wiki/canonical-lints/
+
+# The default Clippy value is capped at 8 bytes, which was chosen to improve performance on 32-bit.
+# Given that we are building for the future and even low-end mobile phones have 64-bit CPUs,
+# it makes sense to optimize for 64-bit and accept the performance hits on 32-bit.
+# 16 bytes is the number of bytes that fits into two 64-bit CPU registers.
+trivial-copy-size-limit = 16
+
+# END LINEBENDER LINT SET


fearless_simd/src/core_arch/aarch64/mod.rs
```diff
@@ -2138,8 +2138,8 @@ impl Neon {
         fn vdup_n_p8(value: p8) -> poly8x8_t;
         fn vdup_n_p16(value: p16) -> poly16x4_t;
         fn vdup_n_f32(value: f32) -> float32x2_t;
-        fn vldrq_p128(a: *const p128) -> p128;
-        fn vstrq_p128(a: *mut p128, b: p128);
+        unsafe fn vldrq_p128(a: *const p128) -> p128;
+        unsafe fn vstrq_p128(a: *mut p128, b: p128);
```
DJMcNab, line 2142: If you dig into the implementations, it's trivial that these are unsafe

fearless_simd/src/core_arch/x86/sse2.rs
```diff
@@ -34,8 +34,7 @@ impl Sse2 {
 
     delegate! { arch:
         fn _mm_pause();
-        #[allow(clippy::not_unsafe_ptr_arg_deref)]
-        fn _mm_clflush(p: *const u8);
+        unsafe fn _mm_clflush(p: *const u8);
```
DJMcNab, line 37: It's not entirely clear if this should be unsafe, but [there is evidence](https://www.felixcloutier.com/x86/clflush) that this can page fault.

LaurenzV (approved):

PR #69, opened 2025-08-21
Fix soundness of `simd_dispatch` using target-features 1.1

DJMcNab:
Previously, you would be able to write an unsafe function, and then call it using `simd_dispatch` without needing the `unsafe` token at the call site. E.g.

```rust
/// # Safety
///
/// `value` must be `Some`
#[inline(always)]
unsafe fn evil<S: Simd>(simd: S, value: Option<u32>) -> u32 {
     value.unwrap_unchecked()
}

simd_dispatch!(disguised(level, value: Option<u32>) -> u32);

fn unwitting(){
    disguised(None);
}
```

This is a real soundness hole, if `fearless_simd` expands to wider use, as it would effectively mean that external crates are unable to write unsafe "simd-compatible" functions. 

The fix is to require the `unsafe` token in the macro definition in these cases; we detect a true unsafe case without this unsafe block by using target feature 1.1.

In adding support for `unsafe`, I've also added a requirement that each function definition created in `simd_dispatch` is prefixed with the `fn` token. This isn't fundamental, but seems worth requiring.

This does bump up our MSRV to 1.86, for target feature 1.1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ env:
   # If the compilation fails, then the version specified here needs to be bumped up to reality.
   # Be sure to also update the rust-version property in the workspace Cargo.toml file,
   # plus all the README.md files of the affected packages.
-  RUST_MIN_VER: "1.85"
+  RUST_MIN_VER: "1.86"
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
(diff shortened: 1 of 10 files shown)


raphlinus (approved):
Good call. I was worried about this kind of soundness hole but didn't rigorously work out whether it was possible to exploit. It makes sense to leverage tf 1.1 for this; originally I was under the impression it was not needed at all.

Issue #66, opened 2025-08-19
Do not require `std` for using the various SIMD levels


LaurenzV:
See https://github.com/linebender/fearless_simd/pull/62#pullrequestreview-3130928346

PR #70, opened 2025-08-21
Add true support for `no_std` libraries

DJMcNab:
Fixes https://github.com/linebender/fearless_simd/issues/66


fearless_simd/src/lib.rs
```diff
@@ -175,15 +177,42 @@ impl Level {
...
         }
-        #[cfg(all(target_arch = "wasm32", target_feature = "simd128"))]
-        return Level::WasmSimd128(WasmSimd128::new_unchecked());
-        #[cfg(all(feature = "std", any(target_arch = "x86", target_arch = "x86_64")))]
+        #[cfg(target_arch = "wasm32")]
+        {
+            // WASM always either has the SIMD feature compiled in or not.
+            #[cfg(target_feature = "simd128")]
+            return Level::WasmSimd128(WasmSimd128::new_unchecked());
+            #[cfg(not(target_feature = "simd128"))]
```
LaurenzV, line 212: Couldn't we remove `#[cfg(not(target_arch = "wasm32"))]` below and remove the `return` for `#[cfg(not(target_feature = "simd128"))]`  here?

DJMcNab: That would have warnings on `wasm32+simd128` . I think having internally "unconditionally return on wasm32, then have a fallback on non-wasm32", is better than "return on simd128 + wasm32, and then fallback on ((not wasm32) or simd128)"

Issue #54, opened 2025-08-06
Tests are running twice


LaurenzV:
Haven't looked into this yet, but it seems like the tests are being run twice. See also for example this CI run: https://github.com/linebender/fearless_simd/actions/runs/16769236455/job/47480495686

PR #71, opened 2025-08-26
chore: de-dupe tests

ajakubowicz-canva:
Fixes https://github.com/linebender/fearless_simd/issues/54

### Why

tl;dr: Each rust file in `tests/` becomes an integration test entry point. Because we import `harness` from `mod.rs` we run the tests twice. This new structure means we only enter via `mod.rs` and run tests once.

See [Submodules in Integration Tests](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests).

### Test

I tested by running a single duplicated test with `cargo nextest run unzip_low_u32x4_neon --workspace --locked --all-features --no-fail-fast --target aarch64-apple-darwin`

Before:

```sh
────────────
 Nextest run ID af179010-cb8a-4c82-bead-7d35b42513c0 with nextest profile: default
    Starting 2 tests across 5 binaries (449 tests skipped)
        PASS [   0.007s] fearless_simd_tests::harness unzip_low_u32x4_neon
        PASS [   0.007s] fearless_simd_tests::tests harness::unzip_low_u32x4_neon
────────────
     Summary [   0.011s] 2 tests run: 2 passed, 449 skipped
```

After:

```sh
────────────
 Nextest run ID d7c6944c-8458-4dd6-abef-fcfa98ee016b with nextest profile: default
    Starting 1 test across 4 binaries (226 tests skipped)
        PASS [   0.012s] fearless_simd_tests::tests harness::unzip_low_u32x4_neon
────────────
     Summary [   0.017s] 1 test run: 1 passed, 226 skipped
```


LaurenzV:
Ah right. But maybe we can just disable the default test harness like in vello_sparse_tests? I just remember that not adding it to mod.rs can cause some trouble if we want to import functions from other files later on. But not sure.

LaurenzV (approved):

LaurenzV:
But just putting the contents of `harness.rs` into `mod.rs` is also fine for me.

PR #75, opened 2025-08-28
Introduce int-from-mask conversion

Ralith:
I have some logic that needs a -1 if a comparison passes and a 0 otherwise, which makes this very convenient. Not sure if transmute is precisely the idiomatic approach here, but I figure it works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ You can find its changes [documented below](#020-2025-08-26).
 
 This release has an [MSRV][] of 1.86.
 
+## Added
+
+- `SimdInt::from_mask` allows construction of an integer vector from
+  the associated mask type. ([#75][] by [@Ralith][])
+
 ## Fixed
 
 - `Simd` now requires consistent mask types for native-width
--- a/fearless_simd/src/generated/simd_trait.rs
+++ b/fearless_simd/src/generated/simd_trait.rs
@@ -920,6 +920,11 @@ pub trait SimdInt<Element: SimdElement, S: Simd>:
     fn to_float<T: SimdCvtFloat<Self>>(self) -> T {
         T::float_from(self)
     }
+    #[doc = r" Construct the integer vector with the same representation as `mask`"]
+    #[doc = r""]
+    #[doc = r" Unset lanes become 0. Set lanes become -1 or the maximum value for signed or"]
+    #[doc = r" unsigned element types respectively."]
+    fn from_mask(mask: Self::Mask) -> Self;
     fn simd_eq(self, rhs: impl SimdInto<Self, S>) -> Self::Mask;
     fn simd_lt(self, rhs: impl SimdInto<Self, S>) -> Self::Mask;
     fn simd_le(self, rhs: impl SimdInto<Self, S>) -> Self::Mask;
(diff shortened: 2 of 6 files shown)


DJMcNab (commented):
I'm not going to suggest that we block on this, but I'd strongly encourage adding documentation to this method. I know that its peers do not, but if we're adding a new method, it seems good to document it. Short-Medium Term, we do want all of the methods on `Simd` to be documented.

Ralith:
Added docs.

Ralith:
> You can even use bytemuck to do this safely

You cannot, because `Simd`'s associated types do not satisfy the necessary traits.

Ralith:
It doesn't even seem to be possible with `unsafe`:

```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
  --> src/grid.rs:66:13
   |
66 |             mem::transmute::<<S::f32s as SimdBase<f32, S>>::Mask, S::i32s>(x0.simd_ge(y0))
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: source type: `<<S as fearless_simd::Simd>::f32s as SimdBase<f32, S>>::Mask` (this type does not have a fixed size)
   = note: target type: `<S as fearless_simd::Simd>::i32s` (this type does not have a fixed size)
```

DJMcNab:
> You cannot, because `Simd`'s associated types do not satisfy the necessary traits.

Its fields are public though, so you can just do something like:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b257c378c6c5beeaa1d6a83dd021c90e

Anyway, I'm not really well placed to know what functionality is necessary here. Thanks for adding docs!

ajakubowicz-canva (approved):
I like the improved ergonomics and ran into this myself. Is there any downside to providing this method?

conditional-LGTM on: please add at least 1 test and a changelog entry to the root CHANGELOG file.

Ralith:
I added a commit which refines associated types on `Simd` to enforce sensible masks. This is required to allow the result of a comparison on `f32s` to be converted to `i32s`, for example. Tests still TODO.

CHANGELOG.md
```diff
@@ -15,6 +15,15 @@ You can find its changes [documented below](#020-2025-08-26).
...
 
+## Added
+
+- `SimdInt::from_mask` allows construction of an integer vector from
+  the associated mask type
+
+## Fixed
+
+- `Simd` now requires consistent mask types for native-width vectors
+
```
DJMcNab: Nice to see things added to the CHANGELOG. These should however use our usual changelog formatting, i.e.:

```suggestion
### Added

- `SimdInt::from_mask` allows construction of an integer vector from
  the associated mask type. ([#75][] by [@Ralith][])

### Fixed

- `Simd` now requires consistent mask types for native-width vectors. ([#75][] by [@Ralith][])

```
(Plus replacing the comments in the already indicated place at the bottom with the correct links)

DJMcNab (approved):

Ralith:
Thanks!

> follow the usual Linebender policies

Where can I read up on these?

PR #76, opened 2025-09-01
Introduce SimdBase::witness

Ralith:
This is convenient for when you have a generic vector and haven't explicitly passed plumbed through the `Simd`.


ajakubowicz-canva:
I think I understand this PR, but it may help me to see a concrete example (even if it's very trivial).
Thank you!

Ralith:
So far I've only needed this to call `SimdBase::splat` explicitly, either to initialize an accumulator or for arguments to `Select::select`. This is usually not necessary due to the broadcast operator overloads, and might be improved by e.g. tweaks to `Select`, but I can imagine other situations, e.g. returning a hard-coded default value. I suspect it will remain a useful escape hatch indefinitely.

Excerpt from a 4D simplex noise implementation:

```rs
    #[inline(always)]
    fn get<S: Simd>(&self, [x, y, z, w]: [S::f32s; 4]) -> ([[S::i32s; 4]; 5], [[S::f32s; 4]; 5]) {
        // ...
        let mut rank_x = S::i32s::splat(x.witness(), 0);
        let _1 = S::i32s::splat(x.witness(), 1);
        let _0 = S::i32s::splat(x.witness(), 0);
        let cond = x0.simd_gt(y0);
        rank_x = rank_x + cond.select(_1, _0);
       // ...
    }
```

ajakubowicz-canva (approved):
LGTM! Just needs changelog. Thank you!

PR #78, opened 2025-09-01
Implement ops::Shr<u32> for vectors

Ralith:
Changelog pending other merges.


Ralith:
Folded this into a commit in #79.

PR #79, opened 2025-09-01
Introduce shift right by vector

Ralith:
Not sure about the non-x86 cases yet, and I still need a polyfill for the 8-bit case on x86. Suggestions welcome!

### DIFF
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -232,6 +232,10 @@ impl Simd for WasmSimd128 {
         i8x16_shr(a.into(), shift).simd_into(self)
     }
     #[inline(always)]
+    fn shrv_i8x16(self, a: i8x16<Self>, b: i8x16<Self>) -> i8x16<Self> {
+        core::array::from_fn(|i| core::ops::Shr::shr(a.val[i], b.val[i])).simd_into(self)
+    }
+    #[inline(always)]
     fn simd_eq_i8x16(self, a: i8x16<Self>, b: i8x16<Self>) -> mask8x16<Self> {
         i8x16_eq(a.into(), b.into()).simd_into(self)
     }
@@ -345,6 +349,10 @@ impl Simd for WasmSimd128 {
     #[inline(always)]
     fn shr_u8x16(self, a: u8x16<Self>, shift: u32) -> u8x16<Self> {
         u8x16_shr(a.into(), shift).simd_into(self)
+    }
+    #[inline(always)]
+    fn shrv_u8x16(self, a: u8x16<Self>, b: u8x16<Self>) -> u8x16<Self> {
+        core::array::from_fn(|i| core::ops::Shr::shr(a.val[i], b.val[i])).simd_into(self)
     }
     #[inline(always)]
     fn simd_eq_u8x16(self, a: u8x16<Self>, b: u8x16<Self>) -> mask8x16<Self> {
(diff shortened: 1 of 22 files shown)


Ralith:
Added scalar fallbacks for 8-bit x86 and all wasm, and finished neon.

Ralith:
Should be good now. I noticed no tests failed when I accidentally used an AVX2 intrinsic in SSE4.2; we probably need to improve CI there.

fearless_simd_tests/tests/harness/mod.rs
```diff
@@ -906,33 +906,51 @@ fn shr_u8x16<S: Simd>(simd: S) {
...
 
 #[simd_test]
 fn shr_u32x4<S: Simd>(simd: S) {
     let a = u32x4::from_slice(simd, &[u32::MAX, 2147483648, 65536, 256]);
-    assert_eq!(a.shr(8).val, [16777215, 8388608, 256, 1]);
+    assert_eq!((a >> 8).val, [16777215, 8388608, 256, 1]);
+}
+
+#[simd_test]
+fn shrv_u32x4<S: Simd>(simd: S) {
```
LaurenzV, line 948: Perhaps we can add another test where each element is shifted by a different amount?

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -233,7 +233,25 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                     }
                 }
                 OpSig::Binary => {
-                    let args = [quote! { a.into() }, quote! { b.into() }];
+                    let args = match method {
+                        "shrv" if vec_ty.scalar == ScalarType::Int => {
+                            let neg = simple_intrinsic("vneg", vec_ty);
+                            [quote! { a.into() }, quote! { #neg(b.into()) }]
+                        }
```
LaurenzV: Why exactly are we just negating here? Wouldn't a negative shift mean a positive shift to the left instead?

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -233,7 +233,25 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                     }
                 }
                 OpSig::Binary => {
-                    let args = [quote! { a.into() }, quote! { b.into() }];
+                    let args = match method {
+                        "shrv" if vec_ty.scalar == ScalarType::Int => {
```
LaurenzV: I'm wondering whether we can do one match arm with "shrv" and then put the if condition inside of that?

fearless_simd_gen/src/mk_neon.rs
LaurenzV: Ah nevermind, it's because we are using `vshlq` right?

fearless_simd/src/generated/wasm.rs
```diff
@@ -739,6 +755,10 @@ impl Simd for WasmSimd128 {
         i32x4_shr(a.into(), shift).simd_into(self)
     }
     #[inline(always)]
+    fn shrv_i32x4(self, a: i32x4<Self>, b: i32x4<Self>) -> i32x4<Self> {
+        core::array::from_fn(|i| core::ops::Shr::shr(a.val[i], b.val[i])).simd_into(self)
```
LaurenzV, line 759: Interesting, I wonder whether we could use that pattern for the generated code for fallback as well. We should probably first make sure that it doesn't result in performance regressions though.

fearless_simd_gen/src/mk_neon.rs
Ralith: Yeah. As far as I can tell, neon doesn't have a dedicated right shift intrinsic.

Ralith:
> I'm not sure if there is a way to force a machine that has SSE4 but no AVX2.

I think the strategy would be a separate build that configures cargo/rustc to compile for an older x86 target. It might also be possible to explicitly disable the specific target feature, but that's certainly a less well trodden path.

fearless_simd/src/generated/wasm.rs
Ralith: Yeah, ideally long term this generator code path gets shared with fallback outright.

fearless_simd_gen/src/mk_neon.rs
Ralith: I usually prefer guards where possible to reduce rightward drift, but it's not a strong preference.

PR #81, opened 2025-09-06
Enable bitcasting between native-width vectors

Ralith:
Maybe #75 should be reverted in favor of this?


fearless_simd_tests/tests/harness/mod.rs
```diff
@@ -1335,3 +1335,12 @@ fn select_native_width_vectors<S: Simd>(simd: S) {
     let result_i16 = mask_u16.select(a_i16, b_i16);
     assert_eq!(result_i16.as_slice(), vec![50i16; S::i16s::N]);
 }
+
+#[simd_test]
+fn bitcast_native<S: Simd>(simd: S) {
+    let a_i32 = S::i32s::from_slice(simd, &vec![-1; S::i32s::N]);
```
DJMcNab, line 1369: Is this `vec!` even needed (i.e. can we create this as a const array instead).

Obviously for test code it doesn't matter, but it seems somewhat important for user code.

Ralith: Yes, it's necessary:

```
error: constant expression depends on a generic parameter
    --> fearless_simd_tests/tests/harness/mod.rs:1341:49
     |
1341 |     let a_i32 = S::i32s::from_slice(simd, &[-1; S::i32s::N]);
     |                                                 ^^^^^^^^^^
     |
     = note: this may fail depending on what value the parameter takes
```

User code will presumably have a slice to begin with, or be splatting or similar. Failing all else, you can always have a const assert that N isn't bigger than you assume.

PR #85, opened 2025-09-11
Use wrapping integer addition in Fallback

Ralith:
Got bit by this in a unit test.

### DIFF
--- a/fearless_simd_tests/tests/harness/mod.rs
+++ b/fearless_simd_tests/tests/harness/mod.rs
@@ -1348,3 +1348,11 @@ fn bitcast_native<S: Simd>(simd: S) {
         &vec![u32::MAX; S::i32s::N]
     );
 }
+
+#[simd_test]
+fn wrapping_add_u32<S: Simd>(simd: S) {
+    assert_eq!(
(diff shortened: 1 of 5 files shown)


fearless_simd_tests/tests/harness/mod.rs
```diff
@@ -1372,3 +1372,11 @@ fn bitcast_native<S: Simd>(simd: S) {
         &vec![u32::MAX; S::i32s::N]
     );
 }
+
+#[simd_test]
+fn wrapping_add_u32<S: Simd>(simd: S) {
+    assert_eq!(
+        (S::u32s::splat(simd, u32::MAX) + 1).as_slice(),
+        &vec![0; S::u32s::N]
```
LaurenzV, line 1356: Is it not possible to use a normal slice here? Because of generic consts or?

Ralith: This is a normal slice. Array literals can't be dependent on a generic parameter, unfortunately.

LaurenzV: I meant instead of vector, but yeah makes sense.

PR #86, opened 2025-09-12
Introduce left shift by u32

Ralith:


fearless_simd/src/generated/simd_trait.rs
```diff
@@ -117,6 +117,7 @@ pub trait Simd: Sized + Clone + Copy + Send + Sync + Seal + 'static {
     fn xor_i8x16(self, a: i8x16<Self>, b: i8x16<Self>) -> i8x16<Self>;
     fn shr_i8x16(self, a: i8x16<Self>, shift: u32) -> i8x16<Self>;
     fn shrv_i8x16(self, a: i8x16<Self>, b: i8x16<Self>) -> i8x16<Self>;
+    fn shl_i8x16(self, a: i8x16<Self>, shift: u32) -> i8x16<Self>;
```
DJMcNab, line 120: Do we know what our planned strategy for docs on methods on the `Simd` trait is?
If there's no infra for adding this already in place, then I don't suggest adding docs in this PR. But we really should get it set up so that at least when we add new features, they have docs by default

Ralith: I think we need to extend the templating infrastructure in ops.rs to encompass documentation.

PR #91, opened 2025-09-13
Introduce unary integer negation

Ralith:


fearless_simd_tests/tests/harness/mod.rs
```diff
@@ -1142,6 +1142,14 @@ fn neg_f64x2<S: Simd>(simd: S) {
...
 }
 
+#[simd_test]
+fn neg_i32x4<S: Simd>(simd: S) {
+    assert_eq!(
+        (-i32x4::from_slice(simd, &[1, 2, 3, 4])).val,
+        [-1, -2, -3, -4]
+    );
+}
+
```
LaurenzV: Would be great to have another test that at least partially contains negative numbers in the beginning.

Ralith: Good call. I don't think it's important for this to be a separate case, though.

PR #96, opened 2025-09-25
Expression-oriented dispatch macro

Ralith:
The current dispatch situation kind of sucks: no approach lets you close over the environment, `simd_dispatch` can't handle generics, and `WithSimd` requires boilerplate. If I want to, say, call a `Simd`-generic method with several arguments and other generic parameters, I need to jump through a bunch of hoops. Can we do better?

This experiment replicates a closure once for each available level, then invokes the closure through a `#[target_feature]` trampoline. The closure is not technically labeled with the target feature (that would require https://github.com/rust-lang/rust/issues/15701), but much like a `WithSimd::with_simd`  implementation, a closure that is called exactly once seems to in practice get inlined into the trampoline and result in the expected codegen. Type-checking is weird because each copy of the closure has a different argument type (Rust doesn't have generic closures or `impl Trait` in `let` yet), so an incautious developer might e.g. write NEON-specific code and not notice a problem until the first time they compile for x86.

Still, this is IMO a big ergonomic win, and not hard to use correctly if you just use it to delegate to a proper `Simd`-generic function/method -- which now may take any form whatsoever.


Ralith:
> [cargo clippy (wasm32)](https://github.com/linebender/fearless_simd/actions/runs/17998933576/job/51203697091?pr=96#logs)
> error: function `launder` is never used

Is `target_feature = "simd128"` not available in the wasm32 CI environment?

fearless_simd/README.md
```diff
@@ -55,36 +55,29 @@ as possible using SIMD instructions.
...
 A few things to note:
 
-1) This is generic over any `Simd` type.
-2) The [`simd_dispatch`] macro is used to create a multi-versioned version of the given function.
-3) The `_impl` suffix is used by convention to indicate the version of a function which will be dispatched to.
-4) The `impl` function *must* be `#[inline(always)]`.
-   The performance of the SIMD implementation will be poor if that isn't the case. See [the section on inlining for details](#inlining)
+1) `sigmoid` is generic over any `Simd` type.
+2) The [`dispatch`] macro is used to create a multi-versioned closure inlining the given function.
+3) The function or closure passed to [`dispatch!()`] *must* be `#[inline(always)]`.
```
ajakubowicz-canva: > *must* be `#[inline(always)]`

Can this rule be enforced? Is there ever a situation where someone might want to `inline(never)`?

fearless_simd/src/macros.rs
```diff
@@ -112,3 +113,68 @@ macro_rules! simd_dispatch {
...
+/// ```
+///
+/// [`Level`]: crate::Level
+/// [`Level::new()`]: crate::Level::new
+#[macro_export]
+macro_rules! dispatch {
+    ($level:expr, $fn:expr) => {{
+        fn launder<S: $crate::Simd>(x: S) -> impl $crate::Simd {
+            x
+        }
```
ajakubowicz-canva, line 154: Nit: Should `launder` also be `inline(always)`?

fearless_simd/README.md
DJMcNab: Effectively, the internal call to `vectorize` is `inline(never)`, so a second level of never inlining would be purely overhead.

fearless_simd/src/lib.rs
```diff
@@ -196,7 +189,7 @@ impl Level {
     /// you should construct the relevant variants yourself, using whatever
     /// way your specific chip supports accessing the current level.
     ///
-    /// This value will be passed to functions generated using [`simd_dispatch`].
+    /// This value will be passed to functions generated using [`dispatch!()`].
```
DJMcNab: Nit: `dispatch` doesn't generate functions

fearless_simd/src/lib.rs
```diff
@@ -17,36 +17,29 @@
...
 //! ```rust
-//! use fearless_simd::{Simd, simd_dispatch};
+//! use fearless_simd::{Level, Simd, dispatch};
 //!
 //! #[inline(always)]
-//! fn sigmoid_impl<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) { /* ... */ }
+//! fn sigmoid<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) { /* ... */ }
 //!
-//! simd_dispatch!(fn sigmoid(level, x: &[f32], out: &mut [f32]) = sigmoid_impl);
+//! dispatch!(Level::new(), { #[inline(always)] |simd| sigmoid(simd, &[/*...*/], &mut [/*...*/]) });
```
DJMcNab, line 29: How do you feel about a syntax like:
```rust
dispatch!(Level::new(), simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
```

The syntax being a closure doesn't actually contribute anything here, since it isn't the same closure being passed to each variant anyway...
The single difference would be that you can't pass a function with the signature `fn something<S: Simd>(simd: S) -> whatever`. But really, who cares about that use case; if the function has arguments, you need to use this massively more complex form.

I'll propose this as a follow-up PR to this, when we land it.

fearless_simd/src/lib.rs
```diff
@@ -17,36 +17,29 @@
...
 //!
 //! A few things to note:
 //!
-//! 1) This is generic over any `Simd` type.
-//! 2) The [`simd_dispatch`] macro is used to create a multi-versioned version of the given function.
-//! 3) The `_impl` suffix is used by convention to indicate the version of a function which will be dispatched to.
-//! 4) The `impl` function *must* be `#[inline(always)]`.
-//!    The performance of the SIMD implementation will be poor if that isn't the case. See [the section on inlining for details](#inlining)
+//! 1) `sigmoid` is generic over any `Simd` type.
+//! 2) The [`dispatch`] macro is used to create a multi-versioned closure inlining the given function.
```
DJMcNab: Nit: I'm pretty sure `dispatch!` doesn't create a closure, and instead immediately executes the function

DJMcNab (approved):
A couple docs nits, but otherwise this looks good. Thanks!

fearless_simd/README.md
Ralith: I'll reword this to "should". The reasoning is that, if it doesn't get inlined, then it won't benefit from the target features enabled by the caller. However, in my testing it seems to be inlined in practice even without the annotation (probably due to having exactly one caller), so it seems to be a matter of *guaranteeing* good behavior rather than relying on heuristics, and a failure would only be a performance issue anyway.

fearless_simd/src/macros.rs
Ralith: I'd be shocked if a generic identity function on a ZST ever appeared in optimized output, but I suppose it can't hurt.

fearless_simd/src/lib.rs
Ralith: It does as an implementation detail, but you're right, this was misleading. Reworded to focus on observable behavior.

fearless_simd/src/lib.rs
Ralith: I waffled on this for a while. It comes down to the slight friction of a custom syntax, versus the slight friction of defensively applying `#[inline(always)]` in the common literal closure case. The difference is small and I can't think of any use cases that are more difficult either way, so I don't feel strongly, and my inclination is to merge as-is and approve your follow-up.

PR #99, opened 2025-09-30
Improve the syntax of `dispatch`

DJMcNab:
Also does some drive-by docs work

Follow-up to #96

### DIFF
--- a/fearless_simd/README.md
+++ b/fearless_simd/README.md
@@ -64,7 +64,10 @@ use fearless_simd::{Level, Simd, dispatch};
 #[inline(always)]
 fn sigmoid<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) { /* ... */ }
 
-dispatch!(Level::new(), { #[inline(always)] |simd| sigmoid(simd, &[/*...*/], &mut [/*...*/]) });
+// The stored level, which is constructed once for your application.
+let level = Level::new();
+
+dispatch!(Level::new(), simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
 ```
 
 A few things to note:
--- a/fearless_simd/README.md
+++ b/fearless_simd/README.md
@@ -64,10 +64,10 @@ use fearless_simd::{Level, Simd, dispatch};
 #[inline(always)]
 fn sigmoid<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) { /* ... */ }
 
-// The stored level, which is constructed once for your application.
+// The stored level, which you should only construct once in your application.
 let level = Level::new();
 
-dispatch!(Level::new(), simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
+dispatch!(level, simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
 ```
 
 A few things to note:
--- a/fearless_simd/src/macros.rs
+++ b/fearless_simd/src/macros.rs
@@ -133,18 +133,7 @@ macro_rules! simd_dispatch {
 /// To guarantee target-feature-specific code generation, any functions called within the operation should
 /// be `#[inline(always)]`.
 ///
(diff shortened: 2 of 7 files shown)


fearless_simd/src/macros.rs
```diff
@@ -130,8 +140,10 @@ macro_rules! simd_dispatch {
 ///
 /// #[inline(always)]
 /// fn work<S: Simd>(simd: S) { /* ... */ }
+/// // The SIMD level for this run, created once in your application/library's lifecycle.
+/// let level = Level::new();
 ///
-/// dispatch!(Level::new(), work);
+/// dispatch!(level, simd => work(simd));
```
Ralith: nit: Since `dispatch` no longer accepts a function expression, I think this example isn't adding anything over the next one. Let's cut it.

fearless_simd/src/macros.rs
```diff
@@ -140,50 +152,75 @@ macro_rules! simd_dispatch {
...
+                )
+            }
             _ => unreachable!(),
         }
     }};
 }
 
 #[cfg(test)]
-#[allow(
+// This expect now also validates that we haven't missed any levels!
```
Ralith: Ooh, nice!

Ralith (approved):

fearless_simd/src/macros.rs
Ralith: Actually, maybe we could support both forms simultaneously? I think the syntax might be unambiguous...

DJMcNab:
While I remember (for tomorrow), I need to add a note about `return/?` not being supported (easily; for the former you can use `ControlFlow`, and the latter you probably need to manually disambiguate the `E` type)

fearless_simd/src/macros.rs
DJMcNab: We *could*. I really think that zero-argument functions are sufficiently rare that it's not meaningfully worth optimising for.

Ralith: Yeah, fair. It'd be a lot more to document and might confuse readers when used.

fearless_simd/README.md
```diff
@@ -64,7 +64,10 @@ use fearless_simd::{Level, Simd, dispatch};
 #[inline(always)]
 fn sigmoid<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) { /* ... */ }
 
-dispatch!(Level::new(), { #[inline(always)] |simd| sigmoid(simd, &[/*...*/], &mut [/*...*/]) });
+// The stored level, which is constructed once for your application.
+let level = Level::new();
+
+dispatch!(Level::new(), simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
```
ajakubowicz-canva: Is the intent that `level` should be passed here?

fearless_simd/src/macros.rs
```diff
@@ -140,50 +152,75 @@ macro_rules! simd_dispatch {
...
         fn launder<S: $crate::Simd>(x: S) -> impl $crate::Simd {
             x
         }
 
         match $level {
-            $crate::Level::Fallback(fb) => ($fn)(launder(fb)),
+            $crate::Level::Fallback(fb) => {
+                let $simd = launder(fb);
+                // This vectorize call does nothing, but it is reasonable to be consistent here.
+                fb.vectorize(
```
ajakubowicz-canva, line 173: Maybe I am misunderstanding, but should the result of `launder` be used in the changes below?
```suggestion
                let $simd = launder(fb);
                // This vectorize call does nothing, but it is reasonable to be consistent here.
                $simd.vectorize(
```

Ralith: It doesn't matter if it is or not, because the witness type is `Copy`.

ajakubowicz-canva: Ah nice! Thank you for explaining!

Issue #90, opened 2025-09-12
Half sentence in overview docs missing


inkreasing:
The last sentence in the Inlining section in the main docs is incomplete.
https://github.com/linebender/fearless_simd/blob/2a1d6aa7a59627100b3c72181bb6921f4b160adf/fearless_simd/src/lib.rs#L56

Would fix it myself, but i am not sure what is missing and what should be said.

PR #101, opened 2025-10-03
Remove sentence fragment

DJMcNab:
Fixes https://github.com/linebender/fearless_simd/issues/90

The fragment was trying to point to
https://doc.rust-lang.org/reference/attributes/codegen.html#r-attributes.codegen.target_feature.inline

But it was entirely backwards in that case anyway.

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -48,8 +48,7 @@
 //!
 //! Fearless SIMD relies heavily on Rust's inlining support to create functions which have the
 //! given target features enabled.
-//! As such, most functions which you write when using Fearless SIMD should have the `#[inline(always)]` attribute.
-//! This is required because in LLVM, functions with different target features cannot.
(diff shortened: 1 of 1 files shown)


fearless_simd/src/lib.rs
```diff
@@ -48,8 +48,7 @@
 //!
 //! Fearless SIMD relies heavily on Rust's inlining support to create functions which have the
 //! given target features enabled.
-//! As such, most functions which you write when using Fearless SIMD should have the `#[inline(always)]` attribute.
-//! This is required because in LLVM, functions with different target features cannot.
+//! As such, most functions which you write when using Fearless SIMD should have the `#[inline(always)]` attribute..
```
LaurenzV: Are the two . on purpose?

LaurenzV (approved):

fearless_simd/src/lib.rs
DJMcNab: No :)

Issue #92, opened 2025-09-14
Put SIMD levels behind feature flags


LaurenzV:
While the main purpose of this crate is SIMD, it would be nice if there were an option to disable it so that only the "Fallback" level is compiled (e.g. by adding a simd feature that is enabled by default but can be disabled). Ideally, this will also allow us to add a `forbid(unsafe_code)` attribute in case that feature is disabled.

The use case would be crates who want to allow their users to optionally disable SIMD for 1) binary size purposes and 2) safety reasons. Although the disadvantage is that as long as one crate in the dependency chain has the `simd` feature activated, it will enable it for all other crates as well.

Shnatsel:
Done in #258

PR #105, opened 2025-10-06
Handle a non-`Fallback` baseline

### LINKED ISSUE #74: Avoid compiling `Level::Fallback` if the compilation target requires our target features
[#simd > No Fallback on WASM SIMD](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/No.20Fallback.20on.20WASM.20SIMD/with/536366688)

There are a couple of API design consequences here. The main one is that we need to have an alternative to `Level::fallback`, such as `Level::definitely_available` (name TBD).

DJMcNab:
Fixes #74 

A couple of notes:
1) For ease, I removed `simd_dispatch` in this PR. As such, I don't want to land this before [#simd > v0.3.0](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/v0.2E3.2E0/with/543274721)
2) The cfg variants in here are gnarly; I believe that's unavoidable to express what we want. We do need to be clear that we are doing runtime dispatch properly
3) The automated testing story of this PR is an open question. I have a manual test, in the new `sh ./check_targets.sh` at the root. I want to defer running this in CI; I suspect it would be very expensive.

This does not address #92. This does however add a `force_support_fallback` feature, so it doesn't completely remove the fallback functionality.


DJMcNab:
Still outstanding is the changelog. I'm planning to leave this until after we get v0.3.0 out the door.

fearless_simd/src/lib.rs
```diff
@@ -297,16 +350,81 @@ impl Level {
...
+            unreachable_patterns,
+            reason = "On machines which statically support `avx2`, there is only one variant."
+        )]
         match self {
             Level::Avx2(avx2) => Some(avx2),
             _ => None,
         }
     }
 
+    /// Get the weakest statically supported SIMD level.
```
Ralith:
```suggestion
    /// Get the strongest statically supported SIMD level.
```
The weakest statically supported is always `Fallback`.

CHANGELOG.md
```diff
@@ -13,11 +13,24 @@ You can find its changes [documented below](#020-2025-10-10).
...
+  of tests.
+
+### Removed
+
+- Breaking change: The (deprecated) `simd_dispatch!` macro. ([#105][] by [@DJMcNab][])
 
 ## [0.3.0][] (2025-10-10)
 
-This release has an [MSRV][] of 1.86.
+This release has an [MSRV][] of 1.88.
```
ajakubowicz-canva: Was it intended to change this? E.g. I expect 0.3.0 continues to have an MSRV of 1.86?

fearless_simd/src/macros.rs
```diff
@@ -194,7 +78,10 @@ macro_rules! dispatch {
                     || $op,
                 )
             }
-            #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+            #[cfg(all(
+                any(target_arch = "x86", target_arch = "x86_64"),
+                not(all(target_feature = "avx2", target_feature = "fma"))
```
ajakubowicz-canva, line 84: Just to clarify - this line ensures that if `avx2` and `fma` are not both detected we want to fall through to the Avx2 case?

fearless_simd/src/macros.rs
```diff
@@ -212,11 +99,62 @@ macro_rules! dispatch {
...
+            }
             _ => unreachable!(),
         }
     }};
 }
 
+// This macro turns whether the `force_support_fallback` macro is enabled into a boolean literal
+// in `dispatch`, which allows it to be used correctly cross-crate.
+// This trickery is required because macros are expanded in the context of the calling crate, including for
+// evaluating `cfg`s.
```
ajakubowicz-canva, line 135: This pattern is so nice!

fearless_simd/src/lib.rs
```diff
@@ -297,16 +351,85 @@ impl Level {
...
+    /// Note that in these cases, the library should clearly inform the integrator
+    /// that it is using a fallback and so not getting optimal performance (e.g. by panicking if
+    /// `debug_assertions` are enabled, and emitting a log with the "error" level otherwise).
+    /// The messages given should also provide actionable fixes, such as pointing to the
+    /// entry-point which provides a `Level`, or your `"std"` feature.
+    ///
+    /// Note that this is unaffected by the `force-support-fallback` feature.
+    /// Instead, you should use [`Level::fallback`] if you require the fallback level.
+    pub const fn baseline() -> Self {
+        // TODO: How do we possibly test that this method works in all cases?
```
ajakubowicz-canva, line 393: Agreed! But from carefully reading through this PR the various cfg cases _look_ correct to me.

fearless_simd/src/lib.rs
```diff
@@ -281,6 +333,8 @@ impl Level {
     #[inline]
     pub fn as_sse4_2(self) -> Option<Sse4_2> {
         match self {
+            // TODO: Level::Avx2(avx2) => Some(...)
+            #[cfg(not(all(target_feature = "avx2", target_feature = "fma")))]
             Level::Sse4_2(sse42) => Some(sse42),
```
ajakubowicz-canva: Can you elaborate on the TODO? (just in the comment) 
Can it be deleted?

ajakubowicz-canva (approved):
This looks fantastic - and reading through this, I really like it.

As a side note, this is extremely hard to test haha. I ran your script and also poked around the compiled result but I cannot statically determine if all code paths are behaving as expected. However, I think that this is overall a great improvement.

Thank you!


Edit: I am least confident with the x86 changes. Those might be good candidates for a double check from someone else.

fearless_simd/src/macros.rs
DJMcNab: This says "if our target doesn't *statically* support [both `avx2` and `fma`], then we might be on a machine which only has SSE4.2 support. As such, we need to check if the currently detected Level is SSE4.2."

That is, we only need to even support the SSE4.2 level if we don't have support for the better level.

This is more closely documented on `Level` - I'll add a comment pointing there.

fearless_simd/src/macros.rs
DJMcNab: Yeah - I thought I was going to have to make the macro be truly awful for it to work, but the "inversion of control" paradigm luckily worked perfectly here.

fearless_simd/src/lib.rs
DJMcNab: I've added a pointer to `check_targets.sh` as well.

fearless_simd/src/lib.rs
DJMcNab: I've fixed the issue. I didn't want to write a justification here whilst #108 was still in flight, but that's now addressed.

DJMcNab:
I'm going to merge this now; I've pushed a few minor changes in 69ee7f07478573ece8f4311df134a65303c1e5fc (plus a fixup in adcb7f18783c6e1f1b1642972276dac361e6b241), but I don't think those should block forward progress here.

PR #107, opened 2025-10-10
Prepare v0.3.0 release

raphlinus:
This updates the changelog and version numbers in preparation for a release of version 0.3.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,16 +8,16 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 # Changelog
 
-The latest published Fearless SIMD release is [0.2.0](#020-2025-08-26) which was released on 2025-08-26.
-You can find its changes [documented below](#020-2025-08-26).
+The latest published Fearless SIMD release is [0.3.0](#020-2025-10-10) which was released on 2025-10-10.
+You can find its changes [documented below](#020-2025-10-10).
 
 ## [Unreleased]
 
-## [0.2.0][] (2025-08-26)
-
 This release has an [MSRV][] of 1.86.
 
-## [0.3.0][] (2025-10-09)
+## [0.3.0][] (2025-10-10)
+
+This release has an [MSRV][] of 1.86.
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -13,9 +13,13 @@ You can find its changes [documented below](#020-2025-08-26).
 
 ## [Unreleased]
 
+## [0.2.0][] (2025-08-26)
+
```
DJMcNab:
```suggestion
```

This seems to be a copy-paste error

CHANGELOG.md
```diff
@@ -13,9 +13,13 @@ You can find its changes [documented below](#020-2025-08-26).
...
 This release has an [MSRV][] of 1.86.
 
-## Added
+## [0.3.0][] (2025-10-09)
+
```
DJMcNab, line 21: The MSRV line from unreleased needs to be copied under this release.

CHANGELOG.md
```diff
@@ -13,9 +13,13 @@ You can find its changes [documented below](#020-2025-08-26).
 
 ## [Unreleased]
 
+## [0.2.0][] (2025-08-26)
+
 This release has an [MSRV][] of 1.86.
 
-## Added
+## [0.3.0][] (2025-10-09)
```
DJMcNab: There needs to be a corresponding link definition to the tag at the bottom of the changelog

CHANGELOG.md
DJMcNab, line 1: The paragraph at the top of this file needs to be updated,  to link to the new release (in both places, and the date needs updating).

(We realise that this is annoying, and do have plans to make this simpler, but we haven't applied them yet.)

fearless_simd_gen/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fearless_simd_gen"
-version = "0.2.0"
+version = "0.3.0"
```
DJMcNab, line 3: I'm not sure this crate needs to have a version, but I guess bumping it doesn't do any harm either.

raphlinus: Yeah, I saw the version, figured it should match, but it doesn't seem important.

nicoburns: I recommend setting all `publish = false` crates to version `0.0.0`. That makes it obvious that it isn't currently being versioned.

raphlinus: So I dug into this a bit. The `version` field is currently *missing* in fearless_simd_tests, but present in the other two unpublished crates (macros and gen). It used to be required, but is [optional](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field) as of Rust 1.75. I think we should be consistent. I'm wondering whether it's better to elide the version field or set them explicitly to 0.0.0 in all three crates.

PR #108, opened 2025-10-14
A stable, safe implementation of "Struct Target Features"

DJMcNab:
The core contributions of this PR are:

- A trait which a (should be zero-sized) struct can implement, which indicates that it is a type-level proof that a set of target features are enabled.
- The `trampoline` macro, which validates a `#[target_feature(enable = "xxx")]` string against values of one or more of these, ensuring at compile time that a call to a `#[target_feature]` function will be safe; and then calling it.
- A corresponding struct for each target feature on x86[-64], which are code generated.

The state of this feature is:

- It is *not* used for implementing the Fearless SIMD crate.
- ~~The x86-64-v{1,2,3,4} level implementations do not exist/are extremely incomplete.~~
- ~~Some docs are missing (these are however not the most critical docs, it's only docs on the groupings of x86 features).~~
- It does not have support for aarch64 in the architecture levels. This is not hard, it's just data wrangling.

There is also an open licensing question, around the docs taken from the Rust reference. My preference would be to copy https://github.com/rust-lang/reference/blob/1d930e1d5a27e114b4d22a50b0b6cd3771b92e31/LICENSE-MIT#L1 into our LICENSE-MIT, which avoids having to make a decision about copyright-ability here.

My proposed next steps are:

- Discuss this at Renderer Office Hours tomorrow: Done
- If we decide this is a direction we want to follow, clean up and land this PR.
   - Bump our MSRV to 1.89 to take https://github.com/rust-lang/rust/issues/111137 into account: Done
   - Add final docs: Done
   - Validate that things are up-to-date in CI: Done
   - True x86-64-v{1,2,3,4} level support: Done
- Follow-up with:
   - aarch64 support
   - Automatic selection/an enum of x86-64 levels
   - Using it in the implementation of Fearless SIMD itself

For review:

- You can mostly ignore the contents of `fearless_simd_core/x86/xxx/xxx.rs`, as these are entirely automatically generated. The exception is `fearless_simd_core/x86/xxx/mod.rs`, which are hand-written, but don't have any logic.

Discussed on Zulip: [#simd > Removing &#96;safe-wrappers&#96;](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Removing.20.60safe-wrappers.60/with/545377857)


fearless_simd_core/src/x86/adx/adx.rs
```diff
@@ -0,0 +1,92 @@
...
+        write!(f, r#""adx" enabled."#)
+    }
+}
+
+unsafe impl TargetFeatureToken for Adx {
+    const FEATURES: &[&str] = &["adx"];
+
+    #[inline(always)]
+    fn vectorize<R>(self, f: impl FnOnce() -> R) -> R {
+        // Because we want this constant to be eagerly evaluated.
```
DJMcNab: To explain this, this comment should be explaining why we don't use `[Self = self]` here (which is that the macro generates `const _: () = {`, so that the validity gets checked by the compiler eagerly (I'll fix the comment in a follow-up commit)

fearless_simd_core/src/lib.rs
DJMcNab, line 1: This file (and `trampoline.rs`) contains the main code needed to understand this PR.

fearless_simd_core/src/support.rs
```diff
@@ -0,0 +1,231 @@
...
+            Self::No { .. } => panic!("Tokens provided are missing a necessary target feature."),
+        }
+    }
+}
+
+/// Determine whether the features in the target feature string `required` are a subset of the features in `permitted`.
+/// See the module level docs [self].
+///
+/// We require static lifetimes as this is primarily internal to the macro.
+pub const fn is_feature_subset<const N: usize>(
```
DJMcNab, line 51: This function needs the most careful review, because its correctness is being relied upon for safety.

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+//!
+//! This crate introduces the [`trampoline!`] macro, which allows running code in a
+//! statically validated `#[target_feature(enable="some_features")]` environment, based on
+//! externally provided tokens.
+//! This abstraction is designed to be combined with target features 1.1, the recent update
+//! in the Rust compiler to allow calling `#[target_feature]` functions safely from within
+//! other `#[target_feature]` functions.
+//! As such, once you have used the [`trampoline!`] macro, you can call any intrinsic in [`core::arch`].
+//!
+//! This crate also has modules which contain tokens for each Rust target features.
```
taj-p:
```suggestion
//! This crate also has modules which contain tokens for each Rust target feature.
```

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+//! These allow safely validating that a target feature is available, and obtaining a token.
+//! These are grouped by architecture:
+//!
+//! - [`x86`] contains the tokens for both the x86 and x86-64 targets.
+//!   It also contains tokens for each x86-64 microarchitecture level, see [`x86::V1`] for details.
+//! <!-- TODO: Other architectures -->
+//!
+//! # Examples
+//!
+//! At the time of writing, it is not possible to turn scalar values into <abbr title="Single InstructionMultiple Data">SIMD</abbr>
```
taj-p:
```suggestion
//! At the time of writing, it is not possible to turn scalar values into <abbr title="Single Instruction Multiple Data">SIMD</abbr>
```

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+//! vector types safely using only the standard library.
+//! These examples use [bytemuck](https://crates.io/crates/bytemuck) for this.
+//!
+//! <!-- TODO -->
```
taj-p: Is this TODO deliberately left for completing in a later PR?

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+//!
+//! Note that for `aarch64`'s neon, you will want to enable bytemuck's `aarch64_simd` feature.
+//! This is also the case for WASM with `wasm_simd`, but note that this crate
+//! [isn't needed on WASM][attributes.codegen.target_feature.wasm], as it is safe to
+//! call `#[target_features]` on that platform.
+//!
+//! # Crate Feature Flags
+//!
+//! <!-- TODO -->
```
taj-p: Is this a deliberately left TODO?

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+// These lints shouldn't apply to examples.
+#![warn(clippy::print_stdout, clippy::print_stderr)]
+// Targeting e.g. 32-bit means structs containing usize can give false positives for 64-bit.
+#![cfg_attr(target_pointer_width = "64", warn(clippy::trivially_copy_pass_by_ref))]
+// END LINEBENDER LINT SET
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![no_std]
+
+// TODO: Do we want both an `x86` and `x86_64` module?
```
taj-p: This question is outside my expertise sorry :(

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+#[cfg(any(target_arch = "x86", target_arch = "x86_64", doc))]
+pub mod x86;
+
+pub mod trampoline;
+
+#[cfg(feature = "std")]
+extern crate std;
+
+/// Token that a set of target feature is available.
```
taj-p:
```suggestion
/// Token that indicates a set of target features is available.
```

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+///
+/// Note that this trait is only meaningful when there are values of this type.
+/// That is, to enable the target features in `FEATURES`, you *must* have a value
+/// of this type.
+///
+/// Values which implement this trait are used in the second argument to [`trampoline!`],
```
taj-p:
```suggestion
/// Values which implement this trait are used in the first argument to [`trampoline!`],
```

I thought it was passed to the first argument of `trampoline!`

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// To do so, you should instead use [`trampoline!`] directly - this is a convenience wrapper around `trampoline`
+    /// for cases where either autovectorisation is sufficient, or dispatch to simd intrinsics is handled elsewhere.
+    fn vectorize<R>(self, f: impl FnOnce() -> R) -> R;
+}
+
+/// Run an operation in a context with specific target features enabled, validated with [`TargetFeatureToken`] values.
+///
+/// This is effectively a stable implementation of the "Struct Target Features" Rust feature,
+/// which at the time of writing is neither in stable or nightly Rust.
+/// This macro can be used to make SIMD dispatch safe in addition to make explicit SIMD, both safely.
```
taj-p:
```suggestion
/// This macro can be used to make SIMD dispatch and explicit SIMD safe.
```

Optional improvement on this sentence

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+///
+/// # Reference
+///
+/// These reference examples presume that you have (values in brackets are the "variables"):
```
taj-p: "These reference examples presume that you have <what>" - is sentence fragment missing its completion?

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+///
+/// - An expression (`token`) of a type (`Token`) which is `TargetFeatureToken` for some target features (`"f1,f2,f3"`);
```
taj-p:
```suggestion
/// - An expression (`token`) of a type (`Token`) which is a `TargetFeatureToken` for some target features (`"f1,f2,f3"`);
```

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+/// - A function (signature `fn uses_simd(val: [f32; 4]) -> [f32; 4]`) which is safe but enables a subset of those target features (`"f1,f2"`);
```
taj-p:
```suggestion
/// - A function (signature `fn uses_simd(val: [f32; 4]) -> [f32; 4]`) which is safe but requires a subset of those target features (`"f1,f2"`);
```

Is requires the right word here?

fearless_simd_core/src/lib.rs
ajakubowicz-canva:
```suggestion
//! At the time of writing, it is not possible to turn scalar values into <abbr title="Single Instruction Multiple Data">SIMD</abbr>
```

fearless_simd_core/src/lib.rs
ajakubowicz-canva: Just to clarify is this awaiting a code example? May be helpful to document what the TODO is awaiting.

fearless_simd_core/src/support.rs
```diff
@@ -0,0 +1,234 @@
...
+    let mut finished = false;
+    'input_feature: while !finished {
+        let mut comma_idx = 0;
+        // Find the first comma in required_bytes, or the end of the string.
+        while comma_idx < required_bytes.len() && required_bytes[comma_idx] != b',' {
+            comma_idx += 1;
+        }
+        // `comma_idx` is now the index of the comma, e.g. if the string was "sse,", idx would be 3
+        // This is the feature we need to validate exists in permitted.
+        let (to_find, remaining_required) = &required_bytes.split_at(comma_idx);
```
taj-p:
```suggestion
        let (to_find, remaining_required) = required_bytes.split_at(comma_idx);
```

fearless_simd_core/src/support.rs
```diff
@@ -0,0 +1,234 @@
...
+        if let [comma, rest @ ..] = remaining_required {
+            if *comma != b',' {
+                panic!("Internal failure of expected behaviour.");
+            } else {
+                required_bytes = rest;
+            }
+        } else {
+            // Exit out of the loop after this iteration.
+            // Note that for input of `""`` and "sse,", we still need to search
+            // for the input target feature `` (i.e. the empty string), to match Rust's behaviour here.
```
taj-p, line 75: 👍

taj-p (commented):
Soundness makes sense to me! Let me know if these TODOs are deliberate, once you've solved the question of whether to split out `x86` into `x86` and `x86_64`, and I'm happy to approve

fearless_simd_core/src/lib.rs
ajakubowicz-canva: Nit: This line is a bit awkward to read.

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+///
+/// let a = [1., 2., 3., 4.];
+/// let Some(token) = token else { return scalar_fallback(a) };
+///
+/// trampoline!(Token = token => "f1,f2", uses_simd(a: [f32; 4]) -> [f32; 4])
+/// ```
+///
+/// Note that a function only operating on 128 bytes is probably too small for checking
+/// whether a token exists just for it is worthwhile.
+/// However, if you have amorphised the cost of that check between many function calls,
```
ajakubowicz-canva: Nit: `amorphised` -> `amortised`?

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+/// }
+///
+/// let a = [1., 2., 3., 4.];
+/// let Some(token) = token else { return scalar_fallback(a) };
+///
+/// trampoline!(Token = token => "f1,f2", uses_simd(a: [f32; 4]) -> [f32; 4])
+/// ```
+///
+/// Note that a function only operating on 128 bytes is probably too small for checking
+/// whether a token exists just for it is worthwhile.
```
ajakubowicz-canva: Nit: grammar

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+/// /// Perform some computation using SIMD.
+/// #[target_feature(enable = "f1,f2")]
+/// fn uses_simd(val: [f32; 4]) -> [f32; 4] {
+///     // ...
+/// }
+///
+/// let a = [1., 2., 3., 4.];
+/// let Some(token) = token else { return scalar_fallback(a) };
+///
+/// trampoline!(Token = token => "f1,f2", uses_simd(a: [f32; 4]) -> [f32; 4])
```
ajakubowicz-canva, line 192: I have a minor misunderstanding regarding why the token feature set needs to be declared `"f1,f2"` here, whilst the feature set is also declared on the `uses_simd` function?

When would the feature string passed to trampoline and the function target feature strings diverge? On this line what happens if in the trampoline call `"f1,f2"` is accidentally written as `"f1"`?

Similarly, I need clarification of the the utility of multiple tokens. E.g. `[Token = token, Sse = my_sse] => "f1,f2,sse"`?

ajakubowicz-canva: > Similarly, I need clarification of the the utility of multiple tokens. E.g. [Token = token, Sse = my_sse] => "f1,f2,sse"?

Attempt at answering my own question.
The list of tokens provided to trampoline provides an explicit list of permitted/witnessed features. The function passed to trampoline  (`uses_simd` in this example), has _required features_ declared in the `target_feature` attribute. Thus, trampoline enforces that it is only safe to call this function if the provided tokens contain the subset of required features?

Very very cool.

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+        // This ensures that the user can't add `#![target_feature(enable = "xxx")]` to their block.
+        // Soundness: Either of generic_args and `$where` could be used to exit the function item early, so aren't
+        // inside an unsafe block.
+        fn trampoline_impl$(<$($generic_args)*>)?($($arg_name: $arg_type),*) $(-> $ret)? $(where $($where)*)? { $op }
+
+        $(
+            // We validate that we actually have a token of each claimed type.
+            let _: $token_type = $token;
+        )+
+        // We use a const item rather than a const block to ensure that.
```
ajakubowicz-canva: Nit: Sentence fragment.

fearless_simd_core/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+
+//! An abstraction to allow safely running custom `#[target_feature]` functions on stable Rust.
+//!
+//! This crate introduces the [`trampoline!`] macro, which allows running code in a
+//! statically validated `#[target_feature(enable="some_features")]` environment, based on
+//! externally provided tokens.
+//! This abstraction is designed to be combined with target features 1.1, the recent update
+//! in the Rust compiler to allow calling `#[target_feature]` functions safely from within
+//! other `#[target_feature]` functions.
+//! As such, once you have used the [`trampoline!`] macro, you can call any intrinsic in [`core::arch`].
```
ajakubowicz-canva, line 12: I think I am starting to understand the power of this abstraction. Per [Stabilize target_feature_11 #134090](https://github.com/rust-lang/rust/pull/134090), it is unsafe to call a function with `target_feature` declared unless the caller is in a context with those features. Thus the initial call into the target_feature context is unsafe. This `trampoline!` provides a safe alternative.

fearless_simd_core/src/support.rs
```diff
@@ -0,0 +1,234 @@
...
+        expect_any_failure("a3,a1,a2", [&["a"]]);
+        expect_success("a1,a2", [&["a1", "a2", "a3"]]);
+        expect_failure("a1,a2", [&["a1", "a3"]], "a2");
+
+        expect_failure("a1b,a2b", [&["a1b", "a3b"]], "a2b");
+        expect_failure("a1b,a2b", [&["a1b", "a3b"]], "a2b");
+        expect_failure("a1b,a2b", [&["a1b", "a3b"]], "a2b");
+        expect_failure("a1b,a2b", [&["a1b", "a3b"]], "a2b");
+    }
+
```
ajakubowicz-canva, line 224: It may be worth adding a test to cover duplicate required features. E.g. this is valid:

```rs
#[target_feature(enable = "sse,sse")]
    fn sse_mul_f32s(a: [f32; 4], b: [f32; 4]) -> [f32; 4] {
        let a: __m128 = bytemuck::must_cast(a);
        let b: __m128 = bytemuck::must_cast(b);
        bytemuck::must_cast(_mm_mul_ps(a, b))
    }
```

Maybe a test:

```rs
expect_success("sse,sse", [&["sse"]]);
```

Similarly it may be worth adding tests for these other cases:
 - `expect_failure("a1,a2", [&["a1,a2"]], "a1");`
 - `expect_success("a,b,c", [&["c"], &["b"], &["a"]]);`

fearless_simd_core/src/support.rs
```diff
@@ -0,0 +1,234 @@
...
+    #[test]
+    fn empty_feature() {
```
ajakubowicz-canva, line 233: `expect_success(",,,,,,", [&[""]]);` – list of empty target features.

fearless_simd_core/src/lib.rs
ajakubowicz-canva: I think the "glamour shot" described in comment https://github.com/linebender/fearless_simd/pull/108#issuecomment-3407267104 makes perfect sense to me now! I really like this PR. Apologies that it's taking me a while to get through. On Monday I plan to go through the code that's outside fearless_simd_core.

fearless_simd_core/src/lib.rs
DJMcNab: Yeah, that was my approximate intention. I'll plan to follow up with that very soon after we land this, but I don't want to potentially block this entire PR on getting those reviewed now.

fearless_simd_core/src/lib.rs
DJMcNab: Yeah, that's correct. This is something we'd update closer to release time - in particular, the `std` feature currently does nothing other than existing for forward-compatability, so there aren't actually any meaningful feature flags.

fearless_simd_core/src/lib.rs
DJMcNab: I think I'll remove this as a TODO, and talk about this explicitly in the docs.

fearless_simd_core/src/lib.rs
DJMcNab: So, this is the intro to the below list. But you're right that it reads strangely, I'll reword!

fearless_simd_core/src/lib.rs
DJMcNab: This is trying to say that it is attributed with `#[target_features(enable = "f1,f2")]`. Again, I'll reword

fearless_simd_core/src/lib.rs
DJMcNab: Yeah, exactly. This is due to our dependency of the target features 1.1 Rust feature, which moves that safety check into the Rust compiler.

And yes, you can use multiple tokens exactly as you describe. I'll see about adding some docs for that.

fearless_simd_core/src/lib.rs
DJMcNab: See https://github.com/linebender/fearless_simd/pull/108#discussion_r2458118935

fearless_simd_core/src/lib.rs
DJMcNab: To put this another way, declaring the target feature string in the macro body allows us to move the target features from just being an attribute, to being both an attribute but also a const value, which we can then perform validation on.

fearless_simd_core/src/lib.rs
DJMcNab: I just removed this block, as it's not really all that helpful.

Shnatsel:
Hey! I'm a new maintainer brought on board to help with dwindling maintenance resources for fearless_simd. I've been looking through the PR backlog to help get things moving.

I think this design is excellent, and it's a shame that it's been languishing in a PR for so long. However, I would struggle to maintain it once merged, since it adds a lot of non-trivial code.

There's also not a lot of synergy with the rest of fearless_simd. This PR is a better version of [archmage](https://crates.io/crates/archmage), letting users combine it with target feature 1.1 and [safe_unaligned_simd](https://crates.io/crates/safe_unaligned_simd) to write fully safe intrinsic code for arbitrary CPU extensions. Right now fearless_simd is mostly safe wrappers around portable operations for fixed CPU feature sets, with optional drop-down into intrinsics for the rare cases when you need that. Fine-grained type-level tracking of CPU features is not really beneficial, and forces trade-offs in the `Simd` types.

I think the best way forward is to publish this work as a separate crate. It would be a clear improvement over `archmage` and I'd be happy to see this made available to users instead of languishing in a stalled PR. I'm happy to help with announcing and promoting the separate crate.

DJMcNab:
I'm glad that you agree that this solves the issues! I do have a few follow-up thoughts from your comments.

> However, I would struggle to maintain it once merged, since it adds a lot of non-trivial code.

As a note on this, obviously the only non-trivial code in this PR is:
1) Fearless_simd_gen, which is at least just string manipulation, but is a little messy.
2) the `is_feature_subset` macro, which is definitely frustrating to review and maintain
3) the trampoline macro. Is the difficulty you forsee there just due to the variety of ways to call it? I could definitely see an argument for slimming that down (although I haven't paged this back in; it's possible that I've already done what I'm thinking of), and would possibly be in favour of it.

> I think the best way forward is to publish this work as a separate crate.

I struggle slightly to see how this would be any different to the current state. To say something obvious; this PR already suggests making a separate crate...

Shnatsel:
> I struggle slightly to see how this would be any different to the current state. To say something obvious; this PR already suggests making a separate crate...

Yeah, I wasn't clear about this, sorry. I think this crate deserves its own repository and its own set of maintainers, independent from the fearless_simd repository. Between fearless_simd having somewhat different priorities and all the maintainers being also busy with other things, I just don't think there is much we can contribute. This PR being inactive for the past 6 months is ample proof of that.

I'm happy to have some sort of interop between the two, e.g. converting the SIMD tokens between the representations, but we'd have to put that into a separate crate to avoid coupling the semver versions of both projects and having to bump both in lockstep.

PR #111, opened 2025-10-25
Introduce Scalar Simd impl

Ralith:
This `Simd` implementation has a native lane width of 1, which is convenient for reusing SIMD-capable logic to compute a single value, either because that's all you need, or to help handle an unaligned total number of inputs.

Blockers:
- Current draft doesn't compile because using `u8` as both an element and a mask type results in some type collisions. Probably easy to fix by adding a newtype. I previously tried `bool` as a mask type but that didn't quite work out.
- Every `Simd` method is copied from `Fallback`. That's a lot of duplication. Should this replace `Fallback`? That might result in worse code for applications which actually do want data-parallelism but aren't targeting an environment otherwise supported by this crate. Is that a realistic concern?
- Does this even solve a real problem? You can always pad unwanted lanes. That might be slower, though -- it would certainly impair any autovec.


Shnatsel:
This would conflict with a number of methods that aren't lane-wise, such as `swizzle_dyn`, and so wouldn't slot into the trait hierarchy, which would rather defeat the point of trying to unify vectors and scalars.

There hasn't been much interest in this, so I'm going to go ahead and close this, but we can revisit this if compelling use cases arise.

Ralith:
FWIW I *very* commonly want this for cases where I want only a single datapoint from a SIMD-capable algorithm. I'm getting by for now just computing 4/8 values and throwing most of them away, but that kind of sucks.

In abstract, it seems like code that abstracts over 4-vectors and 8-vectors should also be able to handle 1-vectors.

PR #115, opened 2025-11-12
Implement "real" AVX2 intrinsics and clean up x86 codegen

### LINKED ISSUE #114: Actually make use of AVX2's increased lane width
While SSE operates on 128-bit vectors, as NEON and WebAssembly do, AVX2 uses 256-bit vectors.

Currently, it seems all of fearless_simd is hardcoded to use 128-bit vectors. The `Simd` trait has an associated `Block` type, and implementing that trait requires `Block` to specifically be 128-bit. I assume there's some design rationale behind this that I'm missing.

LLVM might be able to optimize the split/combine stuff back into native 256-bit AVX2 operations, but I wouldn't count on it.

valadaptive:
Resolves #114.

This may be best reviewed one commit at a time; one of them moves a lot of stuff around.

This PR updates the x86 codegen to use actual AVX2 intrinsics (the ones starting with `_mm256`). This is mostly straightforward, but there are a few operations that require special attention. I've included some other x86 codegen fixes and improvements that are somewhat interwoven:

- I've added tests for several operations that were previously untested. Mainly these are 256-bit zip/unzip, widen/narrow, split/combine, and integer equality comparisons. Note that these test cases were generated by Claude.

- The x86 codegen now actually generates the correct code for integer equality comparisons. Previously, it incorrectly generated "greater than" comparisons instead.

- It also now uses the `blendv` family for "select" operations. Intel's manual says these are available starting in SSE4.1. Not sure if there's a reason this wasn't done before.

- For SSE4.2-level unzip operations, I've changed the codegen.

    Previously, for `unzip_low`, it would shuffle the inputs to put the even-indexed elements in both the lower and upper halves of the values, then use `unpacklo` to select just the lower halves. Likewise, for `unzip_high`, it would shuffle the inputs to put the *odd*-indexed elements in both halves, and use `unpacklo` once more.

    I've changed this so that `unzip_low` and `unzip_high` both use a shuffle operation that moves the even-indexed elements into the lower halves and the odd-indexed elements into the upper halves. `unzip_low` uses `unpacklo` to select the lower halves, and `unzip_high` uses `unpackhi` to select the upper halves. This means that if the user calls both `unzip_low` and `unzip_high`, the shuffle operation's result can be shared.

- I've implemented 8-bit multiplication based on [this StackOverflow answer](https://stackoverflow.com/questions/8193601/sse-multiplication-16-x-uint8-t).

On the AVX2 side, most existing 128-bit operations have a straightforward 256-bit counterpart, but some are more involved:

- The zip/unzip operations are a bit more complicated, since most AVX2 swizzle operations operate *within* each 128-bit lane. For 32-bit and larger operations, there are special "lane-crossing" shuffles we can use instead. Operations on smaller scalars require a combination of intra-lane and "lane-crossing" shuffles.

- Splitting a 256-bit vector to a 128-bit one, or combining two 128-bit vectors into a 256-bit one, can be done directly with AVX2 intrinsics.

- Widen/narrow operations can be done a bit more efficiently in AVX2. Widening a u8x16 to a 16x16 can be done with a single `_mm256_cvtepu8_epi16`. Narrowing a u16x16 to a u8x16 can done with two shuffles: one to extract the lower bits of each 16-bit value within each 128-bit lane, and one to combine the two lanes.

I've consolidated much of the x86 codegen from `x86_common.rs`, `arch/avx2.rs`, `arch/sse4_2.rs`, and `arch/x86_common.rs` into a single `arch/x86.rs` file. I did this in the middle of some other commits; sorry! The main AVX2 codegen was implemented before the reorganization, but the split/combine and widen/narrow ops were implemented afterwards.

In the future, I'd like to rework and tidy up the codegen a bit more. For instance, we're passing in things like vector types' widths alongside those very same vector types, which is redundant. The `Arch` trait is also very much not pulling its weight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ This release has an [MSRV][] of 1.88.
 ### Added
 
 - All vector types now implement `Index` and `IndexMut`. ([#112][] by [@Ralith][])
+- 256-bit vector types now use native AVX2 intrinsics on supported platforms. ([#115][] by [@valadaptive][])
+- 8-bit integer multiplication is now implemented on x86. ([#115][] by [@valadaptive][])
+
+### Fixed
+
+- Integer equality comparisons now function properly on x86. Previously, they performed "greater than" comparisons.
+  ([#115][] by [@valadaptive][])
 
 ### Changed
 
@@ -27,6 +34,7 @@ This release has an [MSRV][] of 1.88.
   A consequence of this is that the available variants on `Level` are now dependent on the target features you are compiling with.
   The fallback level can be restored with the `force_support_fallback` cargo feature. We don't expect this to be necessary outside
   of tests.
+- Code generation for `select` and `unzip` operations on x86 has been improved. ([#115][] by [@valadaptive][])
 
 ### Removed
 
@@ -86,6 +94,7 @@ No changelog was kept for this release.
 
 [@Ralith]: https://github.com/Ralith
 [@DJMcNab]: https://github.com/DJMcNab
+[@valadaptive]: https://github.com/valadaptive
 
 [#75]: https://github.com/linebender/fearless_simd/pull/75
 [#76]: https://github.com/linebender/fearless_simd/pull/76
@@ -103,6 +112,7 @@ No changelog was kept for this release.
 [#96]: https://github.com/linebender/fearless_simd/pull/96
 [#99]: https://github.com/linebender/fearless_simd/pull/99
 [#105]: https://github.com/linebender/fearless_simd/pull/105
+[#115]: https://github.com/linebender/fearless_simd/pull/115
 
 [Unreleased]: https://github.com/linebender/fearless_simd/compare/v0.3.0...HEAD
 [0.3.0]: https://github.com/linebender/fearless_simd/compare/v0.3.0...v0.2.0


fearless_simd_gen/src/mk_avx2.rs
```diff
@@ -82,20 +76,18 @@ fn mk_simd_impl() -> TokenStream {
...
-            // FMA). In the future, we'll obviously want to use AVX2 intrinsics for 256 bit.
-            let b1 = (vec_ty.n_bits() > 128 && !matches!(method, "split" | "narrow"))
-                || vec_ty.n_bits() > 256;
+            let too_wide = (vec_ty.n_bits() > 256 && !matches!(method, "split" | "narrow"))
+                || vec_ty.n_bits() > 512;
 
-            let b2 = !matches!(method, "load_interleaved_128")
-                && !matches!(method, "store_interleaved_128");
+            let acceptable_wide_op = matches!(method, "load_interleaved_128")
+                || matches!(method, "store_interleaved_128");
```
DJMcNab, line 83: Just want to check these these don't need to be `load_interleaved_256`

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -83,18 +76,18 @@ fn mk_simd_impl() -> TokenStream {
...
-            let b1 = (vec_ty.n_bits() > 128 && !matches!(method, "split" | "narrow"))
+            let too_wide = (vec_ty.n_bits() > 128 && !matches!(method, "split" | "narrow"))
                 || vec_ty.n_bits() > 256;
 
-            let b2 = !matches!(method, "load_interleaved_128")
+            let acceptable_wide_op = !matches!(method, "load_interleaved_128")
                 && !matches!(method, "store_interleaved_128");
 
-            if b1 && b2 {
+            if too_wide && acceptable_wide_op {
```
DJMcNab: The variable named `acceptable_wide_op` seems to have opposite meanings in this and the avx2 file. Pointing this out as a point of suspicion, rather than something which I've confirmed is wrong.

DJMcNab (commented):
I don't really have the domain knowledge to validate all of the logic in here, but it's good to see more testing. I've read through the code and pointed at what I can see which seems suspect.
Hopefully we can discuss at office hours, and see if anyone else is interested in reviewing this. But I'd be happy landing this by the end of this week if we don't get other review; it can always be reviewed post-merge.

It might be worth also running Vello's tests with this version (would it make sense to also run the benchmarks?)

fearless_simd_gen/src/mk_avx2.rs
valadaptive: I believe there is no `load_interleaved_256`. The name `load_interleaved_128` is a bit confusing since it's actually performing a 512-bit load (aka 64 bytes); not sure where 128 comes in.

fearless_simd_gen/src/mk_sse4_2.rs
valadaptive: I'd renamed the variables here for consistency with the AVX2 version, but forgot to invert the conditions. Fixed that now.

fearless_simd_gen/src/mk_avx2.rs
DJMcNab: :+1: I thought it was probably right - I was just playing a bit of "spot the difference" with the sse4.2 version

LaurenzV: Yeah perhaps not the best name, the 128 was because it's basically interleaving it in steps of 4.

LaurenzV:
The tests of vello_sparse_tetss should run with AVX2 as well in CI, I think.

valadaptive:
I ran `cargo test --workspace --release` in Vello and everything passes. I was a bit confused because some AVX2 tests *appeared* to be missing, but that's just a consequence of `cargo test`'s output being shuffled around a bit. I've made my tentative Vello updates public at https://github.com/linebender/vello/pull/1288; I'll just wait for the CI now.

valadaptive:
I'll go ahead and merge this since the existing tests, my new tests, and the Vello tests all pass. The current x86 code is a bit dodgy anyway (for example, equality comparisons being broken), and I think this PR is an improvement.

This should unblock a fair amount of stuff.

PR #116, opened 2025-11-12
Clean up the codegen a bit (particularly x86)

valadaptive:
This builds on top of #115. There are no functional changes to the generated code (besides what #115 does), but cleans up the `fearless_simd_gen` code:

- The `Arch` trait has been removed. It operated at the wrong level of abstraction--it makes no sense to call e.g. `mk_avx2::make_method` with any `Arch` implementation other than `X86`.

- Many code generation functions in the AVX2 and SSE4.2 modules used to pass in the vector type along with its scalar and total bit widths. The former provides the latter, so we can stop passing all three in and just pass in the vector type.

### DIFF
--- a/fearless_simd_gen/src/arch/fallback.rs
+++ b/fearless_simd_gen/src/arch/fallback.rs
@@ -3,7 +3,6 @@
 
 #![expect(
     clippy::match_single_binding,
-    unreachable_pub,
     reason = "TODO: https://github.com/linebender/fearless_simd/issues/40"
 )]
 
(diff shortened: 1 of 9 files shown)


valadaptive:
I've rebased this now that #115 has landed. I've made a couple more changes: in addition to removing the `Arch` trait, I've now removed the marker structs that used to implement it. I've also made a few more functions `pub(crate)`, fixing most (but not all) `unreachable_pub` lints.

Issue #117, opened 2025-11-12
Expose `Level::new()` on `no_std`?


valadaptive:
The `Level::try_detect()` documentation says says:

> Should be used in libraries if they wish to handle the case where target features cannot be detected at runtime. Most users should prefer `new`.

This is not very helpful for `no_std` users, since they cannot use `Level::new()`. Instead, the proper pattern for `no_std` users is `Level::try_detect().unwrap_or(Level::baseline())`.

I believe `Level::new()` should behave similarly to `Level::try_detect().unwrap_or(Level::baseline())`, and be available even without the `std` feature enabled. This also opens up the possibility of [`no_std` runtime feature detection on x86 by directly parsing the CPUID](https://crates.io/crates/core_detect) (runtime feature detection on aarch64 isn't a big deal, since Neon is almost always statically enabled).

LaurenzV:
I think that’s what we had initially but we changed it for some reason, maybe Daniel remembers why.

Shnatsel:
I was looking into using https://github.com/gz/rust-cpuid or https://docs.rs/cpufeatures/latest/cpufeatures/ for this as well, but https://github.com/linebender/fearless_simd/issues/157 gave me pause

PR #121, opened 2025-11-12
In `try_detect`, use the baseline iff detection wouldn't do better

DJMcNab:
This is part of the work on #117; from a docs and end-user perspective this is pretty much done, but to truly fix that issue we would also need to add a `core_detect` feature.

As discussed in [#office hours > Renderer 2025-11-12](https://xi.zulipchat.com/#narrow/channel/359642-office-hours/topic/Renderer.202025-11-12/with/555231948)

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -207,28 +207,23 @@ pub enum Level {
 impl Level {
     /// Detect the available features on the current CPU, and returns the best level.
     ///
-    /// If no SIMD instruction set is available, a scalar fallback will be used instead.
+    /// If no SIMD instruction set supported by Fearless SIMD is available,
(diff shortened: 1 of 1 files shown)


valadaptive (commented):
After thinking about this a bit more, I'm worried this could be an even bigger footgun. This PR changes `try_detect` to return the baseline if it's the highest level that *we* support.

This means that if we later add support for new instruction sets, like SVE on AArch64 or AVX-512 on x86, `try_detect` could stop returning a value!

This means that if a library goes with the pattern recommended in this PR:

```rust
let level = Level::try_detect().expect("[LIBRARY_NAME] requires the standard library for this platform, etc etc...");
```

it could stop working and panic at runtime as a result of us adding support for *more* instruction sets. That seems counterintuitive.

PR #122, opened 2025-11-14
Build WebAssembly tests for the `wasm32-wasip1` target and run them using wasmtime

valadaptive:
[WASI](https://wasi.dev/) is now mature enough that we can build binaries that target WebAssembly, and they will "just work" if run in a WebAssembly runtime.

This means we can get rid of the [increasingly-unmaintained](https://blog.rust-lang.org/inside-rust/2025/07/21/sunsetting-the-rustwasm-github-org/) `wasm-pack` and `wasm-bindgen` machinery, and run tests the same way we do for any other target, including doctests.

We *could* add a `.cargo/config.toml` to this repository that specifies the rustflags and runner, but that would affect every package and cargo command. Not sure if that's desirable.

Note that some of the doctests for `wasm32-wasip1` fail to compile, probably due to #105:

<details>

<summary>Error messages</summary>

```
---- fearless_simd/src/macros.rs - macros::dispatch (line 34) stdout ----
error[E0599]: no variant or associated item named `Fallback` found for enum `Level` in the current scope
  --> fearless_simd/src/macros.rs:43:1
   |
12 | dispatch!(level, simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ variant or associated item not found in `Level`
   |
   = note: this error originates in the macro `$crate::dispatch` which comes from the expansion of the macro `dispatch` (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0599`.
Couldn't compile the test.
---- fearless_simd/src/lib.rs - (line 23) stdout ----
error[E0599]: no variant or associated item named `Fallback` found for enum `Level` in the current scope
  --> fearless_simd/src/lib.rs:33:1
   |
13 | dispatch!(level, simd => sigmoid(simd, &[/*...*/], &mut [/*...*/]));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ variant or associated item not found in `Level`
   |
   = note: this error originates in the macro `$crate::dispatch` which comes from the expansion of the macro `dispatch` (in Nightly builds, run with -Z macro-backtrace for more info)
```
</details>

I've `ignore`d those doctests for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,7 +241,7 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_STABLE_VER }}
-          targets: wasm32-wasip1
+          targets: wasm32-unknown-unknown,wasm32-wasip1
 
       - name: restore cache
         uses: Swatinem/rust-cache@v2
@@ -252,6 +252,12 @@ jobs:
         uses: taiki-e/install-action@v2
         with:
           tool: cargo-nextest,wasmtime-cli
+
+      # We test using wasm32-wasip1, but want to make sure that wasm32-unknown-unknown still builds
+      - name: cargo build (wasm32-unknown-unknown)
+        run: |
+          cargo build --target wasm32-unknown-unknown \
+            --config 'target.wasm32-unknown-unknown.rustflags = "-Ctarget-feature=+simd128"'
 
       - name: cargo nextest
         run: |


fearless_simd/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //! To call a function with the best available target features and get the associated `Simd`
 //! implementation, use the [`dispatch!()`] macro:
 //!
-//! ```rust
+//! ```rust,ignore
```
Ralith: Does this put the doctests at risk of bitrotting? Why do they break the build?

valadaptive: Augh, figured out why the doctests didn't pass. `RUSTFLAGS` and `RUSTDOCFLAGS` both need to be set to the exact same values, or weird things happen (https://github.com/rust-lang/rust/issues/134172, see also https://github.com/rust-lang/cargo/issues/12739 and [this IRLO discussion thread](https://internals.rust-lang.org/t/can-we-do-something-about-rustflags-rustdocflags/21513)). After setting `target.wasm32-wasip1.rustdocflags` in the test config, things work.

.github/workflows/ci.yml
```diff
@@ -241,22 +241,31 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_STABLE_VER }}
-          targets: wasm32-unknown-unknown
```
ajakubowicz-canva, line 244: First, thank you for the awesome PR! This comment can be treated as non-blocking – but I am just curious and honestly lacking knowledge of other `wasm32` targets.

We (Canva) currently use the `vello_hybrid` / `vello_cpu` renderers via `wasm32-unknown-unknown` using `wasm-bindgen` directly. From that perspective I'd be concerned that we lose that test coverage.

Is there a solution for migrating from `wasm32-unknown-unknown` to `wasm32-wasip1` for Rust on the browser?

valadaptive: I don't think so at the moment; there's https://github.com/wasm-bindgen/wasm-bindgen/issues/3421 but it's kinda dead (as is wasm-bindgen in general, unfortunately).

The Vello tests will still run on `wasm32-unknown-unknown`, so test coverage should be maintained through those. I'm also not removing `wasm32-unknown-unknown` support from this library; this is just a matter of making the tests easier to run and extend.

It might be worth adding a CI check that `wasm32-unknown-unknown` still *builds*, but I seriously doubt there'll be any functional differences between `wasm32-wasip1` and `wasm32-unknown-unknown`. For example, there are a lot of JavaScript packages on NPM that run all their unit tests on Node, but that doesn't mean those packages don't work in the browser.

ajakubowicz-canva: That sounds good to me! Thank you!

valadaptive: I've added an additional step to the "cargo test (wasm32)" workflow that runs `cargo build --target wasm32-unknown-unknown`. This should ensure that target still builds.

fearless_simd_dev_macros/src/lib.rs
```diff
@@ -41,15 +41,14 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
     #[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
     let include_avx2 = false;
     // Note that we cannot feature-gate this with `target_arch`. If we run
-    // `wasm-pack test --headless --chrome`, then the `target_arch` will still be set to
+    // `cargo test --target wasm32-wasip1`, then the `target_arch` will still be set to
     // the operating system you are running on. Because of this, we instead add the `target_arch`
     // feature gate to the actual test.
     let include_wasm = !exclude_wasm(&input_fn_name.to_string());
```
DJMcNab, line 47: It's worth checking whether we can use `CARGO_CFG_TARGET_ARCH` (ideally for all of these tests, because in theory we could be cross-compiling to run through miri or qemu).

This should be a follow-up/an issue though.

DJMcNab (approved):
I've not followed these steps locally, but this looks much less fragile to me. There is value in this stuff being tested in the browser, to ensure that we're aware (and ideally can report it upstream) if we're exercising code paths which some browser doesn't support or miscompiles. However, a WASI emulation layer would be a much nicer path towards that than requiring a JavaScript layer, and that isn't exactly high-priority.

fearless_simd_dev_macros/src/lib.rs
valadaptive: We can't; it's not exposed at the time proc macros are run.

valadaptive:
I'll go ahead and merge this. If we want to do something as sophisticated as checking for miscompilations in each browser engine, it'd probably be helpful to have complex integration tests instead of just the unit tests we have here. All of that can be deferred to the future.

Issue #124, opened 2025-11-14
Make sure we're not using any unsupported intrinsics in the various x86 levels


valadaptive:
x86's SIMD support is a huge mess, and it's not clearly delineated by vector width. I noticed this when looking at https://github.com/linebender/fearless_simd/pull/97/: `_mm_mullo_epi64` is not an SSE intrinsic nor even an AVX2 intrinsic, but requires AVX512! This is very hard to catch, and I only noticed it because Clippy flagged its usage as an MSRV violation.

We should add some sort of check to ensure we don't accidentally use any unsupported intrinsics on x86. My first thought was the Clippy `disallowed_methods` lint, but Clippy lints can only be configured on a per-workspace basis, and we need something that can vary per-*file*.

My idea is this: using whatever method @raphlinus used to generate the `core_arch` modules (it doesn't seem to be checked into the repository), we can gather a list of intrinsics that are supported at a given feature level. Then, for the generated implementations (currently `sse4_2.rs` and `avx2.rs`), we create an inner module that re-exports only the *supported* intrinsics from `core::arch::x86_64`, and wildcard-import from that inner module. This is done to avoid any "unused import" lint warnings while not requiring us to maintain a full list of every single intrinsic we call into.

Shnatsel:
We have CI checks for erroneously emitted instructions not available on the CPU, added in https://github.com/linebender/fearless_simd/pull/197 and #198

The caveat is the illegal instruction needs to be actually executed, so it relies on test coverage.

PR #123, opened 2025-11-14
Widen `Avx2`'s associated types to 256 bits

valadaptive:
Split off from #115 to make review of that PR easier.

PR #125, opened 2025-11-14
Add `f64s` and `mask64s` associated types

valadaptive:
Stacked on #123 since it touches the same code.

Adding 64-bit *integer* types is tricky because SSE4.2 and AVX2 are missing some important 64-bit operations, and I don't feel comfortable implementing those until #124 is addressed. However, we already have `mask64x[N]` and `f64x[N]` types.

The `f32s` associated type has a `Bytes = <Self::u32s as Bytes>::Bytes` constraint, which misled me into thinking that a float type *needs* an associated int type, but that constraint is only there to assert that the two share the same `Bytes` type.


valadaptive:
This has been rebased and can be approved now.

DJMcNab (approved):

PR #126, opened 2025-11-15
Build all tests for better cross-compilation

valadaptive:
This was discussed a bit in #122. We were previously omitting architecture-specific tests based on the *host* architecture, which is incorrect and prevents cross-compiling tests to run under something like qemu.

We can't access `CARGO_CFG_TARGET_ARCH` at the time that proc macros are evaluated, so we just have to build every architecture-specific test and use `#[cfg]` to disable ones that aren't applicable to the current architecture.

Even more annoyingly, there's no way to conditionally ignore a test at runtime. This means that if the CPU we're running the tests on doesn't support a target feature, we have to either pass or fail that test. I've chosen to fail, to prevent faulty code from appearing to pass tests.

I've also updated the architecture-specific `exclude` functions to add `#[ignore]` attributes rather than omitting the tests entirely. This makes it more clear that the tests exist, but are not being run because they don't yet pass.

### DIFF
--- a/fearless_simd_tests/tests/mod.rs
+++ b/fearless_simd_tests/tests/mod.rs
@@ -43,29 +43,41 @@ fn generic_cast<S: Simd>(x: S::f32s) -> S::u32s {
     unused_variables,
     reason = "The constructed `Level` is only used in some cfgs."
 )]
+#[allow(
+    dead_code,
+    reason = "The `UNSUPPORTED_LEVEL_MESSAGE` is only used in some cfgs."
+)]
 #[test]
 fn supports_highest_level() {
+    const UNSUPPORTED_LEVEL_MESSAGE: &str = "This means that some of the other tests in this run may be false positives, that is, they have been marked as succeeding even though they would actually fail if they could run.\n\
+        When these tests are run on CI, any false positives should be caught.\n\
+        However, please open a thread in the #simd channel on the Linebender Zulip if you see this message.\n\
+        That would allow us to know whether it's worth us setting up the tests to run on an emulated system (such as using QEMU).";
+
+    let level = Level::new();
+
     // When running tests locally, ensure that every SIMD level to be tested is actually supported. The tests themselves
     // will return early and pass if run with an unsupported SIMD level.
     //
     // We skip this on CI because some runners may not support all SIMD levels--in particular, the macOS x86_64 runner
     // doesn't support AVX2.
     if std::env::var_os("CI").is_none() {
-        let level = Level::new();
-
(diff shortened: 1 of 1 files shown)


valadaptive:
Hmm, it seems this has caught something already. The macOS x86 CI machines don't support AVX2. Should I:

- Change the tests to pass if the required target features aren't available?
- Run them in Rosetta 2 on a macOS aarch64 runner?
- Just hold off on this for now?

DJMcNab:
I think the only reasonable options within the bounds that libtest gives us is to pass the test.

I think we should have one test, which is ignored on CI (i.e. with an environment variable), which requires that you have all the supported levels on your machine.
That ensures that contributors with underpowered machines know that some of the test successes might be phantom.

Ideally, we'd point to Qemu to fix this, although I'm not sure how realistic that is.

fearless_simd_tests/tests/mod.rs
```diff
@@ -37,3 +37,35 @@ fn saturate_float_to_int<S: Simd>(simd: S) {
...
+    //
+    // We skip this on CI because some runners may not support all SIMD levels--in particular, the macOS x86_64 runner
+    // doesn't support AVX2.
+    if std::env::var_os("CI").is_none() {
+        let level = Level::new();
+
+        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+        assert!(
+            level.as_avx2().is_some(),
+            "This machine supports AVX2 and below"
```
DJMcNab: I think it would be worth giving a bit more context here (same applies to aarch64, so you might want to extract part of this into a `const`)

```suggestion
            "This machine does not support every `Level` supported by Fearless SIMD (currently AVX2 and below).\n\
             This means that some of the other tests in this run may be false positives, that is, they have been marked as succeeding even though they would actually fail if they could run.\n\
             When these tests are run on CI, any false positives should be caught.\n\
             However, please open a thread in the #simd channel on the Linebender Zulip if you see this message.\n\
             That would allow us to know whether it's worth us setting up the tests to run on an emulated system (such as using QEMU)."
```

DJMcNab (approved):
Ideally, we'd validate automatically that CI *does* run all of these tests, but that's begging for flaky CI.

I wonder if it's worth setting up QEMU emulation for this anyway; but that should be a follow-up.

fearless_simd_tests/tests/mod.rs
valadaptive: I've updated the assertion message for x86 and AArch64. I've also pulled out the WASM SIMD128 assert from the CI check and given it its own message, since it should always be *statically* enabled and not require any runtime feature detection.

PR #130, opened 2025-11-17
Implement `core::ops::Not` for integer types

valadaptive:
I don't know if there's a reason this wasn't implemented previously.


Ralith (approved):
Not omitted intentionally, just nobody'd gotten to it yet. Might be useful to conduct a thorough survey of applicable traits.

PR #134, opened 2025-11-19
Implement all float<->int conversions on x86

valadaptive:
This PR implements fully-compliant conversions for f32->i32, f32->u32, i32->f32, and u32->f32. They are implemented thusly:

- f32 to i32: Check if any operands are not less than 2147483648.0. This catches out-of-range values and NaN, which is not less than any number. If so, clamp the values and also handle NaN. This is about 25% slower than raw `_mm_cvttps_epi32` in my microbenchmark.

- f32 to u32: Clamp any negative or NaN values to 0 using `_mm_max_ps`. We then check again if any operands are above 2147483648.0. If so, subtract that value from the input to give us the "residual" value, convert it to i32, and add it to the final result. There is another check to ensure we don't exceed the *unsigned* range. This is about as fast as f32->i32.

- i32 to f32: This is easy, since it's the one intrinsic that Intel actually gives us.

- u32 to f32: I will admit I don't fully understand how this one works. It's what LLVM emits, and it's faster than [the other conversion I found](https://stackoverflow.com/questions/34066228/how-to-perform-uint32-float-conversion-with-sse). It requires no actual conversion instructions.

I've tested these conversions exhaustively on every 32-bit bit pattern, and they all return the correct results.

### DIFF
--- a/fearless_simd_tests/tests/mod.rs
+++ b/fearless_simd_tests/tests/mod.rs
@@ -7,6 +7,7 @@
 )]
 
 use fearless_simd::*;
+use fearless_simd_dev_macros::simd_test;
 
 mod harness;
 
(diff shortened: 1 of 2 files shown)


fearless_simd_tests/tests/mod.rs
```diff
@@ -3,35 +3,13 @@
...
-        [0; 4]
-    );
-    assert_eq!(
-        <[i32; 4]>::from(simd.cvt_i32_f32x4(simd.splat_f32x4(f32::INFINITY))),
-        [i32::MAX; 4]
-    );
-    assert_eq!(
-        <[i32; 4]>::from(simd.cvt_i32_f32x4(simd.splat_f32x4(-f32::INFINITY))),
-        [i32::MIN; 4]
-    );
```
LaurenzV, line 32: Why are we removing these?

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -768,30 +768,151 @@ pub(crate) fn handle_cvt(
...
+                        unsafe {
+                            let a = a.into();
+
+                            let mut converted = #convert(a);
+
+                            // In the common case where everything is in range, we don't need to do anything else.
+                            let in_range = #cmplt(a, #set1_float(2147483648.0));
+                            let all_in_range = #movemask(in_range) == #all_ones;
+
+                            if !all_in_range {
```
LaurenzV, line 816: Have you benchmarked whether having this branch is faster then just doing the calculation and then doing the select? Just curious.

fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -768,30 +768,151 @@ pub(crate) fn handle_cvt(
...
+                                // If we are above i32::MAX (2147483647), clamp to it.
+                                converted = #blend(#set1_int(i32::MAX), converted, #cast_to_int(in_range));
+                                // Set NaN to 0. Using `and` seems slightly faster than `blend`.
+                                let is_not_nan = #cast_to_int(#cmpord(a, a));
+                                converted = #and(converted, is_not_nan);
+                                // We don't need to handle negative overflow because Intel's "invalid result" sentinel
```
LaurenzV, line 822: Just to be clear, Intel here means all SSE4.2 CPUs, right?

fearless_simd_tests/tests/mod.rs
LaurenzV, line 1: Can you maybe add the exhaustive test you mentioned for all 32-bit patterns and mark it with `ignore`? So that we can rerun it in the future if we want to.

fearless_simd_gen/src/mk_sse4_2.rs
valadaptive: I have. Doing it without this branch takes about 1.2 seconds on every bit pattern, whereas doing it with the branch takes about 1.07 seconds. Given that most floats are probably not NaN or huge, I'd expect the branch to be highly predictable.

fearless_simd_gen/src/mk_sse4_2.rs
valadaptive: Everything [SSE2 and up](https://www.felixcloutier.com/x86/cvtps2dq). There are several forms of the same instruction (depending on vector width) and they all return -2147483648. The description I linked to seems incorrect--it states that the indefinite integer value is "(2^w)-1, where w represents the number of bits in the destination format", but it's actually 2^(w-1).

fearless_simd_tests/tests/mod.rs
valadaptive: Hm. I added some saturating tests in `harness/mod.rs`, but I apparently forgot to add any with positive or negative infinity. I'll add some.

fearless_simd_tests/tests/mod.rs
valadaptive: I've added the exhaustive tests for all four conversions (u32>f32, i32->f32, f32->u32, f32->i32). I rebased this PR on the latest main so I could use the new `from_fn` method to construct the test vectors.

Issue #40, opened 2025-07-23
Tracking burning down clippy issues


ajakubowicz-canva:
As part of ensuring `fearless_simd` adheres to Linebender standards ([#simd > Linebender Standards @ 💬](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Linebender.20Standards/near/527927335)), fearless_simd needs to adhere to the standard Linebender CI: https://github.com/linebender/fearless_simd/pull/39

In fixing up the CI, this raised many clippy errors that have been placed in `expects`. This task tracks triaging those expect attributes.

They can be found in the code by searching for this issue number.

PR #135, opened 2025-11-21
Fix a bunch of Clippy lints

valadaptive:
Mostly in `fearless_simd_gen`, but some in `fearless_simd` itself. Making progress on #40. Fingers crossed this passes CI; there's probably at least one configuration where I missed a spot.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,10 @@ clippy.redundant_feature_names = "warn"
 clippy.wildcard_dependencies = "warn"
 # END LINEBENDER LINT SET
 
+# This was upgraded from a "pedantic" to a "style" lint in Rust 1.88, and downgraded again in 1.89 due to
+# https://github.com/rust-lang/rust-clippy/issues/15151.
+clippy.uninlined_format_args = "allow"
+
 [workspace.dependencies]
 fearless_simd = { path = "fearless_simd" }
 fearless_simd_dev_macros = { path = "fearless_simd_dev_macros" }


fearless_simd/src/core_arch/mod.rs
```diff
@@ -5,6 +5,7 @@
 
 #![expect(
     missing_docs,
+    clippy::new_without_default,
```
LaurenzV, line 8: Why is there a new lint that now doesn't work?

fearless_simd/src/generated.rs
```diff
@@ -5,6 +5,7 @@
     missing_docs,
     clippy::cast_possible_truncation,
     clippy::unseparated_literal_suffix,
+    clippy::use_self,
```
LaurenzV, line 8: Same as above

fearless_simd/src/core_arch/mod.rs
valadaptive: This was moved in from the parent module. The architecture tokens have a `new` method but no `default` method. Previously, this also applied to `Level`, so the lint was allowed for all of `fearless_simd`.

fearless_simd/src/generated.rs
valadaptive: Same as above; this was moved in from the root of `fearless_simd`. I fixed some `use_self` lints in the root, but not in the generated code.

fearless_simd/src/core_arch/mod.rs
LaurenzV: I see!

PR #136, opened 2025-11-21
Implement new min/max/min_precise/max_precise semantics

### LINKED ISSUE #133: Nail down the semantics for `min`/`max` and `min_precise`/ `max_precise`
Different architectures have different semantics for floating-point minimum and maximum, particularly around signed zero and NaN. We currently provide the `min`/`max` operations (which are assumed to map to the "fastest" operations on the target platform), and `precise_min`/`precise_max` (which our tests currently say should behave like ARM's [FPMinNum](https://developer.arm.com/documentation/ddi0596/2021-03/Shared-Pseudocode/Shared-Functions?lang=en#shared.functions.float.fpminnum.FPMinNum)/FPMaxNum).

There are a variety of minimum/maximum semantics to choose from, not all of which are available on all platforms:

## IEEE 754-2019

The IEEE 754-2019 standard defines two sets of min/max operations: `minimum`/`maximum` and `minimumNumber`/`maximumNumber`.

The first set, `minimum`/`maximum`, returns NaN if either operand is NaN. The latter, `minimumNumber`/`maximumNumber`, returns NaN only if both operands are NaN. If only one operand is NaN, it returns the other operand.

Both operations treat negative zero as less than zero.

`minimumNumber`/`maximumNumber` seem like the most "correct" operations, but they have the downside of not being implemented on all platforms.

## x86

x86 provides the `minps`/`maxps`/`minpd`/`maxpd` instructions. As far as I can tell, they map to the following pseudocode:

```rust
fn minps(a: f32, b: f32) -> f32 {
  if a < b { a } else { b }
}

fn maxps(a: f32, b: f32) -> f32 {
  if a > b { a } else { b }
}
```

This implies the following semantics:
- If either operand is NaN, it returns the second operand *b* (since the comparison will always be false).
- If both operands are zero, it returns the second operand *b* regardless of either operand's sign (since negative and positive zero compare as equal).
- Otherwise, it returns the smallest operand as expected.

## WebAssembly

WebAssembly provides the `fmin`/`fmax` and `fpmin`/`fpmax` abstract operations.

The first set,  `fmin`/`fmax`, implement IEEE 754-2019's `minimum` and `maximum` operations exactly. That is, they return NaN if either operand is NaN, and treat negative zero as less than zero.

The second set, `fpmin`/`fpmax`, are similar to Intel's instructions except with the operands flipped:
```rust
fn fpmin(a: f32, b: f32) -> f32 {
  if b < a { b } else { a }
}

fn fpmax(a: f32, b: f32) -> f32 {
  if b > a { b } else { a }
}
```

That is, if either operand is NaN or both operands are zero, they return the *first* operand. https://github.com/linebender/fearless_simd/pull/23 updated our WebAssembly `min_precise`/`max_precise` implementations to flip the operand order and match Intel's behavior.

## AArch64 w/ NEON

NEON actually implements both sets of IEEE 754-2019 operations. In [the list of abstract operations](https://developer.arm.com/documentation/ddi0596/2021-03/Shared-Pseudocode/Shared-Functions), there are `FPMin`/`FPMax` and `FPMinNum`/`FPMaxNum`. The former implements IEEE 754's `minimum`/`maximum` and the latter implements `minimumNumber`/`maximumNumber`.

## Rust

Rust exposes min/max operations that choose the non-NaN operand if possible, but are non-deterministic if both inputs are zero. For instance, [the `f32::min` documentation](https://doc.rust-lang.org/std/primitive.f32.html#method.min) states:

> In particular, if the inputs compare equal (such as for the case of `+0.0` and `-0.0`), either input may be returned non-deterministically.

[In practice, on x86 this compiles down to a `minps` plus a `cmpunordps` to select the non-NaN operand. On AArch64, it compiles down to an `fminnm` (the `FPMinNum` abstract operation described above).](https://godbolt.org/z/vcxj115rz)

## fearless_simd?

Now we need to decide which semantics to expose.

Ideally, we would just expose IEEE 754-2019's operations. That's what the current tests assume we do, at least. Unfortunately, neither x86 nor WebAssembly implement `minimumNumber`/`maximumNumber` semantics. In fact, x86 doesn't even implement `minimum`/`maximum` semantics!

As #23 notes, there is value in min/max operations that can "massage" a NaN into some other value. For example, if you're keeping a running tally of "lowest/highest value seen so far", it's helpful to be able to ignore NaNs.

We *could* implement the following operations:

- "Logical" minimum/maximum. These behave like `if a < b { a } else { b }`/`if a > b { a } else { b }`. The asymmetry seems a bit ugly, but most min/max use cases have some inherent asymmetry anyway (the "running tally", for instance, or the Vello use case that prompted #23). These are optimal on x86, but compile to a compare/select on AArch64.

- "Precise" minimum/maximum. These are the IEEE 754 `minimumNumber`/`maximumNumber` operations. These are optimal on AArch64, but will require quite a bit of polyfilling on x86, especially with regards to signed zero. These will also be difficult in the fallback implementation, since Rust doesn't expose the operations we need. [WebAssembly doesn't expose these either.](https://github.com/WebAssembly/design/issues/1548)

- "Precise with NaN" minimum/maximum. These are the IEEE 754 `minimum`/`maximum` operations. These are optimal on AArch64, but will require quite a bit of polyfilling on x86, especially with regards to signed zero. These will also be difficult in the fallback implementation, since Rust doesn't expose the operations we need.

- "Rust-y" minimum/maximum. These match Rust's semantics of returning whichever operand is non-NaN if the other is NaN, but makes no guarantees about signed zero. These are optimal on AArch64, but require an additional `cmpunord` and `blendv` on x86. These will be very easy to implement in the fallback implementation.

- "Relaxed" minimum/maximum. These behave like the "logical" "precise", or "Rust-y" operations, depending on whichever is fastest on the target architecture. That means that if the *second* operand is non-NaN, the result is guaranteed to never be NaN. These are optimal on x86, requiring only a `minps`/`maxps` operation, but won't let us use `fmin`/`fmax` on AArch64. I'm not sure if that's a big deal; we'll probably need to benchmark `fmin`/`fmax` versus `fminnm`/`fmaxnm`.

- "Fast" minimum/maximum. These behave like the "logical", "precise with NaN", or "Rust-y" operations, depending on whichever is fastest on the target architecture. These are optimal on both x86 and AArch64, but NaN values can easily pollute the result.

Implementing all of those would be tedious and result in quite an unwieldy API surface. We therefore need to decide which subset is most useful. It's unfortunate that each fully-deterministic operation (logical, precise, and precise with NaN) is slow on at least one target architecture.

I think that the best set of operations to expose is "logical" and "relaxed". The "logical" min/max is deterministic, and the "relaxed" min/max is fast(er). This doesn't give us *any* operations that map to `fmin`/`fmax` on AArch64, but I don't know if `fminnm`/`fmaxnm` are actually slower.

valadaptive:
Resolves #133. Depends on #134, since I use a helper method I introduced in that PR.

In #133, I brought up the question of what semantics we should use for floating-point minimum and maximum operations. There were many different options available, all with annoying tradeoffs on different architectures.

~~I believe the conclusion we came to during office hours was to implement the following:~~

- `min` and `max`:
  - If the first operand is NaN, return the second operand.
  - If the second operand is NaN, return either the first operand or NaN non-deterministically.
  - Otherwise, if both operands are zero, regardless of sign, return either non-deterministically. 
  - Otherwise, return the minimum/maximum of the two operands.

These can be implemented using a single `minps`/`maxps`/`minpd`/`maxpd` on x86 and the `pmin` and `pmax` operations in WebAssembly[^1], both of which return the first operand if the second operand is NaN.

On AArch64 and (eventually) RISC-V, these can be implemented using `fminnm`/`fmaxnm` (AArch64) and `fmin`/`fmax` (RISC-V), both of which return the non-NaN operand if *either* the first or second operand is NaN.

- `min_precise` and `max_precise`:
  - If the first operand is NaN, return the second operand.
  - If the second operand is NaN, return the first operand.
  - Otherwise, if both operands are zero, regardless of sign, return either non-deterministically. 
  - Otherwise, return the minimum/maximum of the two operands.

On x86 and WebAssembly, these can be implemented using the looser "return second operand if first is NaN" semantics, followed by a check/select to pick the first operand if the second operand is NaN.

On AArch64 and RISC-V, these are implemented using the exact same instructions as the regular `min` and `max`.

(EDIT: per a discussion on Zulip, we've decided to allow the non-precise min/max to return NaN if either operand is NaN. This means some algorithms might be forced to use the slower min_precise/max_precise on x86 even if it's unnecessary, but could be faster on WebAssembly.)

---

As I noted in #133, there are some tradeoffs. For the chosen semantics, WebAssembly draws the short straw.

WebAssembly exposes two different min/max operations. The `pmin`/`pmax` operations return the first operand if the second is NaN[^1], and the `min`/`max` operations return NaN if *either* operand is NaN.

`pmin` and `pmax` are efficient on x86, but require a compare/select on AArch64 and probably RISC-V as well.

`min`/`max` are efficient on AArch64 (they map to the `fmin`/`fmax` instructions). On x86, they're quite slow since their behavior needs to be polyfilled, including signed-zero handling. They're probably also slow on RISC-V, which doesn't have *any* instructions for "min/max, but NaN if either operand is NaN".

~~WebAssembly's "relaxed SIMD" extension (now supported on all platforms) provides "relaxed min/max" operations which non-deterministically perform a `min`/`max` or a `pmin`/`pmax`. By requiring our non-precise `min`/`max` to return a non-NaN value if the second operand is non-NaN, we cannot use those relaxed operations.~~

(EDIT: The relaxed SIMD extension allows for a wider range of results! Rust's documentation for these intrinsics was just incorrect.)

If we instead chose *not* to guarantee semantics around NaN, and allowed our non-precise `min` and `max` to return NaN if either operand is NaN, we could make use of WebAssembly's "relaxed SIMD" operations. However, this would lose performance in the case where client code knows that one operand is non-NaN.

As I noted in #133, and as the Vello code motivating #23 demonstrates, consumers of fearless_simd often know that one operand is non-NaN. In that case, they can still use non-precise `min`/`max` and get efficient code on both x86 and AArch64. If we loosened the non-precise `min`/`max` semantics, however, they would be forced to use the precise versions, costing an extra `cmpunord` and `blendv` on x86 that are completely unnecessary.

~~As I mentioned, RISC-V doesn't expose *any* instructions directly corresponding to WebAssembly's `min`/`max`. [There's an open issue about adding such operations to WebAssembly](https://github.com/WebAssembly/design/issues/1548), but even if they were, the "relaxed SIMD" proposal has already been ratified and couldn't make use of them.~~

~~Since WebAssembly has already locked itself into a design that will be unavoidably slow on RISC-V, I'm willing to give them the short end of the stick when it comes to performance.~~

[^1]: WebAssembly actually flips the operands so that it returns the *second* operand if the *first* is NaN, but we can just flip them back.

### DIFF
--- a/fearless_simd/src/generated/neon.rs
+++ b/fearless_simd/src/generated/neon.rs
@@ -136,11 +136,11 @@ impl Simd for Neon {
     }
     #[inline(always)]
     fn max_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        unsafe { vmaxnmq_f32(a.into(), b.into()).simd_into(self) }
+        unsafe { vmaxq_f32(a.into(), b.into()).simd_into(self) }
     }
     #[inline(always)]
     fn min_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        unsafe { vminnmq_f32(a.into(), b.into()).simd_into(self) }
+        unsafe { vminq_f32(a.into(), b.into()).simd_into(self) }
     }
     #[inline(always)]
     fn max_precise_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
@@ -1124,11 +1124,11 @@ impl Simd for Neon {
     }
     #[inline(always)]
     fn max_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
-        unsafe { vmaxnmq_f64(a.into(), b.into()).simd_into(self) }
+        unsafe { vmaxq_f64(a.into(), b.into()).simd_into(self) }
     }
     #[inline(always)]
     fn min_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
-        unsafe { vminnmq_f64(a.into(), b.into()).simd_into(self) }
+        unsafe { vminq_f64(a.into(), b.into()).simd_into(self) }
     }
     #[inline(always)]
     fn max_precise_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -125,11 +125,11 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn max_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        f32x4_pmax(b.into(), a.into()).simd_into(self)
+        f32x4_max(a.into(), b.into()).simd_into(self)
     }
     #[inline(always)]
     fn min_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        f32x4_pmin(b.into(), a.into()).simd_into(self)
+        f32x4_min(a.into(), b.into()).simd_into(self)
     }
     #[inline(always)]
     fn max_precise_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
@@ -1087,11 +1087,11 @@ impl Simd for WasmSimd128 {
     }
     #[inline(always)]
     fn max_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
-        f64x2_pmax(b.into(), a.into()).simd_into(self)
+        f64x2_max(a.into(), b.into()).simd_into(self)
     }
     #[inline(always)]
     fn min_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
-        f64x2_pmin(b.into(), a.into()).simd_into(self)
+        f64x2_min(a.into(), b.into()).simd_into(self)
     }
     #[inline(always)]
     fn max_precise_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
--- a/fearless_simd_gen/src/arch/neon.rs
+++ b/fearless_simd_gen/src/arch/neon.rs
@@ -30,6 +30,8 @@ fn translate_op(op: &str) -> Option<&'static str> {
         "shr" => "vshl",
         "shrv" => "vshl",
         "shl" => "vshl",
+        "max_precise" => "vmaxnm",
+        "min_precise" => "vminnm",
         "madd" => "vfma",
         "msub" => "vfms",
         _ => return None,
--- a/fearless_simd_gen/src/mk_neon.rs
+++ b/fearless_simd_gen/src/mk_neon.rs
@@ -263,21 +263,6 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                                 #vbsl(sign_mask, b.into(), a.into()).simd_into(self)
                             }
                         }
-                        "min" | "max" | "min_precise" | "max_precise"
-                            if vec_ty.scalar == ScalarType::Float =>
-                        {
-                            let intrinsic = simple_intrinsic(
-                                if method == "max" || method == "max_precise" {
-                                    "vmaxnm"
-                                } else {
-                                    "vminnm"
-                                },
-                                vec_ty,
-                            );
-                            quote! {
-                                #intrinsic(a.into(), b.into()).simd_into(self)
-                            }
-                        }
                         _ => {
                             let args = [quote! { a.into() }, quote! { b.into() }];
                             let expr = neon::expr(method, vec_ty, &args);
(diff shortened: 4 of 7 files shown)


fearless_simd_dev_macros/src/lib.rs
```diff
@@ -128,17 +128,14 @@ fn exclude_fallback(_test_name: &str) -> bool {
...
 }
 
-fn exclude_wasm(test_name: &str) -> bool {
-    matches!(
-        test_name,
-        "min_precise_f32x4_with_nan" | "max_precise_f32x4_with_nan"
-    )
+fn exclude_wasm(_test_name: &str) -> bool {
+    false
 }
```
valadaptive, line 145: I can remove these in a future PR.

ajakubowicz-canva:
> For the chosen semantics, WebAssembly draws the short straw.


hahaha! Classic WebAssembly. Great PR!

PR #137, opened 2025-11-21
Add `SimdBase::from_fn`

### LINKED ISSUE #132: `from_fn` method for `SimdBase` types, mirroring `core::array::from_fn`
The title says it all. Basically just [`core::array::from_fn`](https://doc.rust-lang.org/stable/core/array/fn.from_fn.html), but for `SimdBase` types, and the function is passed the lane number. This is useful for e.g. operating over rows in parallel, where you need to know the index of each row.

valadaptive:
Resolves #132. Does what it says on the tin / in the docs.

Not sure how much sense this makes for the mask types, but they already have `from_slice` and `splat`. Maybe that should be a future issue.


LaurenzV (approved):
Makes sense!

PR #143, opened 2025-11-24
Use WebAssembly's relaxed SIMD intrinsics if available

### LINKED ISSUE #142: Make use of the WebAssembly "relaxed SIMD" operations
The [WebAssembly relaxed SIMD proposal](https://github.com/WebAssembly/relaxed-simd) has been standardized, and is apparently [now shipping in Chrome 114 and Node 21](https://webassembly.org/features/). As such, we should make use of it within fearless_simd for increased performance.

As with SIMD128, it's a compile-time-only feature; people will have to ship separate binaries with it enabled. As such, I think the best option is not to expose it as its own feature level, but instead transparently use it if the requisite target feature is enabled.

valadaptive:
Resolves #142.

This PR makes use of WASM's new "relaxed SIMD" intrinsics if the corresponding target feature is enabled. I opted against making it a new SIMD level, since most operations are the exact same and WASM doesn't have runtime feature detection anyway.

The relevant intrinsics for us are the relaxed fused multiply-add (performs an FMA if the hardware supports it, or a regular multiply followed by add if not), and relaxed "lane select" (selects based on the high bit of each lane as x86 prefers, or bitwise as AArch64 prefers, depending on the hardware). (EDIT: Following https://github.com/linebender/fearless_simd/pull/136, we can also use relaxed min/max operations for floats.)

It does mean we need to double our tests; maybe it would be worth adding a `.cargo/config.toml` to remove some of the boilerplate?

### DIFF
--- a/fearless_simd/README.md
+++ b/fearless_simd/README.md
@@ -112,9 +112,12 @@ TODO: Talk about writing versions of functions which can be called in other `S: 
 # WebAssembly
 
 WASM SIMD doesn't have feature detection, and so you need to compile two versions of your bundle for WASM, one with SIMD and one without,
-then select the appropriate one for your user's browser. This can be done via [the `wasm-feature-detect` library](https://github.com/GoogleChromeLabs/wasm-feature-detect).
+then select the appropriate one for your user's browser. This can be done via [the `wasm-feature-detect`
+library](https://github.com/GoogleChromeLabs/wasm-feature-detect).
 
-You can compile WebAssembly with the SIMD128 feature enabled via the `RUSTFLAGS` environment variable (`RUSTFLAGS="-Ctarget-feature=+simd128"`), or by adding the compiler flags in your [Cargo config.toml](https://doc.rust-lang.org/cargo/reference/config.html):
+You can compile WebAssembly with the SIMD128 feature enabled via the `RUSTFLAGS` environment variable
+(`RUSTFLAGS="-Ctarget-feature=+simd128"`), or by adding the compiler flags in your [Cargo
+config.toml](https://doc.rust-lang.org/cargo/reference/config.html):
 
 ```toml
 [target.'cfg(target_arch = "wasm32")']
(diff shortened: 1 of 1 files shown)


fearless_simd_gen/src/mk_wasm.rs
```diff
@@ -164,8 +164,25 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                             quote! {sub}
                         };
 
-                        // TODO: `relaxed-simd` has madd.
+                        let c = if method == "msub" {
+                            // WebAssembly just... forgot fused multiply-subtract? It seems the
+                            // initial proposal
+                            // (https://github.com/WebAssembly/relaxed-simd/issues/27) confused it
+                            // with negate multiply-add, and nobody ever resolved the confusion.
```
valadaptive, line 204: Seriously, what's going on here? Am I missing something? Negate multiply-add is clearly not the same thing as multiply-subtract. Surely this isn't a mistake that sailed through to standardization after years of spec discussion, and I'm missing some rationale for omitting fused multiply-subtract?

ajakubowicz-canva: Haha - it does indeed look like multiply-subtract was converted into [𝗋𝖾𝗅𝖺𝗑𝖾𝖽_𝗇𝗆𝖺𝖽𝖽](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-vec).

[I left a comment since I am also curious. ](https://github.com/WebAssembly/relaxed-simd/issues/27#issuecomment-3587681828)

ajakubowicz-canva (approved):
Excellent!

fearless_simd/src/lib.rs
ajakubowicz-canva, line 1: I love this doc expansion. At the risk of scope creep, it could also be valuable to touch on the fact that loading WASM with unsupported instructions for a given browser results in a crash at WASM validation time even if that code path is never run.

Non blocking though.

ajakubowicz-canva (approved):
Doc improvements are amazing!

PR #148, opened 2025-11-29
Clean up the code generation a bit more

valadaptive:
I'm trying to move towards defining everything about a SIMD operation in one place. In order to make every operation maximally descriptive, I've made some changes (in order):

- I've changed the `OpSig` enum to have struct variants instead of tuple variants, and standardized the naming of those variant fields across the codegen targets.
- I've centralized the trait-based ops for each scalar type within one place: the new `CoreOpTrait` enum. This shuffles around the order of some of the impls and trait bounds, but ensures they're all defined in the same order. There is one functional API change here: the `SimdMask` trait now has all the trait bounds for bitwise operations. These traits were already implemented for every mask type, and I believe their omission was unintentional.
- I've added documentation for every `OpSig` variant.

### DIFF
--- a/fearless_simd_gen/src/generic.rs
+++ b/fearless_simd_gen/src/generic.rs
@@ -212,10 +212,10 @@ pub(crate) fn generic_op(op: &str, sig: OpSig, ty: &VecType) -> TokenStream {
             }
         }
         OpSig::Reinterpret {
-            target_ty: target,
+            target_ty,
             scalar_bits,
         } => {
-            let mut half = reinterpret_ty(ty, target, scalar_bits);
+            let mut half = reinterpret_ty(ty, target_ty, scalar_bits);
             half.len /= 2;
             let combine = Ident::new(&format!("combine_{}", half.rust_name()), Span::call_site());
             quote! {
--- a/fearless_simd_gen/src/mk_wasm.rs
+++ b/fearless_simd_gen/src/mk_wasm.rs
@@ -387,7 +387,7 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                         }
                     }
                 }
-                OpSig::WidenNarrow { target_ty: target } => {
+                OpSig::WidenNarrow { target_ty } => {
                     match method {
                         "widen" => {
                             assert_eq!(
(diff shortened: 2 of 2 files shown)


fearless_simd_gen/src/generic.rs
```diff
@@ -208,8 +211,11 @@ pub(crate) fn generic_op(op: &str, sig: OpSig, ty: &VecType) -> TokenStream {
                 }
             }
         }
-        OpSig::Reinterpret(scalar, scalar_bits) => {
-            let mut half = reinterpret_ty(ty, scalar, scalar_bits);
+        OpSig::Reinterpret {
+            target_ty: target,
```
LaurenzV: Perhaps just keep the `target_ty` name here?

fearless_simd_gen/src/ops.rs
```diff
@@ -8,27 +8,55 @@ use crate::types::{ScalarType, VecType};
...
-    // NEON instructions `vld4q_f32`, while `LoadInterleaved(64, 4)` would correspond to
-    // `vld4_f32`.
-    LoadInterleaved(u16, u16),
-    StoreInterleaved(u16, u16), // TODO: fma
+    /// Takes an argument of an array of a certain scalar type, with the length (`block_size` * `block_count`) / [scalar
+    /// type's byte size]. Returns a vector type of that scalar type and length.
+    ///
+    /// First argument is the base block size (i.e. 128), second argument is how many blocks. For example,
+    /// `LoadInterleaved(128, 4)` would correspond to the NEON instructions `vld4q_f32`, while `LoadInterleaved(64, 4)`
+    /// would correspond to `vld4_f32`.
```
LaurenzV, line 55: I don't believe we are actually using 64, 4 anywhere right? It's fine to keep in the documentation, just wanted to double check.

fearless_simd_gen/src/mk_wasm.rs
```diff
@@ -381,7 +387,7 @@ fn mk_simd_impl(level: Level) -> TokenStream {
                         }
                     }
                 }
-                OpSig::WidenNarrow(to_ty) => {
+                OpSig::WidenNarrow { target_ty: target } => {
```
LaurenzV: Same here, rename is probably not necessary?

fearless_simd_gen/src/generic.rs
valadaptive: Good catch. I originally named the field `target`, but renamed it to `target_ty` to avoid potential confusion with the concept of a target architecture. I tried to rename all its uses as well, but I missed this (and the other one).

fearless_simd_gen/src/ops.rs
valadaptive: Correct. The block size is 128 for everything we generate. I believe the block count is also always 4.

PR #149, opened 2025-11-29
Further tidy up codegen and somewhat rework the API

valadaptive:
Depends on #148.

This PR intertwines codegen and API changes, unfortunately. Some of the codegen cleanups depend on it.

All the inherent impl methods on individual SIMD types are now gone entirely. Most of them were previously already accessible via the `core::ops` traits, `SimdBase`/`SimdInt`/`SimdFloat`/`SimdMask`, or individual bespoke traits. The exceptions were split/combine. There was no inherent `split` method at all, and I've removed the inherent `combine` method. In its place I've added new `SimdCombine` and `SimdSplit` traits, which are implemented for types of the proper vector width.

Since all the operations need to be accessed via trait methods, I've added a `prelude` module to `fearless_simd`, which re-exports *just* the traits. This lets API consumers import every necessary trait to access all the methods, without being forced to either import every trait individually or just wildcard-import every single item from `fearless_simd`.

I'm not entirely happy with the current `SimdCvtTruncate` and `SimdCvtFloat` traits--it's unclear which is the source and which is the destination, `to_int` and `to_float` are *technically* implemented for all `SimdFloat` and `SimdInt` types even though they only *work* on 32-bit ones, and they require passing in the full vector equivalent (with `S` type parameter) as a type. Improving that is future work.

Aside from all that, this PR is about further centralizing the codegen in preparation for adding documentation generation. I want to make it as operation-centric as possible, moving op-specific special cases out of each module generation method and into the operation metadata itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This release has an [MSRV][] of 1.88.
 - The `from_fn` method on vector types. ([#137][] by [@valadaptive][])
 - The `load_interleaved` and `store_interleaved` operations now use native intrinsics on x86, instead of using the fallback implementations. ([#140][] by [@valadaptive][])
 - The `ceil` and `round_ties_even` operations on floating-point vector types. (Rust's `round` operation rounds away from zero in the case of ties. Many architectures do not natively implement that behavior, so it's omitted.) ([#145][] by [@valadaptive][])
+- A `prelude` module, which exports all the traits in the library but not the types. ([#149][] by [@valadaptive][])
 
 ### Fixed
 
@@ -44,6 +45,17 @@ This release has an [MSRV][] of 1.88.
 - Code generation for `select` and `unzip` operations on x86 has been improved. ([#115][] by [@valadaptive][])
 - Breaking change: The native-width associated types (`f32s`, `u8s`, etc.) for the `Avx2` struct have been widened from 128-bit
   types (like `f32x4`) to 256-bit types (like `f32x8`). ([#123][] by [@valadaptive][])
+- Breaking change: All the vector types' inherent methods have been removed. Any remaining functionality has been moved
+  to trait methods. ([#149][] by [@valadaptive][])
+
+  Some functionality is exposed under different names:
+  - Instead of the `reinterpret` methods, use the `bitcast` method on the `Bytes` trait. (e.g. `foo.reinterpret_i32()`
+    -> `foo.bitcast::<i32x4<_>>()`)
+  - Instead of the `cvt` methods, use the `to_int` or `to_float` convenience methods on the `SimdFloat` and `SimdInt`
+    traits (e.g. `foo.cvt_u32()` -> `foo.to_int::<u32x4<_>>()`)
+
+  Some functionality (such as `split` or `combine`) is exposed under new traits. You may use the new `prelude` module to
+  conveniently import all of the traits.
 
 ### Removed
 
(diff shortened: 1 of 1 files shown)


fearless_simd_gen/src/ops.rs
```diff
@@ -59,130 +59,238 @@ pub(crate) enum OpSig {
...
+pub(crate) fn overloaded_ops_for(scalar: ScalarType) -> Vec<CoreOpTrait> {
+    let base = match scalar {
+        ScalarType::Float => FLOAT_OPS,
+        ScalarType::Int | ScalarType::Unsigned => INT_OPS,
+        ScalarType::Mask => MASK_OPS,
+    };
+    // We prepend the negate operation only for signed integer types.
+    (scalar == ScalarType::Int)
+        .then_some(CoreOpTrait::Neg)
+        .into_iter()
```
ajakubowicz-canva, line 232: Clever & TIL

fearless_simd_gen/src/ops.rs
```diff
@@ -201,84 +310,196 @@ pub(crate) fn ops_for_type(ty: &VecType, cvt: bool) -> Vec<(&str, OpSig)> {
...
+                scalar_bits: 32,
+            },
+        )),
+        (ScalarType::Int, 32) => ops.push(Op::new(
+            "cvt_f32",
+            OpKind::OwnTrait,
+            OpSig::Cvt {
+                target_ty: ScalarType::Float,
+                scalar_bits: 32,
+            },
```
ajakubowicz-canva, line 396: No action - just an observation that it looks like the `method` string could almost be derived from the other data.

fearless_simd_tests/tests/mod.rs
```diff
@@ -74,11 +74,11 @@ fn test_f32_to_i32_exhaustive<S: Simd>(simd: S) {
         || {
             for i in (0..u32::MAX).step_by(4) {
                 let floats = f32x4::from_fn(simd, |n| f32::from_bits(n as u32 + i));
-                let ints = floats.cvt_i32().val;
+                let ints = floats.to_int::<i32x4<_>>().val;
```
ajakubowicz-canva, line 77: Does this API change require a changelog?

ajakubowicz-canva (approved):
I really love the move towards operation centric generation.

> I'm not entirely happy with the current SimdCvtTruncate and SimdCvtFloat traits--it's unclear which is the source and which is the destination, to_int and to_float are technically implemented for all SimdFloat and SimdInt types even though they only work on 32-bit ones, and they require passing in the full vector equivalent (with S type parameter) as a type. Improving that is future work.

Is there an open issue for this? What is the current behavior when called on non 32 bit variants? Is it a panic?

fearless_simd_gen/src/ops.rs
valadaptive: For the `cvt` operations it can be, but a lot of the `OpSig` types are more generic (`Unary`, `Binary`, etc) and can't. In a followup, it might be worth putting the op name into the specific `OpSig` variants for which it can't be derived.

fearless_simd_tests/tests/mod.rs
valadaptive: All the "fully switch to trait methods" stuff will require a changelog entry. I don't like to add them too early because they end up causing merge conflicts with other changelog additions, but it might be worth adding them now.

valadaptive:
> I really love the move towards operation centric generation.
> 
> > I'm not entirely happy with the current SimdCvtTruncate and SimdCvtFloat traits--it's unclear which is the source and which is the destination, to_int and to_float are technically implemented for all SimdFloat and SimdInt types even though they only work on 32-bit ones, and they require passing in the full vector equivalent (with S type parameter) as a type. Improving that is future work.
> 
> Is there an open issue for this? What is the current behavior when called on non 32 bit variants? Is it a panic?

It's a type error. The issue is more a *documentation* one. For instance, `f64x2` implements `SimdFloat` and hence has a `to_int` method that shows up in documentation and autocomplete, but there's no valid type to actually call it with.

PR #154, opened 2025-12-05
Generate documentation for operations and vector types

valadaptive:
Now that we've moved towards a more operation-centric design for the codegen, we can more easily add documentation for each operation.

I've updated the codegen to add documentation strings for every operation, as defined on both the `Simd` trait and the vector types. There's some basic templating to deal with the argument names being different between the two.

I've also added documentation for each vector type, and filled in some basic docs for the manually-defined traits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This release has an [MSRV][] of 1.88.
 - The `ceil` and `round_ties_even` operations on floating-point vector types. (Rust's `round` operation rounds away from zero in the case of ties. Many architectures do not natively implement that behavior, so it's omitted.) ([#145][] by [@valadaptive][])
 - A `prelude` module, which exports all the traits in the library but not the types. ([#149][] by [@valadaptive][])
 - The `any_true`, `all_true`, `any_false`, and `all_false` methods on mask types. ([#141][] by [@valadaptive][])
+- Documentation for most traits, vector types, and operations. ([#154][] by [@valadaptive][])
 
 ### Fixed
 
@@ -146,6 +147,7 @@ No changelog was kept for this release.
 [#141]: https://github.com/linebender/fearless_simd/pull/141
 [#145]: https://github.com/linebender/fearless_simd/pull/145
 [#149]: https://github.com/linebender/fearless_simd/pull/149
+[#154]: https://github.com/linebender/fearless_simd/pull/154
 
 [Unreleased]: https://github.com/linebender/fearless_simd/compare/v0.3.0...HEAD
 [0.3.0]: https://github.com/linebender/fearless_simd/compare/v0.3.0...v0.2.0


fearless_simd_gen/src/ops.rs
```diff
@@ -96,143 +98,461 @@ pub(crate) enum OpKind {
...
+        OpKind::VecTraitMethod,
+        OpSig::Ternary,
+        "Compute `({arg0} * {arg1}) + {arg2}` (fused multiply-add) for each element.\n\n\
+        Depending on hardware support, the result may be computed with only one rounding error, or may be implemented as a regular multiply followed by an add, which will result in two rounding errors.",
+    ),
+    Op::new(
+        "msub",
+        OpKind::VecTraitMethod,
+        OpSig::Ternary,
+        "Compute `({arg0} * {arg1}) + {arg2}` (fused multiply-subtract) for each element.\n\n\
```
LaurenzV: Should be a minus probably?

fearless_simd/src/generated/ops.rs
```diff
@@ -6209,363 +7111,418 @@ impl<S: Simd> core::ops::ShrAssign<u32> for u32x16<S> {
...
     }
 }
 impl<S: Simd> core::ops::Sub<f64> for f64x8<S> {
     type Output = Self;
+    #[doc = "Subtract two vectors element-wise."]
     #[inline(always)]
     fn sub(self, rhs: f64) -> Self::Output {
         self.simd.sub_f64x8(self, rhs.simd_into(self.simd))
     }
 }
```
DJMcNab, line 6830: The new doc line isn't really correct; it's more like `"Subtract `rhs` from each element of self."` or similar.
The same also applies to `Add<[Scalar]>`, and maybe other traits as well?

I'm perfectly happy for this to just because a TODO, because I imagine it gets more complicated still to add this.

fearless_simd_gen/src/ops.rs
```diff
@@ -96,143 +98,461 @@ pub(crate) enum OpKind {
...
-    Op::new("not", OpKind::Overloaded(CoreOpTrait::Not), OpSig::Unary),
-    Op::new("shl", OpKind::Overloaded(CoreOpTrait::Shl), OpSig::Shift),
-    Op::new("shr", OpKind::Overloaded(CoreOpTrait::Shr), OpSig::Shift),
     Op::new(
         "shrv",
         OpKind::Overloaded(CoreOpTrait::ShrVectored),
         OpSig::Binary,
+        "Shift each element right by the corresponding element in another vector.\n\n\
+        For unsigned integers, zeros are shifted in on the left. For signed integers, the sign bit is replicated.\n\n\
+        This operation is not implemented in hardware on all platforms. On WebAssembly, and on x86 platforms without AVX2, this will use a fallback scalar implementation.",
```
LaurenzV, line 404: I'm wondering whether we should "leak" those implementation details in the docs?

fearless_simd/src/generated/ops.rs
```diff
@@ -12,357 +12,411 @@ use crate::{
...
 impl<S: Simd> core::ops::Add for f32x4<S> {
     type Output = Self;
+    #[doc = "Add two vectors element-wise."]
     #[inline(always)]
     fn add(self, rhs: Self) -> Self::Output {
         self.simd.add_f32x4(self, rhs)
     }
 }
 impl<S: Simd> core::ops::AddAssign for f32x4<S> {
+    #[doc = "Add two vectors element-wise."]
```
LaurenzV, line 30: I'm wondering whether these two should have different doc comments? Non-blocking though

fearless_simd_gen/src/ops.rs
valadaptive: Good catch!

fearless_simd_gen/src/ops.rs
valadaptive: I think it's good to warn people about this particular performance cliff since it becomes completely scalar.

fearless_simd/src/generated/ops.rs
valadaptive: Perhaps at some point, but it'd be tricky.

PR #155, opened 2025-12-05
Add a vectored shift left operation

valadaptive:
For symmetry with the existing vectored shift right operation. I took the opportunity to clean up the codegen a bit, moving the x86 implementation entirely into `mk_avx2` since SSE4.2 just uses the fallback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This release has an [MSRV][] of 1.88.
 - A `prelude` module, which exports all the traits in the library but not the types. ([#149][] by [@valadaptive][])
 - The `any_true`, `all_true`, `any_false`, and `all_false` methods on mask types. ([#141][] by [@valadaptive][])
 - Documentation for most traits, vector types, and operations. ([#154][] by [@valadaptive][])
+- A "shift left by vector" operation, to go with the existing "shift right by vector". ([#155][] by [@valadaptive][])
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


fearless_simd/src/generated/sse4_2.rs
```diff
@@ -786,6 +798,10 @@ impl Simd for Sse4_2 {
         unsafe { _mm_sll_epi16(a.into(), _mm_cvtsi32_si128(shift as _)).simd_into(self) }
     }
     #[inline(always)]
+    fn shlv_u16x8(self, a: u16x8<Self>, b: u16x8<Self>) -> u16x8<Self> {
+        core::array::from_fn(|i| core::ops::Shl::shl(a.val[i], b.val[i])).simd_into(self)
```
LaurenzV, line 802: Can't we use `_mm_sll_epi16` here (also for i16)?

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -969,6 +985,10 @@ impl Simd for Sse4_2 {
         unsafe { _mm_sll_epi32(a.into(), _mm_cvtsi32_si128(shift as _)).simd_into(self) }
     }
     #[inline(always)]
+    fn shlv_i32x4(self, a: i32x4<Self>, b: i32x4<Self>) -> i32x4<Self> {
+        core::array::from_fn(|i| core::ops::Shl::shl(a.val[i], b.val[i])).simd_into(self)
```
LaurenzV, line 989: Can't we use _mm_sll_epi32 here (also for u32)?

LaurenzV (approved):

fearless_simd/src/generated/sse4_2.rs
valadaptive: The `_mm_sll_*` functions use the same shift amount for every lane--the `count` argument is a vector type, but it only actually looks at its lowest lane and uses that as the shift amount everywhere. Actual vectored shift operations only exist in AVX2, and only for 32-bit and 64-bit types (it would be good to mention that in the documentation too, but type-specific documentation is another todo list item).

LaurenzV: TIL... Maybe add a comment for that in fearless_simd_gen? Otherwise, LGTM!

Issue #150, opened 2025-12-03
Consider switching to intrinsic-specific SIMD types


valadaptive:
Right now, the various SIMD types (`f32x4`, etc.) are represented as `#[repr(align)]`ed arrays. The *native* vector types on the other hand (e.g. `__m128` on x86, `float32x4_t` on AArch64), are defined in Rust itself as experimental `#[repr(simd)]` types.

My (completely untested) hypothesis is that the `#[repr(simd)]` native vector types are lowered directly to LLVM's vector types. The arrays, on the other hand, have to be autovectorized in an optimization pass. This usually happens, but is not guaranteed. If this hypothesis is true, that would explain why the compiler does things like [pessimize `simd_from` to a memset](https://github.com/linebender/fearless_simd/blob/45ae1187499d87402fc56d4fe1934fef68021f72/fearless_simd_gen/src/mk_simd_types.rs#L113-L118) in some cases.

We should verify this hypothesis, and switch to using the native vector types if so. I'm thinking that the `Simd` trait would define an associated type for each vector type, and the architecture-specific implementations would define the associated types to be the native vector types (or arrays of the native vector types, for ones wider than are natively supported).

valadaptive:
I've thought about it further, and there is in fact an annoying restriction with switching entirely to associated types. All the various operations we want to implement on each vector type would have to be reimplemented for each *specific* associated type--you can't just define a trait for all vector types and blanket-implement all the desired traits on it due to coherence rules.

I'm prototyping a design that makes the associated native vector types more of an implementation detail, with operations to convert them to/from arrays implemented as associated methods on `Simd`, just like all the other operations.

PR #159, opened 2025-12-09
Use native vector types to back SIMD types

valadaptive:
Resolves https://github.com/linebender/fearless_simd/issues/150.

In #150, I began wondering if this would involve fully switching the vector types over to being associated types on the `Simd` trait. For trait coherence reasons, it turns out we don't want to do that. As such, there *is* a meaningful reason to keep the generic vector types around, and this change is much less intrusive than I'd previously thought.

This PR switches the SIMD vector types (`f32x4<S>`, etc) from storing aligned arrays to storing architecture-specific vector types (e.g. `__m128`, `float32x4_t`, `v128`). For vectors wider than the maximum native vector width, arrays of vector types are used (e.g. `[__m128; 4]`). As I mentioned in #150, if we use arrays, we're essentially relying on the autovectorizer to consistently turn array operations into vector operations. The main point of writing a SIMD library is to not do that. An example of this going wrong is [the workaround introduced to prevent LLVM from pessimizing a vector load into a `memset`](https://github.com/linebender/fearless_simd/blob/bcdc8b390689076be4349d64e4686e52e3b3c384/fearless_simd_gen/src/mk_simd_types.rs#L144-L149). This workaround seems to have *prevented* some scalar Vello code from being autovectorized.

To ensure that the arrays of native vector types are aligned to the level they should be, they're wrapped in generic alignment-increasing types (`Aligned128`, `Aligned256`, and `Aligned512`). I'm not sure if this lets us remove the alignment attribute from the public-facing vector types, so I've left it there.

The split/combine operations now directly operate on these wrapped arrays instead of calling `copy_from_slice`. I've found that while the slice copies could *usually* be elided, they were not *always* elided. In addition, every single operation on wide vectors calls split/combine, sometimes recursively. Requiring the compiler to optimize out all those slice copies is both fragile and bad for compile times.

We do maintain the ability to treat the vector types as arrays. This requires quite a bit of transmuting, which is unfortunate. It also requires that the native vector types are in fact interpretable as arrays. For the x86 types, the Rust documentation [guarantees it](https://doc.rust-lang.org/core/arch/x86_64/struct.__m128.html), saying this about all of them:

> The in-memory representation of this type is the same as the one of an equivalent array (i.e. the in-memory order of elements is the same, and there is no padding); however, the alignment is different and equal to the size of the type. Note that the ABI for function calls may not be the same.

The AArch64 and WASM documentation is not explicit about layout, but their vector types *are* marked with `#[repr(simd)]`, which means the same thing. It'd probably be good to get a guarantee that the layout is part of the stable API surface, but I'm not sure who to ask. I've run the test suite on Miri and it passes on x86, but can't run on AArch64 or WebAssembly due to missing support for some intrinsics.

The operations for loading/storing vector types as arrays are now methods on the `Simd` trait, implemented by each specific platform, just like most other operations.

To avoid exposing the native vector types as part of the public API, I've defined them as associated types on a new `ArchTypes` trait, which is now a supertrait of `Simd`. The trait is not publicly exported. This means I had to silence some `unnameable_types` lints on both `ArchTypes` and the aligned wrapper types, which are also an internal implementation detail.

I believe the only public-facing change is that the `val` field on vector types is now private. You can still construct a vector type from an array via `SimdFrom`/`SimdInto`, and convert a vector type to an array (or a reference to one) via `Deref`/`DerefMut` or the `From`/`Into` conversions. I kind of wonder if we should go all the way and make the `simd` field private too, since you can access it via the `witness` method.

If you want to test the performance impact of these changes, you can run vello_bench on [this branch, without this PR](https://github.com/valadaptive/vello/tree/fearless-simd-bench-pre-native-vectors), and [this branch, which includes this PR](https://github.com/valadaptive/vello/tree/fearless-simd-native-vectors).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ This release has an [MSRV][] of 1.88.
   conveniently import all of the traits.
 - Breaking change: The `madd` and `msub` methods have been renamed to `mul_add` and `mul_sub`, matching Rust's naming conventions.
   ([#158][] by [@Shnatsel][])
+- Breaking change: the `val` field on SIMD vector types is now private, and vector types are no longer represented as arrays internally. To access a vector type's elements, you can use the `Into` or `Deref` traits to obtain an array, or the `as_slice`/`as_mut_slice` methods to obtain a slice. ([#159][] by [@valadaptive][])
 
 ### Removed
 
@@ -155,6 +156,7 @@ No changelog was kept for this release.
 [#154]: https://github.com/linebender/fearless_simd/pull/154
 [#155]: https://github.com/linebender/fearless_simd/pull/155
 [#158]: https://github.com/linebender/fearless_simd/pull/158
+[#159]: https://github.com/linebender/fearless_simd/pull/159
 [#167]: https://github.com/linebender/fearless_simd/pull/167
 
 [Unreleased]: https://github.com/linebender/fearless_simd/compare/v0.3.0...HEAD


LaurenzV:
Since that's a rather big change, I'm wondering whether we should maybe first get out a new release of fearless_simd with what we currently have and update vello_cpu to use that. 🤔

LaurenzV:
image
image

That's what I get in Instruments, but that doesn't seem very helpful...

fearless_simd/src/generated/avx2.rs
```diff
@@ -65,6 +103,50 @@ impl Simd for Avx2 {
         unsafe { _mm_set1_ps(val).simd_into(self) }
     }
     #[inline(always)]
+    fn load_array_f32x4(self, val: [f32; 4usize]) -> f32x4<Self> {
+        f32x4 {
+            val: unsafe { crate::support::Aligned128(_mm_loadu_ps(val.as_ptr() as *const _)) },
+            simd: self,
+        }
```
LaurenzV, line 110: Can we not just call `load_array_ref` here and pass a reference to the array?

LaurenzV: (Same for all other instances)

fearless_simd/src/generated/avx2.rs
```diff
@@ -1401,6 +1923,50 @@ impl Simd for Avx2 {
...
+    }
+    #[inline(always)]
+    fn as_array_mut_mask64x2(self, a: &mut mask64x2<Self>) -> &mut [i64; 2usize] {
+        unsafe { core::mem::transmute::<&mut __m128i, &mut [i64; 2usize]>(&mut a.val.0) }
+    }
+    #[inline(always)]
+    fn cvt_from_bytes_mask64x2(self, a: u8x16<Self>) -> mask64x2<Self> {
+        unsafe {
+            mask64x2 {
+                val: core::mem::transmute(a.val),
```
LaurenzV, line 1955: Just for my personal understanding, a transmute like this one is save because both types are wrapped in `Aligned128`, so they are guaranteed to have the same alignment, right? Otherwise, it probably wouldn't be save to cast something like [u8; 16] into [u64; 2]?

fearless_simd/src/generated/avx2.rs
```diff
@@ -5159,190 +6996,195 @@ impl Simd for Avx2 {
...
+            },
+        )
     }
 }
 impl<S: Simd> SimdFrom<__m256, S> for f32x8<S> {
     #[inline(always)]
     fn simd_from(arch: __m256, simd: S) -> Self {
         Self {
-            val: unsafe { core::mem::transmute(arch) },
+            val: unsafe { core::mem::transmute_copy(&arch) },
```
LaurenzV, line 7015: What's the reason for switching here?

fearless_simd_gen/src/arch/neon.rs
```diff
@@ -102,8 +101,7 @@ fn neon_array_type(ty: &VecType) -> (&'static str, &'static str, usize) {
 pub(crate) fn opt_q(ty: &VecType) -> &'static str {
     match ty.n_bits() {
         64 => "",
-        128 => "q",
-        _ => panic!("unsupported simd width"),
+        _ => "q",
```
LaurenzV: I think it makes sense to be more explicit here and panic if we encounter a unknown number of bits.

fearless_simd/src/generated/avx2.rs
LaurenzV: (already resolved)

fearless_simd/src/generated/avx2.rs
valadaptive: Yes; I've just now added that as a comment in the `generic_from_bytes` codegen function. Unfortunately, it won't show up in generated code.

fearless_simd/src/generated/avx2.rs
valadaptive: Because this is implemented for *all* `S: Simd`, we don't know the size of the `val` field anymore. `transmute` only lets you transmute between types of the same statically-known size, whereas `transmute_copy` doesn't have that restriction.

We cannot change these conversion impls to specific token types, such as `impl SimdFrom<__m256, Avx2> for f32x8<Avx2>`, because of downcasting. User code needs to be able to downcast any arbitrary `f32x8<S>` into `__m256`, since we provide no operations to "prove" that it's actually a `f32x8<Avx2>`.

fearless_simd_gen/src/arch/neon.rs
valadaptive: I've re-added the panic message.

fearless_simd/src/generated/avx2.rs
valadaptive: For posterity's sake, here's what I said on Zulip:

> I haven't actually checked if it generates different IR or assembly, but all the "load" intrinsics take a pointer. If I were to implement `from_slice` in terms of "load array by value", I'd have to deref the slice into an array, pass it into the load_array method, take a reference to the array I just deref'd, call the "load" intrinsic, and hope the compiler elides the copy. That's the exact type of operation I'm trying to avoid by switching to native vectors.

PR #162, opened 2025-12-11
Add mul_neg_add method to floats

Shnatsel:
This is needed for porting https://github.com/QuState/PHastFT/ to fearless_simd

Named after the equivalent method in `wide`: https://docs.rs/wide/latest/wide/struct.f32x8.html#method.mul_neg_add

AI use disclosure: this is generated wholesale by Claude Opus 4.5, using mul_add and mul_sub as reference. The structure is nearly identical, so I'm not terribly surprised that AI could manage it, even though it's still impressive. The result holds up to my review and the newly added tests pass, but I'm not deeply familiar with this crate so take that with a grain of salt.


valadaptive:
> I have to reopen this. While the regular `mul_add` optimizes fine on x86 (AVX2 and AVX-512 both), it does NOT automatically optimize on ARM (NEON). This was caused a massive regression for my FFT implementation because FFTs are just FMAs plus loads/stores.

Have you checked that this PR actually improves ARM codegen? The Rust intrinsics themselves lower to a separate negate + multiply-add (the file is too big to display in GitHub, but [here it is](https://github.com/rust-lang/stdarch/blob/main/crates/core_arch/src/arm_shared/neon/generated.rs)):

```rust
pub fn vfmsq_f32(a: float32x4_t, b: float32x4_t, c: float32x4_t) -> float32x4_t {
    unsafe {
        let b: float32x4_t = simd_neg(b);
        vfmaq_f32(a, b, c)
    }
}
```

Issue #29, opened 2025-07-09
Shuffle/swizzle operations


valadaptive:
I'd like to replace my [homegrown SIMD implementation](https://github.com/valadaptive/ntsc-rs/blob/93e533d2564d86b283ca22f119db34b538a33049/crates/ntscrs/src/f32x4.rs) with this crate. My use case requires support for a "shuffle" / "swizzle" / "permute" operation, which this crate currently doesn't appear to have.

One obstacle I encountered when implementing shuffles in my SIMD implementation is that different architectures' intrinsics take different types for shuffle indices (they differ in signedness and size), so we can't use const generics to specify them currently. However, LLVM appears to optimize constant permutes into shuffle instructions.

My thought is to give each SIMD type an associated `Indices` type (name not final), which is like its `Mask` or `Block` types, and is a `[usize; N]`. A shuffle function would take it as input.

AndrewJakubowicz:
Thanks for opening this issue!
We briefly spoke about this at the rendering office hours last week ([#office hours > Renderer 2025-07-09 @ 💬](https://xi.zulipchat.com/#narrow/channel/359642-office-hours/topic/Renderer.202025-07-09/near/527907647)).

At a high level, this is definitely something that would be a very welcome addition 😄 !
As an additional reference, portable SIMD has a carefully designed [swizzling API](https://doc.rust-lang.org/std/simd/trait.Swizzle.html). The open question is what API enables LLVM to recognize the shuffle/swizzle so it can potentially optimize it.

Would you be interested in drafting a proof-of-concept?

valadaptive:
I can look into drafting one in the next few days.

I'm not sure how much special stuff needs to be done for LLVM to optimize shuffles. In my testing, it can optimize `_mm_permutevar_ps` into a constant shuffle. Not sure about `_mm_shuffle_epi8` on non-AVX2 targets, or WASM `i8x16_swizzle` -> `i8x16_shuffle`. As far as I know, NEON doesn't even have a constant swizzle instruction.

Shnatsel:
Cheap dynamic shuffles within blocks are implemented in #266 (merged)

Arbitrary dynamic shuffles are implemented in #276 (waiting on review)

Constant-index shuffles are indeed tricky. They can be implemented on stable with a proc macro, but it's questionable whether that's worth the complexity and the compile-time hit.

Shnatsel:
#354 adds `concat_swizzle_dyn()`, and I think we can call this done once that's merged.

[...]

PR #164, opened 2025-12-13
Add a "slide" operation (like x86's `alignr` and ARM's `vext`)

valadaptive:
Depends on #159. Working on this is what sent me down that rabbit hole in the first place.

Progress towards #29, implementing the first lane-shuffling operations.

This PR adds an operation that concatenates two vectors and then takes a window of the concatenation. In other words, it takes two `n`-element vectors and a "window shift" of `s`, and returns the *last* `n - s` elements of the first vector concatenated with the *first* `s` elements of the second. This is like the `vext` family on ARM or `alignr` on x86.

This can be used to implement "shift items" or "rotate items" operations, by providing a zero vector for one operand to get "shift" behavior and providing the same operand twice to get "rotate" behavior.

There are two variants of this operation: one that operates over the full width of the vector and one that operates within 128-bit blocks. Even on AVX2, `_mm256_alignr_epi8` operates within 128-bit lanes, and it takes some extra permutes to make a full-width version, so I think it makes sense to provide a per-block version. This will also be the case when I implement fully-general swizzles.

The shift amount is provided as a const generic argument, since the underlying intrinsics also expose it that way. In many cases, we need to do math on that const generic argument before passing it to the intrinsic--we might need to convert it from a `usize` to `i32`, divide it by the number of bytes per scalar element, wrap it modulo 16, etc. Rust doesn't let us do this yet, so I've added "faux-dynamic" versions of the `alignr`/`vext` intrinsics that are implemented as a huge match statement, one for each of the 16 byte shift amounts. Since we inline everything, these should be evaluated at compile time.

I haven't yet confirmed that this generates the LLVM IR that we expect on all targets. The codegen *seems* to be producing `shufflevector`s on x86 and AArch64 (I haven't looked at WebAssembly), but all the functions go through some level of indirection via `vectorize` or `call_once` or something so it's hard to match them up to what they're supposed to be doing.

I'm not fully tied to the name "slide". It's hard to find a good name for this operation. x86's "alignr" makes me think of memory alignment, and ARM's "vext" ("vector extend") sounds like you're just combining two vectors into a wider one.

### DIFF
--- a/fearless_simd_tests/tests/harness/mod.rs
+++ b/fearless_simd_tests/tests/harness/mod.rs
@@ -3212,10 +3212,19 @@ fn store_slice_f32x4<S: Simd>(simd: S) {
 fn slide_f32x4<S: Simd>(simd: S) {
     let a = f32x4::from_slice(simd, &[1.0, 2.0, 3.0, 4.0]);
     let b = f32x4::from_slice(simd, &[5.0, 6.0, 7.0, 8.0]);
-    assert_eq!(*a.slide::<0>(b), [1.0, 2.0, 3.0, 4.0]); // returns a
+    assert_eq!(*a.slide::<0>(b), [1.0, 2.0, 3.0, 4.0]);
     assert_eq!(*a.slide::<1>(b), [2.0, 3.0, 4.0, 5.0]);
     assert_eq!(*a.slide::<2>(b), [3.0, 4.0, 5.0, 6.0]);
-    assert_eq!(*a.slide::<4>(b), [5.0, 6.0, 7.0, 8.0]); // returns b
+    assert_eq!(*a.slide::<4>(b), [5.0, 6.0, 7.0, 8.0]);
+}
+
+#[simd_test]
+fn slide_within_blocks_f32x4<S: Simd>(simd: S) {
+    let a = f32x4::from_slice(simd, &[1.0, 2.0, 3.0, 4.0]);
+    let b = f32x4::from_slice(simd, &[5.0, 6.0, 7.0, 8.0]);
+    assert_eq!(*a.slide_within_blocks::<0>(b), *a);
+    assert_eq!(*a.slide_within_blocks::<1>(b), *a.slide::<1>(b));
+    assert_eq!(*a.slide_within_blocks::<4>(b), *b);
 }
 
 #[simd_test]
@@ -3239,6 +3248,18 @@ fn slide_f32x8<S: Simd>(simd: S) {
 }
 
 #[simd_test]
+fn slide_within_blocks_f32x8<S: Simd>(simd: S) {
+    let a = f32x8::from_slice(simd, &[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0]);
+    let b = f32x8::from_slice(simd, &[9.0, 10.0, 11.0, 12.0, 13.0, 14.0, 15.0, 16.0]);
+    assert_eq!(*a.slide_within_blocks::<0>(b), *a);
+    assert_eq!(
+        *a.slide_within_blocks::<1>(b),
+        [2.0, 3.0, 4.0, 9.0, 6.0, 7.0, 8.0, 13.0]
+    );
+    assert_eq!(*a.slide_within_blocks::<4>(b), *b);
+}
+
+#[simd_test]
 fn slide_f32x16<S: Simd>(simd: S) {
     let a: Vec<f32> = (1_i16..=16).map(f32::from).collect();
     let b: Vec<f32> = (17_i16..=32).map(f32::from).collect();
@@ -3257,12 +3278,38 @@ fn slide_f32x16<S: Simd>(simd: S) {
 }
 
 #[simd_test]
+fn slide_within_blocks_f32x16<S: Simd>(simd: S) {
+    let a: Vec<f32> = (1_i16..=16).map(f32::from).collect();
+    let b: Vec<f32> = (17_i16..=32).map(f32::from).collect();
+    let a = f32x16::from_slice(simd, &a);
+    let b = f32x16::from_slice(simd, &b);
+
+    assert_eq!(*a.slide_within_blocks::<0>(b), *a);
+
+    let expected_1: [f32; 16] = [
+        2.0, 3.0, 4.0, 17.0, 6.0, 7.0, 8.0, 21.0, 10.0, 11.0, 12.0, 25.0, 14.0, 15.0, 16.0, 29.0,
+    ];
+    assert_eq!(*a.slide_within_blocks::<1>(b), expected_1);
+
+    assert_eq!(*a.slide_within_blocks::<4>(b), *b);
+}
+
+#[simd_test]
 fn slide_f64x2<S: Simd>(simd: S) {
     let a = f64x2::from_slice(simd, &[1.0, 2.0]);
     let b = f64x2::from_slice(simd, &[3.0, 4.0]);
     assert_eq!(*a.slide::<0>(b), [1.0, 2.0]);
     assert_eq!(*a.slide::<1>(b), [2.0, 3.0]);
     assert_eq!(*a.slide::<2>(b), [3.0, 4.0]);
+}
+
+#[simd_test]
+fn slide_within_blocks_f64x2<S: Simd>(simd: S) {
+    let a = f64x2::from_slice(simd, &[1.0, 2.0]);
+    let b = f64x2::from_slice(simd, &[3.0, 4.0]);
+    assert_eq!(*a.slide_within_blocks::<0>(b), *a);
+    assert_eq!(*a.slide_within_blocks::<1>(b), *a.slide::<1>(b));
+    assert_eq!(*a.slide_within_blocks::<2>(b), *b);
 }
 
 #[simd_test]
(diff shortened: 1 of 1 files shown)


fearless_simd_tests/tests/harness/mod.rs
```diff
@@ -2984,3 +2984,495 @@ fn index_consistency<S: Simd>(simd: S) {
...
+    assert_eq!(*a.slide::<1>(b), [2, 3]);
+    assert_eq!(*a.slide::<2>(b), [3, 4]);
+}
+
+#[simd_test]
+fn slide_mask64x4<S: Simd>(simd: S) {
+    let a = mask64x4::from_slice(simd, &[1, 2, 3, 4]);
+    let b = mask64x4::from_slice(simd, &[5, 6, 7, 8]);
+    assert_eq!(*a.slide::<0>(b), [1, 2, 3, 4]);
+    assert_eq!(*a.slide::<2>(b), [3, 4, 5, 6]); // crosses block
```
LaurenzV: Don't all slide tests cross the block? Just wondering why we add a comment here but not to the others. 🤔

fearless_simd_tests/tests/harness/slide_exhaustive.rs
LaurenzV, line 1: Do we really need all these tests? I feel like the ones in mod.rs should be enough?

fearless_simd_gen/src/ops.rs
```diff
@@ -297,7 +310,12 @@ impl Op {
             .collect::<Vec<_>>();
         let method_ident = Ident::new(self.method, Span::call_site());
         let sig_inner = match &self.sig {
-            OpSig::Splat | OpSig::LoadInterleaved { .. } | OpSig::StoreInterleaved { .. } => {
+            OpSig::Splat => {
```
LaurenzV, line 323: Why was `Splat` changed here as well?

fearless_simd_gen/src/ops.rs
```diff
@@ -910,6 +955,14 @@ const MASK_OPS: &[Op] = &[
     ),
 ];
 
+pub(crate) fn base_trait_ops() -> Vec<Op> {
```
LaurenzV, line 974: Small thing, but as far as I can see this method is only used in one place, and returning an iterator here should be sufficient?

fearless_simd_gen/src/ops.rs
```diff
@@ -1319,8 +1386,8 @@ impl OpSig {
     }
     fn vec_trait_arg_names(&self) -> &'static [&'static str] {
         match self {
-            Self::Splat
-            | Self::LoadInterleaved { .. }
+            Self::Splat => &["simd", "val"],
```
LaurenzV, line 1406: Again, on purpose that splat was changed here?

fearless_simd_tests/tests/harness/mod.rs
LaurenzV, line 1: As far as I can tell, there are no tests for `slide_within_blocks` (except in the exhaustive ones), would probably be good to have as well.

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -478,3 +548,25 @@ impl Level for Neon {
...
+        let shift_i32 = i32::try_from(shift).unwrap();
+        quote! { #shift => vextq_u8::<#shift_i32>(a, b) }
+    });
+
+    quote! {
+        /// This is a version of the `vext` intrinsic that takes a non-const shift argument. The shift is still
+        /// expected to be constant in practice, so the match statement will be optimized out. This exists because
+        /// Rust doesn't currently let you do math on const generics.
+        #[inline(always)]
+        unsafe fn dyn_vext_128(a: uint8x16_t, b: uint8x16_t, shift: usize) -> uint8x16_t {
```
LaurenzV, line 567: If we already have an unsafe block inside we don't need it on the function itself, no?

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -395,6 +399,72 @@ impl Level for Neon {
...
+                let block_wrapper = vec_ty.aligned_wrapper();
+                let bytes_ty = vec_ty.reinterpret(ScalarType::Unsigned, 8);
+                let combined_bytes = bytes_ty.rust();
+                let scalar_bytes = vec_ty.scalar_bits / 8;
+                let num_items = vec_ty.len;
+                let to_bytes = generic_op_name("cvt_to_bytes", vec_ty);
+                let from_bytes = generic_op_name("cvt_from_bytes", vec_ty);
+
+                let byte_shift = if scalar_bytes == 1 {
+                    quote! { SHIFT }
```
LaurenzV, line 415: In this case we probably wouldn't need to call the dyn methods either but can just call the intrinsic, right?

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -1489,4 +1562,131 @@ impl X86 {
...
+
+    fn sse42_slide_helpers() -> TokenStream {
+        let mut fns = vec![];
+
+        for num_blocks in [2_usize, 4_usize] {
+            let helper_name = format_ident!("cross_block_alignr_128x{}", num_blocks);
+            let blocks_idx = 0..num_blocks;
+
+            // Unroll the construction of the blocks. I tried using `array::from_fn`, but the compiler thought the
+            // closure was too big and didn't inline it.
```
LaurenzV, line 1623: Even if you annotate the closure with` #[inline(always)]`?

fearless_simd/src/generated/avx2.rs
```diff
@@ -7187,3 +8351,105 @@ impl<S: Simd> From<mask64x4<S>> for __m256i {
...
+        [
+            cross_block_alignr_one(&regs, 0, shift_bytes),
+            cross_block_alignr_one(&regs, 2, shift_bytes),
+        ]
+    }
+}
+#[doc = r" Concatenates `b` and `a` (each 1 x __m256i = 2 blocks) and extracts 2 blocks starting at byte offset"]
+#[doc = r" `shift_bytes`. Extracts from [b : a] (b in low bytes, a in high bytes), matching alignr semantics."]
+#[inline(always)]
+unsafe fn cross_block_alignr_256x1(a: __m256i, b: __m256i, shift_bytes: usize) -> __m256i {
```
LaurenzV, line 8632: Same as in a different location, we probably don't need to mark the function as unsafe if we use an unsafe block inside?

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -7529,3 +8597,76 @@ impl<S: Simd> From<mask64x2<S>> for __m128i {
...
+        {
+            let [lo, hi] = crate::support::cross_block_slide_blocks_at(&b, &a, 1usize, shift_bytes);
+            unsafe { dyn_alignr_128(hi, lo, shift_bytes % 16) }
+        },
+    ]
+}
+#[doc = r" Concatenates `b` and `a` (each N blocks) and extracts N blocks starting at byte offset `shift_bytes`."]
+#[doc = r" Extracts from [b : a] (b in low bytes, a in high bytes), matching `alignr` semantics."]
+#[inline(always)]
+unsafe fn cross_block_alignr_128x4(
```
LaurenzV, line 8889: I'm fine leaving it this way for now, but I'm wondering whether this is really going to be faster than just using the fallback approach? Have you done any benchmarks on that? (Also for 128x2 and 256x2 in AVX2) Especially because cross_block_slide_blocks_at does quite a bit of work and is called 4 times.

LaurenzV: (Also applies to NEON and WASM, basiaclly anywhere where we polyfill a larger vector width than the base one.)

LaurenzV (commented):
I haven't tried to fully understand all of the logic of the added (helper) methods, but I've gotten a good overview and overall it seems fine to add. However, I do have some comments/remarks.

fearless_simd_tests/tests/harness/mod.rs
valadaptive: Yeah, that comment is redundant. I'll remove it.

fearless_simd_tests/tests/harness/slide_exhaustive.rs
valadaptive: These tests aren't compiled in by default, but given how intricate the block-selection logic is, I think it's good to have *some* way to exhaustively test all shift amounts.

fearless_simd_gen/src/ops.rs
valadaptive: I noticed that the signatures for `splat` and `slide` could be autogenerated, which avoids duplicating the doc comments in two different places. For `splat` this isn't a big deal, since the doc comment is quite simple, but `slide`'s doc comment is a lot more involved.

Looking back on it, I probably should just revert it for `splat`.

fearless_simd_gen/src/ops.rs
valadaptive: Returning a `Vec<Op>` matches what `vec_trait_ops_for` and `overloaded_ops_for` do.

fearless_simd_gen/src/mk_neon.rs
valadaptive: It's marked as unsafe to match the unsafety of the underlying `vextq_[...]` function that it calls. Rust [now wants you to use `unsafe` blocks even inside unsafe functions](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html).

fearless_simd_gen/src/mk_neon.rs
valadaptive: Our shift amount is a `usize`, whereas the `vext` intrinsic family takes it as an `i32`. Converting between the two is impossible on stable Rust.

fearless_simd_gen/src/mk_x86.rs
valadaptive: Marking it as `#[inline(always)]` will inline it into `array::from_fn`, but `array::from_fn` is *not* marked as `#[inline(always)]`, so the compiler chooses not to inline it here.

fearless_simd/src/generated/sse4_2.rs
valadaptive: `cross_block_slide_blocks_at` is intended to be fully evaluated at compile time. All it does is determine which blocks to combine given a shift value, and since the shift value is constant, the selected blocks should be too. If Rust had better const-eval support, we could make everything a `const fn`, but sadly we're not there yet.

fearless_simd_gen/src/ops.rs
valadaptive: I rediscovered the reason I ended up having to generate it for `splat` as well.

I added code to autogenerate method signatures for every op in `BASE_OPS` that has `OpKind::BaseTraitMethod`. That applies to the new `slide` methods, but also to the existing `splat` methods.

In order to avoid touching the existing `splat` methods, I would need to special-case them. I think it's cleaner to just autogenerate them both.

PR #167, opened 2025-12-13
Separate precise/non-precise float-to-int conversions

valadaptive:
The extra work necessary to emulate saturating behavior on x86 is in fact slowing things down. This PR separates the float-to-int conversion methods into non-precise and precise variants. The non-precise ones can return implementation-defined results if any values are NaN or out of range for the destination type. This is useful if you're already doing your own clamping.

On x86 and WebAssembly with `relaxed_simd`, we can use faster operations for the non-precise versions. On AArch64 and the fallback implementation, the precise versions just call the non-precise ones since they behave identically.

Since we never released a version of fearless_simd with the saturating conversions implemented everywhere, I think this isn't a breaking change.

I don't think we need to define non-precise/precise operations in the int-to-float direction as well. Out-of-range values are simply rounded to the nearest representable value, and I think that's consistent across all architectures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,12 +29,13 @@ This release has an [MSRV][] of 1.88.
 - The `any_true`, `all_true`, `any_false`, and `all_false` methods on mask types. ([#141][] by [@valadaptive][])
 - Documentation for most traits, vector types, and operations. ([#154][] by [@valadaptive][])
 - A "shift left by vector" operation, to go with the existing "shift right by vector". ([#155][] by [@valadaptive][])
+- "Precise" float-to-integer conversions, which saturate out-of-bounds results and convert NaN to 0 across all platforms. ([#167][] by [@valadaptive][])
 
 ### Fixed
 
 - Integer equality comparisons now function properly on x86. Previously, they performed "greater than" comparisons.
   ([#115][] by [@valadaptive][])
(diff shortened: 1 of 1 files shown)


valadaptive:
I've updated the `SimdCvt[...]` trait implementations to include documentation for the specific types being converted to/from.

fearless_simd_gen/src/mk_simd_trait.rs
```diff
@@ -169,6 +169,9 @@ fn mk_simd_float() -> TokenStream {
             #[inline(always)]
             fn to_int<T: SimdCvtTruncate<Self>>(self) -> T { T::truncate_from(self) }
 
+            #[inline(always)]
+            fn to_int_precise<T: SimdCvtTruncate<Self>>(self) -> T { T::truncate_from_precise(self) }
```
Ralith, line 187: Docs here too, maybe?

fearless_simd_tests/tests/harness/mod.rs
Ralith, line 1: Some test coverage for the relaxed methods would be nice.

valadaptive: Not sure what I should be testing for. The `cvt_u32_f32x4` and `cvt_i32_f32x4` tests still exist; I've just changed the saturating/NaN/infinity tests to use the precise operations, since those are the only ones that guarantee consistent semantics.

Ralith: Ah, missed that there were separate cases already :+1:

PR #171, opened 2025-12-15
Add (scalarized) safe gather and scatter ops

valadaptive:
Depends on #170. The PR stack is getting a bit large.

When looking into gradient rendering in Vello, I noticed that we often perform many reads from the gradient color LUT. Each of those reads requires a bounds check.

The indices into this LUT come from vector types, so theoretically, we should be able to elide all the bounds checks like this:

```rust
let mut indices: u32x16<S> = [...];
assert!(!self.lut.is_empty());
assert!(self.lut.len() <= u32::MAX);
indices = indices.min((self.lut.len() - 1) as u32);
```

Unfortunately, the compiler still does not recognize that the indices are guaranteed to be in-bounds. To elide the bounds checks, we need to perform the `min` operation on every index *individually*, after converting it to a `usize`.

We can avoid this by introducing a "gather" operation here which ensures all the indices are valid (panicking if the source slice is empty and clamping the indices), then doing a bunch of unchecked accesses in a loop. There's an equivalent "scatter" operation which does the same thing but for writes.

These operations work on arbitrary slice types and `gather` returns an array. The only vector type involved is the one that holds all the indices.

This PR intentionally does not introduce any operations that gather into/scatter from *vector* types. That also means that no hardware gather/scatter instructions are used right now, and the performance benefit comes solely from avoiding bounds checks. I'm not sure if we should be reserving the names `gather` and `scatter` for operations that work on vectors.

### DIFF
--- a/fearless_simd_dev_macros/src/lib.rs
+++ b/fearless_simd_dev_macros/src/lib.rs
@@ -29,6 +29,20 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
         .filter(|attr| !attr.path().is_ident("simd_test"))
         .collect();
 
+    // If this is a `#[should_panic]` test, run it with the fallback so it actually panics
+    let should_panic_attr = input_fn
+        .attrs
+        .iter()
+        .find(|attr| attr.path().is_ident("should_panic"));
+    let panic_else = if should_panic_attr.is_some() {
+        quote! {
+            let fallback = fearless_simd::Fallback::new();
+            #input_fn_name(fallback);
+        }
+    } else {
+        quote! {}
+    };
+
     let fallback_snippet = quote! {
         #(#test_attrs)*
         #[test]
@@ -47,6 +61,9 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
     // target features aren't supported. This is not ideal, since it may mislead you into thinking tests have passed
     // when they haven't even been run, but some CI runners don't support all target features and we don't want failures
     // as a result of that.
+    //
+    // However, for #[should_panic] tests, we need to panic if features aren't available to avoid
+    // "test did not panic as expected" failures.
 
     let neon_snippet = quote! {
         #[cfg(target_arch = "aarch64")]
@@ -56,6 +73,8 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
             if std::arch::is_aarch64_feature_detected!("neon") {
                 let neon = unsafe { fearless_simd::aarch64::Neon::new_unchecked() };
                 #input_fn_name(neon);
+            } else {
+                #panic_else
             }
         }
     };
@@ -68,6 +87,8 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
             if std::arch::is_x86_feature_detected!("sse4.2") {
                 let sse4 = unsafe { fearless_simd::x86::Sse4_2::new_unchecked() };
                 #input_fn_name(sse4);
+            } else {
+                #panic_else
             }
         }
     };
@@ -82,6 +103,8 @@ pub fn simd_test(_: TokenStream, item: TokenStream) -> TokenStream {
             {
                 let avx2 = unsafe { fearless_simd::x86::Avx2::new_unchecked() };
                 #input_fn_name(avx2);
+            } else {
+                #panic_else
             }
         }
     };


fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -141,6 +143,94 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+
+                    #[inline(always)]
+                    fn gather<T: Copy>(self, src: &[T]) -> Self::Gathered<T> {
+                        assert!(!src.is_empty(), "gather: source slice must not be empty");
+
+                        // Before ensuring the source slice is bigger than `Self::Element::MAX as usize`, we need to
+                        // make sure that's actually a valid cast. We may eventually get an i64/u64 type, which is
+                        // larger than `usize` on 32-bit platforms. If our `Element` type is wider than `usize`, then
+                        // `Element::MAX` will be larger than any possible slice length.
+                        let inbounds = if core::mem::size_of::<Self::Element>() <= core::mem::size_of::<usize>() &&
```
LaurenzV: Isn't < enough here? Let's say we are on a 32-bit system and we are using a u32 vector, then I don't think the slice length can ever be larger than u32::MAX?

LaurenzV: Actually, after some googling it seems like the length of a slice is even restricted to `isize::MAX`, but in any case I think it still can't happen if the two types have the same memory size?

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -141,6 +143,94 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+                        })
+                    }
+
+                    #[inline(always)]
+                    fn gather_into<T: Copy>(self, src: &[T], dst: &mut [T]) {
+                        assert_eq!(Self::N, dst.len(), "gather_into: destination slice must have the same element count as the vector type");
+                        assert!(!src.is_empty(), "gather_into: source slice must not be empty");
+
+                        // Same logic as for `gather`. See the comments there.
+                        let inbounds = if core::mem::size_of::<Self::Element>() <= core::mem::size_of::<usize>() &&
```
LaurenzV: Same as above.

fearless_simd_gen/src/mk_simd_types.rs
LaurenzV: I guess that might not strictly hold for signed integers (even though we don't support that yet), but we don't support those anyway.

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -141,6 +143,94 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+
+        // Scatter/gather operations
+        if ty.scalar == ScalarType::Unsigned {
+            let min_method = generic_op_name("min", ty);
+            conditional_impls.push(quote! {
+                impl<S: Simd> SimdGather<S> for #name<S> {
+                    type Gathered<T> = [T; #len];
+
+                    #[inline(always)]
+                    fn gather<T: Copy>(self, src: &[T]) -> Self::Gathered<T> {
```
LaurenzV, line 147: I'm wondering whether it makes sense to add a debug assertion that no index is exceeds the slice length? This way it's still possible to detect bugs instead of silently clamping to the largest index. But then we would have to remove the guarantee from the documentation that it will always be clamped.

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -141,6 +143,94 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+                            }
+                        }
+                    }
+                }
+
+                impl<S: Simd> SimdScatter<S> for #name<S> {
+                    #[inline(always)]
+                    fn scatter<T: Copy>(self, src: &[T], dst: &mut [T]) {
+                        assert_eq!(Self::N, src.len(), "scatter: source slice must have the same element count as the vector type");
+                        assert!(!dst.is_empty(), "scatter: source slice must not be empty");
```
LaurenzV:
```suggestion
                        assert!(!dst.is_empty(), "scatter: dest slice must not be empty");
```

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -141,6 +143,94 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+                        {
+                            self
+                        } else {
+                            self.simd.#min_method(self, ((dst.len() - 1) as Self::Element).simd_into(self.simd))
+                        };
+
+                        for i in 0..Self::N {
+                            // Safety: All elements of `inbounds` are in [0, dst.len()). 0 is a valid index, because we
+                            // asserted that `dst` is not empty.
+                            unsafe { *dst.get_unchecked_mut(inbounds[i] as usize) = src[i] };
```
LaurenzV: Can we also use src.get_unchecked here?

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -141,6 +143,94 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+                        let inbounds = if core::mem::size_of::<Self::Element>() <= core::mem::size_of::<usize>() &&
+                            src.len() > Self::Element::MAX as usize
+                        {
+                            self
+                        } else {
+                            self.simd.#min_method(self, ((src.len() - 1) as Self::Element).simd_into(self.simd))
+                        };
+
+                        for i in 0..Self::N {
+                            dst[i] = unsafe {
```
LaurenzV: Can we not use dst.get_unchecked_mut here?

fearless_simd_gen/src/mk_simd_types.rs
valadaptive: Not sure what you mean here. The purpose of the `size_of` check is to avoid `Self::Element::MAX as usize` overflowing.

fearless_simd_gen/src/mk_simd_types.rs
valadaptive: I've updated the code to use unchecked indexing throughout, since we've already asserted all the preconditions.

fearless_simd_gen/src/mk_simd_types.rs
LaurenzV: I mean instead of using <= (less than or equal) in the condition, would it be enough to just use < (less than)?

valadaptive: OK, I think I understand this now after a bit of thought. If the sizes are identical, then you're correct that `src.len() > Self::Element::MAX as usize` can never be true. I've made that change.

valadaptive:
Before merging, we should figure out what's going on with the Vello use case this was intended for. The theory is that unchecked indexing improves performance (I noticed a ton of bounds checks in the generated ARM assembly). Indeed, using unchecked indexing *directly* improves performance:

> btw, I tried the inlining you suggested on main again. If I do that, as mentioned performance drops (from around 470ms to 520ms). however, if I replace the lut check with an unchecked access in the inlined version, it goes down to 460ms! If I do the same on main(just changing to unchecked), I get the same performance, so I think for some reason the compiler is able to optimize some of the index checks away in current main, but it stops working once you inline the iterator... Even though it doesnt make much sense

But the "gather" version is 525ms, slower than not using it.

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -133,6 +135,103 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+                        } else {
+                            // No `max(0)`; we do not implement `SimdGather` for signed integers.
+                            //
+                            // Converting `src.len() - 1` to `Self::Element` will not wrap, because if `src.len() - 1 >=
+                            // Self::Element::MAX`, that means that `src.len() > Self::Element::MAX`, and we take the
+                            // above branch instead.
+                            self.simd.#min_method(self, ((src.len() - 1) as Self::Element).simd_into(self.simd))
+                        };
+
+                        let inbounds = &*inbounds;
```
LaurenzV: I don't think this one's necessary? Same for the other two functions.

LaurenzV: (I'm talking about the `let inbounds = &*inbounds;`.)

LaurenzV (approved):

fearless_simd_gen/src/mk_simd_types.rs
```diff
@@ -133,6 +135,100 @@ pub(crate) fn mk_simd_types() -> TokenStream {
...
+                            // The slice is big enough to accept any index. For instance, if this is a vector of `u8`s,
+                            // `Self::Element::MAX` is 255, so the slice must be at least 256 elements long.
+                            self
+                        } else {
+                            // No `max(0)`; we do not implement `SimdGather` for signed integers.
+                            //
+                            // Converting `src.len() - 1` to `Self::Element` will not wrap, because if `src.len() - 1 >=
+                            // Self::Element::MAX`, that means that `src.len() > Self::Element::MAX`, and we take the
+                            // above branch instead.
+                            self.simd.#min_method(self, ((src.len() - 1) as Self::Element).simd_into(self.simd))
```
Shnatsel, line 165: Running `min()` on every access seems slow. Can we instead find the maximum index in the vector once, cache it, and only do a single bounds check on every load for the whole vector by comparing `SimdGather::max_index()` against `src.len()`? This means `SimdGather` needs to be a struct instead of a trait but I don't think that's a problem.

I've described such a design in more detail here: https://github.com/okaneco/safe_unaligned_simd/issues/37

valadaptive: It's not quite that bad; we're doing it once per gather instead of once per access.

As I understand it, once you've constructed a `GatherIndices`, you can use it for multiple gather operations. Do you have a use case in mind for this? In e.g. Vello, I'd expect the *indices* to vary while the *lookup table* remains constant. You're describing an optimization that would improve performance when the *lookup table* varies but the *indices* remain constant.

Shnatsel: I think your assessment is correct. I wanted to use it to load pixels from in-memory representation of RGBRGBRGBRGB into vectors RRRR, GGGG, BBBB.

This was when working on https://github.com/vstroebel/jpeg-encoder/pull/17 or https://github.com/vstroebel/jpeg-encoder/pull/18, I can't remember which one. But that crate doesn't use fearless_simd and doesn't have plans to, so this is pretty far down on my priority list.

Shnatsel: Also as discussed in https://github.com/okaneco/safe_unaligned_simd/issues/37 scatter/gather instructions are a performance minefield. LLVM will sometimes transform scalar loads into gather and tank performance that way.

PR #172, opened 2025-12-16
Prepare v0.4.0 release

LaurenzV:
I think now is a good time to get this out. I double-checked the changelog and only found one other PR worth mentioning.


LaurenzV:
I wouldn’t mind cutting a v0.5 right after those two PRs are merged, Injust think it would be good to have a “checkpoint” up until now because as we’ve seen #159 is a bigger change that seems to have performance impacts in certain cases, so it would be good to have this release to fall back on in case we notice other problems (I’m sure it will be fine! but I don’t think it hurts either.)

PR #176, opened 2025-12-18
More codegen cleanups

valadaptive:
The first half of a codegen rework. This part includes the smaller changes: using helper methods in more places (`generic_op_name`, `VecType::reinterpret`, the new `VecType::cast`), moving some methods across types so we can pass in fewer arguments, removing old TODO comments, etc.

There are no substantial functional changes, but for consistency's sake and to make it easier to reuse code later, I added a core_arch token for WASM SIMD128 and used it in the `WasmSimd128` level, which matches all the other levels.


fearless_simd/src/core_arch/wasm32/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+
+// There is intentionally no method delegation here because all the WASM SIMD128 methods are enabled or disabled
+// statically--there is no feature detection.
+impl WasmSimd128 {
+    /// Create a SIMD token.
+    #[inline]
+    pub const fn new() -> Self {
+        Self { _private: () }
+    }
+}
```
LaurenzV, line 18: But is it not possible now to create a WASM SIMD token even though the simd128 is not enabled?

LaurenzV (approved):
Seems like a bunch of improvements!

fearless_simd/src/core_arch/wasm32/mod.rs
valadaptive: Whoops, yes it was. I've gated the `wasm32` core_arch module behind the `simd128` target feature to prevent that.

PR #177, opened 2025-12-18
Add `Level` trait in codegen to consolidate the generation logic

valadaptive:
This PR is stacked on top of #176, and adds a `Level` trait to fearless_simd_gen. The different code generation backends (Neon, AVX2, etc) all implement this trait.

The various level-specific code generation functions are now mostly trait methods. The `Level` trait implements some default trait methods, which call into the level-specific ones as necessary. Some of the code that was previously level-specific is actually identical between all implementations, and was moved onto the `Level` trait.

### DIFF
--- a/fearless_simd_gen/src/mk_sse4_2.rs
+++ b/fearless_simd_gen/src/mk_sse4_2.rs
@@ -32,6 +32,10 @@ impl Level for Sse4_2 {
         128
     }
 
+    fn enabled_target_features(&self) -> Option<&'static str> {
+        Some("sse4.2")
+    }
+
     fn arch_ty(&self, vec_ty: &VecType) -> TokenStream {
         arch_ty(vec_ty).into_token_stream()
     }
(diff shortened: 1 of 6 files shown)


fearless_simd_gen/src/mk_sse4_2.rs
```diff
@@ -8,52 +8,76 @@ use crate::arch::x86::{
...
     }
-}
 
-pub(crate) fn mk_sse4_2_impl() -> TokenStream {
-    let imports = type_imports();
-    let arch_types_impl = impl_arch_types(Level.name(), 128, arch_ty);
-    let simd_impl = mk_simd_impl();
-    let ty_impl = mk_type_impl();
+    fn max_block_size(&self) -> usize {
+        128
```
LaurenzV, line 32: Don't we support 256 and 512 too? 🤔 But I probably just forgot what this is actually used for...

fearless_simd/src/generated/wasm.rs
```diff
@@ -82,11 +82,7 @@ impl Simd for WasmSimd128 {
     }
     #[inline]
     fn vectorize<F: FnOnce() -> R, R>(self, f: F) -> R {
-        #[inline]
-        unsafe fn vectorize_simd128<F: FnOnce() -> R, R>(f: F) -> R {
-            f()
-        }
-        unsafe { vectorize_simd128(f) }
+        f()
```
LaurenzV, line 85: This is possible because it's enabled statically, right?

valadaptive: Yes; I left a comment on this in `mk_wasm`.

fearless_simd_gen/src/mk_sse4_2.rs
valadaptive: I realize now that I hadn't actually added any documentation for what these functions represent.

I've added some documentation to all the `Level` functions that the various implementations are expected to implement or potentially override, and also replaced `make_vectorize_body` with a simpler function that only returns the target features that need to be enabled.

PR #178, opened 2025-12-18
Consolidate AVX2 and SSE4.2 generation into one module

valadaptive:
This PR is stacked on top of #177.

After adding the `Level` trait as an abstraction, we can use it to merge the AVX2 and SSE4.2 implementations into a single module. Most of the logic is the same, except for a few functions which need to do something different if AVX2 is supported. This gets rid of a lot of duplicate code.

### DIFF
--- a/fearless_simd_gen/src/mk_x86.rs
+++ b/fearless_simd_gen/src/mk_x86.rs
@@ -230,7 +230,7 @@ impl X86 {
                     let gt = simple_sign_unaware_intrinsic("cmpgt", vec_ty);
 
                     if vec_ty.scalar == ScalarType::Unsigned {
-                        // SSE4.2 only has signed GT/LT, but not unsigned.
+                        // Below AVX-512, we only have signed GT/LT, not unsigned.
                         let set = set1_intrinsic(vec_ty);
                         let sign = match vec_ty.scalar_bits {
                             8 => quote! { 0x80u8 },
@@ -451,8 +451,7 @@ impl X86 {
                         quote! {
                             let (a, b) = self.#split(a);
                             unsafe {
-                                // Note that SSE4.2 only has an intrinsic for saturating cast,
-                                // but not wrapping.
+                                // Below AVX-512. we only have an intrinsic for saturating cast, but not wrapping.
                                 let mask = #mask(0xFF);
                                 let lo_masked = _mm_and_si128(a.into(), mask);
                                 let hi_masked = _mm_and_si128(b.into(), mask);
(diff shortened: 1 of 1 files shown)


fearless_simd_gen/src/mk_x86.rs
```diff
@@ -0,0 +1,1494 @@
...
+    fn make_level_body(&self) -> TokenStream {
+        let level_tok = self.token();
+        match self {
+            Self::Sse4_2 => quote! {
+                #[cfg(not(all(target_feature = "avx2", target_feature = "fma")))]
+                return Level::#level_tok(self);
+                #[cfg(all(target_feature = "avx2", target_feature = "fma"))]
+                {
+                    Level::baseline()
+                }
```
LaurenzV, line 94: Does this mean if we have AVX2 available, it will use that even if we "force" SSE4?

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -0,0 +1,1494 @@
...
+                    let eq_intrinsic = simple_sign_unaware_intrinsic("cmpeq", vec_ty);
+
+                    let max_min_expr = x86::expr(max_min, vec_ty, &args);
+                    quote! { #eq_intrinsic(#max_min_expr, a.into()) }
+                }
+                "simd_lt" | "simd_gt" => {
+                    let gt = simple_sign_unaware_intrinsic("cmpgt", vec_ty);
+
+                    if vec_ty.scalar == ScalarType::Unsigned {
+                        // SSE4.2 only has signed GT/LT, but not unsigned.
```
LaurenzV: I guess this comment must now apply to AVX2 as well?

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -0,0 +1,1494 @@
...
+            ("shr", ScalarType::Int) => "sra",
+            ("shl", _) => "sll",
+            _ => unreachable!(),
+        };
+        let ty_bits = vec_ty.n_bits();
+        let suffix = op_suffix(vec_ty.scalar, vec_ty.scalar_bits.max(16), false);
+        let shift_intrinsic = intrinsic_ident(op, suffix, ty_bits);
+
+        if vec_ty.scalar_bits == 8 {
+            // SSE doesn't have shifting for 8-bit, so we first convert into
```
LaurenzV: Also for AVX?

fearless_simd_gen/src/mk_x86.rs
LaurenzV: (I realize you didn't make that change, but I just noticed it.)

valadaptive: This is correct. I believe it was done in https://github.com/linebender/fearless_simd/pull/105.

If AVX2 is *statically* enabled (e.g. we compile with `-Ctarget-cpu=x86-64-v3` or higher), SSE4.2 just forwards to it. I believe the purpose of this is to avoid compiling SSE4.2 code into the binary if we know we never need to run it.

fearless_simd_gen/src/mk_x86.rs
valadaptive: Good catch; I've gone through and updated these sorts of comments. Unsigned integer comparisons and (for the `narrow` ops) truncating u16->u8 casts are supported only in AVX-512 and above, and 8-bit shifts do not appear to exist on x86 at all.

PR #181, opened 2026-01-03
Add methods for storing vectors back to memory

LaurenzV:
As per title, by having this, we don't have to store them back using `copy_from_slice` (which may or may not optimize into vector store instructions) but can use methods that use the corresponding intrinsics under the hood.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This release has an [MSRV][] of 1.88.
 - A "shift left by vector" operation, to go with the existing "shift right by vector". ([#155][] by [@valadaptive][])
 - "Precise" float-to-integer conversions, which saturate out-of-bounds results and convert NaN to 0 across all platforms. ([#167][] by [@valadaptive][])
 - The `Level::is_fallback` method, which lets you check if the current SIMD level is the scalar fallback. This works even if `Level::Fallback` is not compiled in, always returning false in that case. ([#168][] by [@valadaptive][])
+- Added `store_array` methods to store SIMD vectors back to memory explicitly using intrinsics. ([#181][] by [@LaurenzV][])
 
 ### Fixed
 
@@ -127,6 +128,7 @@ No changelog was kept for this release.
 [@Ralith]: https://github.com/Ralith
 [@DJMcNab]: https://github.com/DJMcNab
 [@valadaptive]: https://github.com/valadaptive
+[@LaurenzV]: https://github.com/LaurenzV
 [@Shnatsel]: https://github.com/Shnatsel
 
 [#75]: https://github.com/linebender/fearless_simd/pull/75


LaurenzV:
(Will add changeling entry upon approval.)

fearless_simd_gen/src/mk_simd_trait.rs
```diff
@@ -167,6 +167,10 @@ fn mk_simd_base() -> TokenStream {
             ///
             /// The slice must be the proper width.
             fn from_slice(simd: S, slice: &[Self::Element]) -> Self;
+            /// Store a SIMD vector into a slice.
+            ///
+            /// The slice must be the proper width.
+            fn store_slice(&self, slice: &mut [Self::Element]);
```
LaurenzV, line 173: Maybe we should rename to `store_into_slice`?

fearless_simd_gen/src/ops.rs
```diff
@@ -106,6 +106,8 @@ pub(crate) enum OpSig {
     /// Takes a single argument of the vector type, or a reference to it, and returns the corresponding array type (e.g.
     /// `[f32; 4]` for `f32x4<S>`) or a reference to it.
     AsArray { kind: RefKind },
+    /// Takes a vector and a mutable reference to an array, and stores the vector elements into the array.
+    ToArray,
```
valadaptive: Maybe this should be renamed `StoreArray`?

fearless_simd_gen/src/generic.rs
```diff
@@ -332,6 +333,49 @@ pub(crate) fn generic_as_array<T: ToTokens>(
...
+
+    let store_unaligned = store_unaligned_block(&native_block_ty);
+    let store_expr = if block_count == 1 {
+        quote! {
+            unsafe { #store_unaligned(dest.as_mut_ptr() as *mut _, a.val.0) }
+        }
+    } else {
+        let blocks = (0..block_count).map(|n| {
+            let offset = n * num_scalars_per_block;
+            let block_idx = proc_macro2::Literal::usize_unsuffixed(n);
```
valadaptive, line 360: Is there a reason to use `usize_unsuffixed` here specifically? It definitely results in nicer generated code but we don't seem to use it in many other places.

fearless_simd_gen/src/mk_simd_trait.rs
valadaptive: I think `store_slice` is fine.

valadaptive (approved):
This looks good! Just a few minor code style nits.

fearless_simd_gen/src/generic.rs
LaurenzV: It does seem to be used in various places, no? 🤔 

image

valadaptive: I vaguely remembered not using it in some similar method, but maybe misremembered?

fearless_simd/src/generated/simd_trait.rs
```diff
@@ -2523,6 +2595,10 @@ pub trait SimdBase<S: Simd>:
     #[doc = r""]
     #[doc = r" The slice must be the proper width."]
     fn from_slice(simd: S, slice: &[Self::Element]) -> Self;
+    #[doc = r" Store a SIMD vector into a slice."]
+    #[doc = r""]
+    #[doc = r" The slice must be the proper width."]
```
Shnatsel, line 2600: > The slice must be the proper width.

This doesn't tell me if it should be exactly the size of the vector or >= the size of the vector.

I've opened https://github.com/linebender/fearless_simd/pull/182 to clarify.

PR #183, opened 2026-01-22
Less overzealous inlining

Shnatsel:
Using `vectorize()` as opposed to directly calling functions annotated `#[inline(always)]` is meant to introduce an inlining barrier and make the compiler make inlining decisions. Therefore putting `#[inline]` on it is rather pointless: if I wanted aggressive inlining, I wouldn't be calling `vectorize()` in the first place.

`#[inline]` also enables cross-crate inlining, but since `vectorize()` is a generic function it will be instantiated in the caller crate, so this is not a concern.

More info about inlining: https://matklad.github.io/2021/07/09/inline-in-rust.html

**Motivation**

~~I'm seeing a large performance regression from switching from `wide` to `fearless_simd` in https://github.com/QuState/PhastFT/pull/58 due to overly aggressive inlining. This alone doesn't fix it, but seems like a sensible first step.~~ It's not clear if over-zealous inlining was a cause of the regression, there are other factors. I'll re-measure once #185 lands.


fearless_simd/src/generated/avx2.rs
```diff
@@ -88,10 +88,8 @@ impl Simd for Avx2 {
     fn level(self) -> Level {
         Level::Avx2(self)
     }
-    #[inline]
```
DJMcNab, line 91: This function should still be `#[inline]` (indeed, I believe that it possibly should be `inline(always)`). This is because we don't want to force LLVM to change its calling convention to call this function, we just want to break unneeded inlining.

fearless_simd/src/generated/neon.rs
```diff
@@ -80,10 +80,8 @@ impl Simd for Neon {
     fn level(self) -> Level {
         Level::Neon(self)
     }
-    #[inline]
```
DJMcNab, line 83: As here

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -93,10 +93,8 @@ impl Simd for Sse4_2 {
             Level::baseline()
         }
     }
-    #[inline]
```
DJMcNab, line 96: And here

fearless_simd/src/generated/fallback.rs
```diff
@@ -146,7 +146,6 @@ impl Simd for Fallback {
         #[cfg(not(feature = "force_support_fallback"))]
         Level::baseline()
     }
-    #[inline]
```
DJMcNab, line 149: (But not here)

fearless_simd/src/generated/wasm.rs
```diff
@@ -80,7 +80,6 @@ impl Simd for WasmSimd128 {
     fn level(self) -> Level {
         Level::WasmSimd128(self)
     }
-    #[inline]
```
DJMcNab, line 83: (Nor here)

PR #185, opened 2026-01-23
Don't use load/store intrinsics

valadaptive:
This is an interesting one! The remaining performance gap in https://github.com/QuState/PhastFT/pull/58 seems to come from subpar performance when loading constants.

I noticed that in Rust's `stdarch`, which defines all the SIMD intrinsics, the x86 load/store intrinsics lower to raw memory operations (`ptr::copy_nonoverlapping`). The AArch64 load/store intrinsics, on the other hand, *do* map to corresponding LLVM intrinsics!

My hypothesis is that the LLVM intrinsics are not lowered until much later in the compilation pipeline, resulting in much fewer optimization opportunities and much worse codegen. If this is the case, we should just use memory operations directly. This also simplifies the code that we generate by quite a bit.

### DIFF
--- a/fearless_simd_gen/src/generic.rs
+++ b/fearless_simd_gen/src/generic.rs
@@ -276,7 +276,7 @@ pub(crate) fn generic_from_array(
 
     // There are architecture-specific "load" intrinsics, but they can actually be *worse* for performance. If they
     // lower to LLVM intrinsics, they will likely not be optimized until much later in the pipeline (if at all),
-    // resulting in substantially worse codegen.
+    // resulting in substantially worse codegen. See https://github.com/linebender/fearless_simd/pull/185.
     let expr = quote! {
         // Safety: The native vector type backing any implementation will be:
         // - A `#[repr(simd)]` type, which has the same layout as an array of scalars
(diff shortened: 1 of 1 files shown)


fearless_simd_gen/src/generic.rs
```diff
@@ -266,33 +266,31 @@ pub(crate) fn generic_block_combine(
...
-            unsafe { #wrapper_ty([
-                #(#load_unaligned(val.as_ptr().add(#blocks) as *const _)),*
-            ]) }
-        }
+        quote! { val }
+    };
+
+    // There are architecture-specific "load" intrinsics, but they can actually be *worse* for performance. If they
+    // lower to LLVM intrinsics, they will likely not be optimized until much later in the pipeline (if at all),
+    // resulting in substantially worse codegen.
```
LaurenzV: Let's also link to this PR for context.

fearless_simd_gen/src/generic.rs
```diff
@@ -333,39 +331,20 @@ pub(crate) fn generic_as_array<T: ToTokens>(
...
-            }
-        });
-        quote! {
-            unsafe {
-                #(#blocks;)*
-            }
+    let store_expr = quote! {
+        unsafe {
+            // Copies `count` scalars from the backing type, which has the same layout as the destination array (see
+            // `generic_as_array`). We know that the source and destination are aligned to at least the alignment of the
```
LaurenzV: Because the slices themselves are guaranteed to have "correct" alignment, right?

Shnatsel: Yes, the existence of a `&[f64]` that doesn't point to aligned `f64`s is UB, see https://doc.rust-lang.org/stable/std/slice/fn.from_raw_parts.html

LaurenzV: Would still be good to clarify in the comment, I think.

PR #188, opened 2026-01-24
Use x86-64-v3 architecture features for AVX2 level

Shnatsel:
More info on x86 architecture levels: https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

I've verified that this allows using the AVX2 level on the earliest CPUs with AVX2 from Intel and AMD:

`rustc --print=cfg --target x86_64-unknown-linux-gnu -C target-cpu=haswell` // earliest Intel AVX2
`rustc --print=cfg --target x86_64-unknown-linux-gnu -C target-cpu=bdver4` // excavator, earliest AMD AVX2

Just in case, I've double-checked with Gemini and it [claims](https://aistudio.google.com/app/prompts?state=%7B%22ids%22:%5B%2217ChKV6ZEfMtSleaUsDoamD8xP5huIr3D%22%5D,%22action%22:%22open%22,%22userId%22:%22100773621717907681037%22,%22resourceKeys%22:%7B%7D%7D&usp=sharing) VIA Nano QuadCore C4650 and Zhaoxin ZX-C have AVX2 without FMA, but this change does not affect them since FMA is already required anyway. 

The newer Zhaoxin KaiXian KX-7000 is missing `cmpxchg16b` and possibly `lzcnt` according to [this](https://openbenchmarking.org/s/ZHAOXIN+KaiXian+KX-7000), so this change does exclude it; we could drop `cmpxchg16b` since it's only used for atomics if supporting obscure Chinese cores is a concern, although I'm not entirely clear on whether they have `lzcnt` or not due to [differences in how it's indicated](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set).

This improves https://github.com/QuState/PHastFT/ performance by 10%-15% but doesn't quite get all the way to `-C target-cpu=x86-64-v3` yet, ~~digging into why is on my TODO.~~ the rest is explained by getting the feature detection out of the hot path: https://github.com/linebender/fearless_simd/pull/188

This does make `Level::new()` marginally more expensive but the README [already instructs](https://github.com/linebender/fearless_simd/tree/main/fearless_simd) users to construct it exactly once.

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -200,7 +200,11 @@ pub enum Level {
         all(target_arch = "aarch64", not(target_feature = "neon")),
         all(
             any(target_arch = "x86", target_arch = "x86_64"),
-            not(target_feature = "sse4.2")
+            not(all(
+                target_feature = "sse4.2",
+                target_feature = "cmpxchg16b",
+                target_feature = "popcnt"
+            ))
         ),
         all(target_arch = "wasm32", not(target_feature = "simd128")),
         not(any(
@@ -334,7 +338,11 @@ impl Level {
             all(target_arch = "aarch64", not(target_feature = "neon")),
             all(
                 any(target_arch = "x86", target_arch = "x86_64"),
-                not(target_feature = "sse4.2")
+                not(all(
+                    target_feature = "sse4.2",
+                    target_feature = "cmpxchg16b",
+                    target_feature = "popcnt"
+                ))
             ),
             all(target_arch = "wasm32", not(target_feature = "simd128")),
             not(any(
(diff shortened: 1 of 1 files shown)


Shnatsel:
@valadaptive could you take a look? This should be straightforward and uncontroversial.

Shnatsel:
Done.

> This also implies we should at some point rename the Level enum variants to X86_V2 and X86_V3 or something, but I'd leave that for later.

I'd actually prefer to keep the enum names as-is. "SSE4.2" and "AVX2" are far more clear than the relatively obscure architecture version numbers that come from LLVM. People who are interested in the finer points of it can read the doc comment.

Shnatsel:
Good catch! I'll update the checks there too.

Shnatsel:
Done.

PR #190, opened 2026-01-25
Explicitly `cast_signed` some masks and sign bits on x86

tomcur:
This fixes the remainder of `clippy::cast_possible_wrap` warnings on x86.

Very similarly to https://github.com/linebender/fearless_simd/pull/186, explicitly reinterpret these bit patterns as signed (i.e., this again is explicitly a no-op at the instruction-level, but effectively checks bit width remains the same). Also uses `cast_signed` for splatting on x86, which just takes the bit pattern as a signed numeric type for broadcasting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,6 @@ This release has an [MSRV][] of 1.88.
 - Breaking change: the `Element` type on the `SimdBase` trait is now an associated type instead of a type parameter. This should make it more pleasant to write code that's generic over different vector types. ([#170][] by [@valadaptive][])
 - The `WasmSimd128` token type now wraps the new `crate::core_arch::wasm32::WasmSimd128` type. This doesn't expose any new functionality as WASM SIMD128 can only be enabled statically, but matches all the other backend tokens. ([#176][] by [@valadaptive][])
 - Breaking change: the `SimdFrom::simd_from` method now takes the SIMD token as the first argument instead of the second. This matches the argument order of the `from_slice`, `splat`, and `from_fn` methods on `SimdBase`. ([#180][] by [@valadaptive][])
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ This release has an [MSRV][] of 1.88.
 - The `WasmSimd128` token type now wraps the new `crate::core_arch::wasm32::WasmSimd128` type. This doesn't expose any new functionality as WASM SIMD128 can only be enabled statically, but matches all the other backend tokens. ([#176][] by [@valadaptive][])
 - Breaking change: the `SimdFrom::simd_from` method now takes the SIMD token as the first argument instead of the second. This matches the argument order of the `from_slice`, `splat`, and `from_fn` methods on `SimdBase`. ([#180][] by [@valadaptive][])
 - Code generation has been improved for shift argument casting on x86 and for scalar fallback. ([#186][] and [#189][] by [@tomcur][])
+- Code generation has been improved for casting of masks and broadcasting of bit patterns on x86. ([#190][] by [@tomcur][])
```
LaurenzV: Will leave it up to you, but I'm not sure such internal changes are significant enough to warrant a changelog update. 😄

xStrom: Yeah, I'd keep changelogs free of internal changes that don't have a significant impact on the user. Frankly we're including too much noise in all of our changelogs.

tomcur: Folded them into one now, but added them as there were other "code generation" entries already.

tomcur: Alright, removed them; agree about the noise. As a general observation, seems there are some different opinions about who we're optimizing these changelogs for (for users or for maintainers of the project itself to skim, and only for the latter knowing about `cast_signed` is potentially interesting in this context).

PR #192, opened 2026-02-02
Better test coverage for 256-bit and 512-bit vectors

Shnatsel:
The newly added tests were generated by Claude Code with Opus 4.5 backend, since there's way too many of them to hand-write.

In the future we might want to add a more rigorous automatic generation of test cases for various vector sizes using an approach along the lines of `fearless_simd_gen`, but this is a simple first step towards increased test coverage.

I kept the newly added tests separate and the original handwritten tests intact to so that this extra code could be easily scrapped if it ever becomes a liability.

The 512-bit test file contains more tests than the 256-bit one because mod.rs already includes some handwritten tests for 256-bit vectors, so they don't have to be duplicated in the 256-bit tests file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,6 +231,59 @@ jobs:
       - name: cargo test --doc
         run: cargo test --doc --workspace --locked --all-features --no-fail-fast --target ${{ matrix.platform.target }}
 
+  test-sde:
+    name: cargo test in an emulator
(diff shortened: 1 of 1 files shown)


Shnatsel:
Github Actions are down right now: https://www.githubstatus.com/incidents/xwn6hjps36ty

But the newly added tests pass for me locally on x86 and in WASM.

PR #193, opened 2026-02-03
Run tests under Address Sanitizer on CI

Shnatsel:
Without a tool like miri or Address Sanitizer memory safety bugs would go undetected.

We can't run intrinsic-heavy code under miri, but we can and should run it under Address Sanitizer on CI.

[Address Sanitizer documentation](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#addresssanitizer) describes what kind of issues it will find. The most likely error in unsafe code is out-of-bounds accesses, which it will always flag. It will not flag alignment or aliasing issues, which miri would catch.

miri would also flag reads from uninitialized memory, and there's a different dedicated sanitizer for it we could enable, but `rg uninit` shows zero uses of uninit memory in fearless_simd so it's not needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,7 @@ jobs:
         with:
           toolchain: nightly
           targets: ${{ matrix.platform.target }}
+          components: rust-src
 
       - name: restore cache
         uses: Swatinem/rust-cache@v2


.github/workflows/ci.yml
```diff
@@ -231,6 +231,33 @@ jobs:
...
+          targets: ${{ matrix.platform.target }}
+          components: rust-src
+
+      - name: restore cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name != 'merge_group' }}
+
+      # doc tests fail to link under ASAN so we use use --lib --bins --tests to exclude them
+      - name: RUSTFLAGS=-Zsanitizer=address cargo +nightly test
```
LaurenzV, line 258: Maybe we can choose a better name here? I first thought this is the command being run.

Shnatsel: It basically is, minus the implementation details.

I followed the line 228 above as reference which calls the job ` cargo nextest` but expands to a complex invocation that boils down to running `cargo nextest`.

I'm happy to change it to any name you like. Edits by maintainers are enabled, you're welcome to rename it directly.

LaurenzV: Ah, I didn't see that we use this naming convention in other places!

PR #195, opened 2026-02-04
Update copyright year for generated code

valadaptive:
Pretty self-explanatory. It'd be nice if there were a way to only update the year for files that were actually updated this year, but that seems like a huge can of worms.


DJMcNab:
I agree with Kaur's analysis. It's messy, because in theory a new file with fully new content should have a newer year, but even in that case, it will still be downstream of the same generator.

I really appreciate you taking on this kind of janitorial task, even though in this case it wasn't necessary.

PR #197, opened 2026-02-04
Run CI tests in Intel SDE with all feature levels

Shnatsel:
Runs tests in Intel Software Development Emulator, like `stdarch` does already, to verify we don't accidentally use an AVX2 instruction on SSE4.2 level or some such.

Part of #124

Some NEON instructions are only available on some cores, and have to checked for at runtime via CPU features, but this PR fully covers the x86 side of things: SSE2 (baseline), SSE4.2, AVX2, AVX-512.

Intel SDE is also the only way to test AVX-512 intrinsics on Github Actions hosted runners, so we'll need it for testing AVX-512 when that arrives. We could have used QEMU for the rest, but QEMU does not support AVX-512.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,8 +278,10 @@ jobs:
         run: ${SDE_PKG}/sde64 -hsw -- cargo test $CARGO_TEST_ARGS
 
       - name: run tests on CPU with AVX-512
-        # Github Actions doesn't give us AVX-512 so this is the only way to exercise AVX-512 codepaths on CI
-        # -icl stands for Ice Lake; technically Skylake added AVX-512 first but it's mostly useless there due to downclocking
+        # Github Actions doesn't give us AVX-512 so this is the only way to exercise AVX-512 codepaths on CI.
+        # -icl stands for Ice Lake. Technically Skylake added AVX-512 first, but it's mostly useless there due to
+        # downclocking. When we do eventually add explicit AVX-512 support, we'll likely target the Ice Lake feature
+        # level.
         run: ${SDE_PKG}/sde64 -icl -- cargo test $CARGO_TEST_ARGS
 
   test-nightly-asan:


.github/workflows/ci.yml
```diff
@@ -231,6 +231,55 @@ jobs:
...
+          toolchain: ${{ env.RUST_STABLE_VER }}
+          targets: ${{ matrix.platform.target }}
+
+      - name: restore cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.event_name != 'merge_group' }}
+
+      - name: install Intel Software Development Emulator
+        run: wget 'https://downloadmirror.intel.com/873619/sde-external-10.5.0-2026-01-13-lin.tar.xz' && tar -xf 'sde-external-10.5.0-2026-01-13-lin.tar.xz'
```
valadaptive:
```suggestion
        run: curl 'https://downloadmirror.intel.com/873619/sde-external-10.5.0-2026-01-13-lin.tar.xz' | tar -Jx
```

.github/workflows/ci.yml
```diff
@@ -231,6 +231,55 @@ jobs:
...
+        # -nhm stands for Nehalem
+        run: sde-external-10.5.0-2026-01-13-lin/sde64 -nhm -- cargo test $CARGO_TEST_ARGS
+
+      - name: run tests on CPU with AVX2
+        # -hsw stands for Haswell
+        run: sde-external-10.5.0-2026-01-13-lin/sde64 -hsw -- cargo test $CARGO_TEST_ARGS
+
+      - name: run tests on CPU with AVX-512
+        # Github Actions doesn't give us AVX-512 so this is the only way to exercise AVX-512 codepaths on CI
+        # -icl stands for Ice Lake; technically Skylake added AVX-512 first but it's mostly useless there due to downclocking
```
valadaptive: I don't think "downclocking" is relevant here because it's an emulator. I'm wondering if we should in fact emulate Skylake here--it supports exactly the AVX-512 subsets that are part of the x86-64-v4 microarchitecture level.

It seems like a good target, since we probably don't want to accidentally use AVX-512 subsets that x86-64-v4 doesn't support. Unless we're fine defining our own bespoke architecture level? I guess `VPOPCNTDQ` could be useful.

.github/workflows/ci.yml
```diff
@@ -231,6 +231,55 @@ jobs:
...
+      - name: install Intel Software Development Emulator
+        run: wget 'https://downloadmirror.intel.com/873619/sde-external-10.5.0-2026-01-13-lin.tar.xz' && tar -xf 'sde-external-10.5.0-2026-01-13-lin.tar.xz'
+
+      # build the tests outside the emulator, otherwise this job would be very slow
+      - name: build tests
+        run: cargo test --no-run $CARGO_TEST_ARGS
+
+      - name: run tests on CPU with only SSE2
+        # -p4p stands for Pentium 4 Prescott, the first Intel 64-bit CPU
+        run: sde-external-10.5.0-2026-01-13-lin/sde64 -p4p -- cargo test $CARGO_TEST_ARGS
```
valadaptive: I wonder if we should make the SDE version string a variable.

.github/workflows/ci.yml
Shnatsel: I don't think there's any value in supporting Skylake in our AVX-512 level. AVX-512 there would just make code slower most of the time, and supporting it would take away a lot of useful extensions like `GFNI` from generations where AVX-512 is actually beneficial.

This was already discussed on Zulip, see [#simd > Levels on x86 @ 💬](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Levels.20on.20x86/near/547012321) although without the specifics.

Ice Lake is chosen because it is the first generation where AVX-512 is actually profitable for performance, and it also includes a lot of useful extensions. It's almost equivalent to Zen 4 as well, the first AMD generation with AVX-512; Zen4 has a couple more extensions but they're fairly inconsequential. I think Ice Lake should be the AVX-512 baseline.

.github/workflows/ci.yml
Shnatsel: I didn't think it would be very useful since the download link has other parts that vary from version to version. But I'm not opposed to it. Edits by maintainers are enabled, please change it if you like.

PR #201, opened 2026-02-21
Initial AVX-512 support

Shnatsel:
This adds the AVX-512 level with Ice Lake as the baseline. This is a deliberate choice: Skylake technically has AVX-512 but is impractically slow, and supporting AVX-512 would both reduce performance on Skylake and reduce performance on other CPUs with actually competent AVX-512 by restricting their instruction sets.

512-bit operations are now backed by native AVX-512 instructions where available.

## Future work

These changes only cover 512-bit vectors. AVX-512 allows more efficient implementations of many ops of smaller sizes, this PR does not tackle them yet.

This keeps the mask representation intact and instead uses extra instructions to cast from masks to full-width vectors and back. This is inefficient but lets us keep the exiting API and avoid dealing with #179 for now.


Shnatsel:
Closing in favor of #231

PR #204, opened 2026-02-22
Add fast approximate reciprocal methods for float vectors

tomcur:
x86 and AArch64 have instructions to calculate fast approximate reciprocals, and these can speed up some algorithms quite nicely (e.g. sprinkling this in Vello's `flatten_simd.rs` results in -4% flattening timings for GhostScript Tiger (actually landing that there requires a bit of thought whether the lowered precision is acceptable of course!)).

There is some detail here that this PR as-is doesn't attempt to solve. x86's `rcp` has about 12 bits of precision, AArch64's `vrecpe` about 8 bits. AArch64 has an additional instruction however, `vrecps`, to perform a Newton refinement step, which bumps the precision to 16 bits. That'd look something like the following.

```rust
let x0 = vrecpeq_f32(a);
x0 * vrecpsq_f32(a, x0); // calculates x0 * (2 - x0 * a), roughly doubling the precision of the `x0` estimate
```

Then, AVX512 introduces `rcp14`, which allows calculating to 14-bit precision with (I believe) the same performance as `rcp`, and extends support to `f64`.

In any case, this method does the simplest thing of just exposing the cheapest hardware estimate, similar to e.g. Highway's `ApproximateReciprocal`.

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -192,6 +192,10 @@ impl Simd for Avx2 {
         unsafe { _mm_sqrt_ps(a.into()).simd_into(self) }
     }
     #[inline(always)]
+    fn approximate_recip_f32x4(self, a: f32x4<Self>) -> f32x4<Self> {
+        unsafe { _mm_rcp_ps(a.into()).simd_into(self) }
+    }
+    #[inline(always)]
     fn add_f32x4(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
         unsafe { _mm_add_ps(a.into(), b.into()).simd_into(self) }
     }
@@ -2182,6 +2186,10 @@ impl Simd for Avx2 {
         unsafe { _mm_sqrt_pd(a.into()).simd_into(self) }
     }
     #[inline(always)]
+    fn approximate_recip_f64x2(self, a: f64x2<Self>) -> f64x2<Self> {
+        self.splat_f64x2(1.0) / a
+    }
+    #[inline(always)]
     fn add_f64x2(self, a: f64x2<Self>, b: f64x2<Self>) -> f64x2<Self> {
         unsafe { _mm_add_pd(a.into(), b.into()).simd_into(self) }
     }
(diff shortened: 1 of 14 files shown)


fearless_simd/src/generated/fallback.rs
```diff
@@ -251,6 +251,10 @@ impl Simd for Fallback {
         .simd_into(self)
     }
     #[inline(always)]
+    fn approximate_recip_f32x4(self, a: f32x4<Self>) -> f32x4<Self> {
+        self.splat_f32x4(1.0) / a
```
LaurenzV: I haven't tried it, does division work without splatting? I think for mutliplication it works at least.

fearless_simd/src/generated/wasm.rs
```diff
@@ -183,6 +183,10 @@ impl Simd for WasmSimd128 {
         f32x4_sqrt(a.into()).simd_into(self)
     }
     #[inline(always)]
+    fn approximate_recip_f32x4(self, a: f32x4<Self>) -> f32x4<Self> {
+        self.div_f32x4(self.splat_f32x4(1.0), a)
```
LaurenzV: Same comment as in fallback

fearless_simd/src/generated/avx2.rs
```diff
@@ -192,6 +192,10 @@ impl Simd for Avx2 {
         unsafe { _mm_sqrt_ps(a.into()).simd_into(self) }
     }
     #[inline(always)]
+    fn approximate_recip_f32x4(self, a: f32x4<Self>) -> f32x4<Self> {
```
LaurenzV, line 195: I'm wondering whether we should just spell reciprocal out. But should be fine this way!

LaurenzV (approved):

fearless_simd/src/generated/avx2.rs
tomcur: Wondered the same thing, but decided to mirror e.g. [`f32::recip`](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.recip).

LaurenzV: Fair!

fearless_simd/src/generated/fallback.rs
tomcur: Done!

PR #206, opened 2026-04-12
Add `interleave` method

Shnatsel:
Adds a new method with API matching std::simd's [`interleave`](https://doc.rust-lang.org/stable/std/simd/struct.Simd.html#method.interleave) method.

The primary motivation is **performance:** on AVX2, `zip_low` followed by `zip_high` requires 6 instructions, while a combined `interleave` function only needs 4 instructions. (With AVX-512 we'd be able to to do it in 2 instructions on 256-bit vectors, but that's not supported by fearless_simd yet and #201 seems stalled).

This also improves API compatibility with `std::simd` as a nice bonus.

AI use disclosure: this work was assisted by Claude Opus 4.5 for the initial commit and 4.6 for the rest. I have manually reviewed the code and take full responsibility for it.


Shnatsel:
I'm seeing non-trivial improvements from this, up to 8% end-to-end for my FFT implementation, so I'd really appreciate this getting reviewed and published as a point release.

LaurenzV:
Haven't forgotten about this, just haven't gotten to it yet. Will try to do soon, if no one else gets to it before me.

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -791,6 +793,68 @@ impl X86 {
...
+    ) -> TokenStream {
+        let unzip_low = generic_op_name("unzip_low", vec_ty);
+        let unzip_high = generic_op_name("unzip_high", vec_ty);
+        quote! {
+            #method_sig {
+                (self.#unzip_low(a, b), self.#unzip_high(a, b))
+            }
+        }
+    }
+
```
LaurenzV, line 857: WHy canw e not apply the same optimization for deinterleave?

LaurenzV (approved):
Seems fine to me, since no one else has commented I presume no one has any objections. Thanks!

fearless_simd_gen/src/mk_x86.rs
Shnatsel: That's an oversight on my part. I was focused on interleaving, since that's the hot path in my code, and forgot to optimize deinterleave. I'll do so shortly.

LaurenzV: It's not big deal, I was just wondering.

PR #207, opened 2026-04-18
Optimize `deinterleave()` for AVX2

Shnatsel:
Follow up to #206

@LaurenzV noted that deinterleaving could be optimized the same way. 

Emitting the proper shuffle is more involved than for zip, but it's already implemented in `handle_unzip()`, so I've simply split it into a helper function and reused it.

This code has good test coverage and all tests pass.


LaurenzV (approved):
Thanks! To be honest I haven't thought this through completely as I'm not super knowledgeable in AVX SIMD, but I trust you and our pretty exhaustive tests all pass so looks good to me!

PR #208, opened 2026-05-02
Consistently use the v2 and v3 x86_64 instruction sets everywhere

Shnatsel:
This shouldn't be an issue on real hardware, but certain emulators with feature combinations that never existed in silicon might complain. Tighten things up just to be sure.

### DIFF
--- a/fearless_simd/src/generated/sse4_2.rs
+++ b/fearless_simd/src/generated/sse4_2.rs
@@ -14,7 +14,7 @@ use crate::{
 use core::arch::x86::*;
 #[cfg(target_arch = "x86_64")]
 use core::arch::x86_64::*;
-#[doc = "The SIMD token for the x86-64-v2 (`sse4.2`, `cmpxchg16b`, and `popcnt`) level."]
+#[doc = "The SIMD token for the x86-64-v2 level."]
 #[derive(Clone, Copy, Debug)]
 pub struct Sse4_2 {
     pub sse4_2: crate::core_arch::x86::Sse4_2,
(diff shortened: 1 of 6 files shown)


fearless_simd/src/generated/sse4_2.rs
```diff
@@ -14,7 +14,7 @@ use crate::{
 use core::arch::x86::*;
 #[cfg(target_arch = "x86_64")]
 use core::arch::x86_64::*;
-#[doc = "The SIMD token for the \"SSE4.2\" level."]
+#[doc = "The SIMD token for the x86-64-v2 (`sse4.2`, `cmpxchg16b`, and `popcnt`) level."]
```
LaurenzV:
```suggestion
#[doc = "The SIMD token for the x86-64-v2 level."]
```

For consistency, since we don't list them for v3 either.

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -61,8 +61,10 @@ impl Level for X86 {
 
     fn token_doc(&self) -> &'static str {
         match self {
-            Self::Sse4_2 => r#"The SIMD token for the "SSE4.2" level."#,
-            Self::Avx2 => r#"The SIMD token for the "AVX2" and "FMA" level."#,
+            Self::Sse4_2 => {
+                "The SIMD token for the x86-64-v2 (`sse4.2`, `cmpxchg16b`, and `popcnt`) level."
```
LaurenzV:
```suggestion
                "The SIMD token for the x86-64-v2 level."
```

LaurenzV (approved):

PR #209, opened 2026-05-02
Correctly feature-gate x86 intrinsics on the `safe_wrappers` feature

Shnatsel:
Previously that feature was only applied to ARM

This is technically semver-breaking but restores the expected behavior and I don't think anyone's depending on this yet, at least not on crates.io


LaurenzV:
Since we removed the module I think this can be closed now. 😄

PR #210, opened 2026-05-16
Prepare v0.4.1 release

LaurenzV:
#208 is not breaking because the restrictions already existed, we just forgot to add them everywhere.

I realized that _in theory_ #207 would be breaking because while the `Simd` trait requires `Seal`, `SimdInt` and `SimdBase` currently do not, I think that's an oversight. However, those traits are not intended to be implemented manually anyway, so let's just fix that in a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,6 +191,7 @@ No changelog was kept for this release.
 [#181]: https://github.com/linebender/fearless_simd/pull/181
 [#185]: https://github.com/linebender/fearless_simd/pull/185
 [#188]: https://github.com/linebender/fearless_simd/pull/188
+[#206]: https://github.com/linebender/fearless_simd/pull/206
 [#207]: https://github.com/linebender/fearless_simd/pull/207
 [#208]: https://github.com/linebender/fearless_simd/pull/208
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -179,8 +191,11 @@ No changelog was kept for this release.
 [#181]: https://github.com/linebender/fearless_simd/pull/181
 [#185]: https://github.com/linebender/fearless_simd/pull/185
 [#188]: https://github.com/linebender/fearless_simd/pull/188
+[#207]: https://github.com/linebender/fearless_simd/pull/207
```
Shnatsel:
```suggestion
[#188]: https://github.com/linebender/fearless_simd/pull/188
[#206]: https://github.com/linebender/fearless_simd/pull/206
[#207]: https://github.com/linebender/fearless_simd/pull/207
```

Otherwise the link above doesn't work

Shnatsel (approved):
One broken link, otherwise LGTM.

FYI you can use `[foo]` instead of `[foo][]` syntax, which is less verbose.

LaurenzV:
> FYI you can use [foo] instead of [foo][] syntax, which is less verbose.

I'm also not sure _why_ we do this but I believe there is a reason, it's also done in the other repos. Therefore, I'll leave it as is for now.

DJMcNab:
We use `[foo][]` as that indicates to markdownlint that this is a link.

In particular, that's useful because forgetting to add links is really easy.

PR #211, opened 2026-05-16
Mark crate types as sealed

LaurenzV:
See also the discussion in https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/.E2.9C.94.20Making.20traits.20in.20fearless_simd.20require.20.60Seal.60/near/595462829. We don't intend these to be impelemented externally, so mark as sealed. This way,  adding new methods isn't a breaking change anymore.


Shnatsel:
This is an improvement but the sealing isn't entirely complete. `SimdCvtTruncate<T>` and `SimdCvtFloat<T>` only seal `Self`, leaving `T` unconstrained.  Here's a PoC that still compiles with an ostensibly-sealed trait: 

[fsimd_review_probe.zip](https://github.com/user-attachments/files/27852456/fsimd_review_probe.zip)

(Security note: be careful with codebases sent in a zip archive rather than received via `git clone`, they may embed git hooks that cause arbitrary code execution on any `git` command, including the shell displaying the branch name)

LaurenzV:
Do you have it in a git branch so I can see it there?

LaurenzV:
Good catch, thanks!

Shnatsel:
Here it is added to the git repo: https://github.com/Shnatsel/fearless_simd/tree/insufficient-sealing-poc

LaurenzV:
Yeah LGTM 👍 if you think it's okay now feel free to approve!

Shnatsel (approved):
LGTM. Thanks!

If we don't do this now, all our subsequent releases are semver hazards too. So I think we want to ship this sooner rather than later.

PR #213, opened 2026-05-17
Show NEON and WASM docs on docs.rs

Shnatsel:
Now that docs.rs [only builds one target by default](https://blog.rust-lang.org/2026/04/04/docsrs-only-default-targets/), we need to explicitly specify the list of targets to show NEON and WASM documentation on docs.rs.

This also enables the `simd128` target feature on docs.rs, which might not match the default local build, but whichever option we pick (enabling simd128 or only showing fallback or having no wasm docs at all) is going to be confusing in some way, so it's a "pick your poison" kind of situation.

### DIFF
--- a/fearless_simd/Cargo.toml
+++ b/fearless_simd/Cargo.toml
@@ -13,9 +13,17 @@ readme = "README.md"
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
Thanks!

Issue #166, opened 2025-12-13
Rework the `safe_wrappers` feature / downcasting ability


valadaptive:
There's a proof-of-concept `safe_wrappers` feature that's supposed to let you use your own custom architecture-specific SIMD code if you want to, by obtaining an architecture-specific SIMD token from `Level` using `as_neon`, `as_avx2`, etc.

This isn't the ideal abstraction for a couple reasons:
1. You need to pass the `Level` all the way down into every SIMD function you use. 
2. The `as_[platform]` methods on `Level` incur a runtime cost. When a SIMD-using function calls some method on the `Simd` token it's passed, that token's type is statically known, because it's a type parameter. `Level`, however, is always an enum.

I think https://github.com/linebender/fearless_simd/pull/108/ starts to move in a different direction, but it creates a lot of autogenerated code, is not fully implemented (WASM is missing), and doesn't actually tie into fearless_simd as implemented. It also seems to have stalled out.

In the meantime, I propose adding `as_[platform]` methods on the `Simd` trait *itself*. The default method implementations would all return `None`, and individual implementations (`Sse4_2`, `Neon`, etc.) could override specific methods to all return `Some(self)`. This means we don't need to pass down the `Level`, and the new `as_[platform]` methods would be statically dispatched.

(/cc @DJMcNab; how much does this relate to #108?)

DJMcNab:
> is not fully implemented (WASM is missing)

So, #108 can't change anything about WASM, because the SIMD level is always global on WASM. 
It is the case that it's not implemented for aarch64 (to be specific, the core abstraction works for all platforms, but the).

> doesn't actually tie into fearless_simd as implemented.

Yes, that's correct; I still believe that the abstraction is valuable for consumers of Fearless SIMD, but I'm less certain about my prior comments around using it to implement Fearless SIMD itself; I do suspect that the overhead would be too much. Ideally, of course, rustc would just give us https://github.com/rust-lang/rfcs/pull/3820 (which is implemented in https://github.com/rust-lang/rust/pull/145573), making that need moot.

> It also seems to have stalled out.

I've just revisited it as part of writing this comment, and realised that this stalling out is my fault. Hopefully we can get it back into gear now.

<hr/>

I'm in favour of the proposed `as_platform` methods. Those wouldn't be needed in a world with #108 (or at least, the natural extension to it), but that could take a while to actually get moving.

Shnatsel:
Fulfilled by #214 and #216

PR #214, opened 2026-05-18
Even KISS-er access to intrinsics

Shnatsel:
An even simpler implementation of the concept from #212, thanks to the review feedback by @DJMcNab

Alternative to https://github.com/linebender/fearless_simd/pull/108 and https://github.com/linebender/fearless_simd/pull/196

Part of #166


fearless_simd/src/kernel_macros.rs
```diff
@@ -0,0 +1,266 @@
...
+
+            // SAFETY: the SIMD token proves that the required target features are available.
+            #[allow(unused_unsafe, reason = "for WASM which has no target feature requirements and is safe to call")]
+            unsafe { __fearless_simd_kernel($token $(, $arg)*) }
+        }
+    };
+}
+
+#[cfg(test)]
+mod tests {
```
LaurenzV, line 207: Maybe an x86 test as well? Or is there a reason not to?

LaurenzV (approved):
I have not tried to manually parse all of the macro arguments etc., but tests pass so LGTM! Although I would say #166 should only be closed once we have ripped out the old `safe_wrappers` code.

fearless_simd/src/kernel_macros.rs
Shnatsel: It's covered by the doctest but sure, why not.

fearless_simd/src/kernel_macros.rs
```diff
@@ -0,0 +1,296 @@
...
+/// ```compile_fail
+/// fearless_simd::kernel! {
+///     unsafe fn should_not_compile(avx2: Avx2) {}
+/// }
+#[macro_export]
+macro_rules! kernel {
+    (
+        $(#[$meta:meta])*
+        $vis:vis fn $name:ident(
+            $token:ident : $token_ty:ident $(, $arg:ident : $arg_ty:ty)* $(,)?
```
DJMcNab, line 72: What do you think about optionally accepting a `fearless_simd::` prefix before the type here?

This isn't blocking.

fearless_simd/src/kernel_macros.rs
```diff
@@ -0,0 +1,296 @@
...
+        $(#[$meta:meta])*
+        $vis:vis fn $name:ident(
+            $token:ident $(, $arg:ident : $arg_ty:ty)* $(,)?
+        ) $(-> $ret:ty)? {
+            $($kernel_body:tt)*
+        }
+    ) => {
+        #[cfg($cfg)]
+        $(#[$meta])*
+        $vis fn $name(
```
DJMcNab, line 187: I believe that this outer function should be `#[inline(always)]` in most cases. But I don't know how that works with the inner function being marked `#[inline]`.

I think this is something which can be tuned. Plus, users can (currently) add their own `#[inline(always)]`, so this isn't blocking.

DJMcNab (approved):

fearless_simd/src/kernel_macros.rs
Shnatsel: I erred on the side of letting users tune the inlining. 

I've already hit this in practice when using fearless_simd for PhastFT, and made a PR to reduce the overly aggressive inlining in an unrelated part of fearless_simd: #183

fearless_simd/src/kernel_macros.rs
Shnatsel: It's not that hard, just a lot of boilerplate: https://github.com/Shnatsel/fearless_simd/commit/897f83e5ec57682343a8a1ae9195f83b969df10a

This is a cosmetic issue and I don't really see the point of supporting spelling out the levels like that, so this is not where I'd prefer to spend the complexity budget. I'm happy to reconsider if we have actual user feedback to this effect.

fearless_simd/src/kernel_macros.rs
```diff
@@ -0,0 +1,296 @@
+// Copyright 2025 the Fearless_SIMD Authors
```
ajakubowicz-canva, line 1: Extreme nit drive by. Should this be `2026`? (No action required)

PR #216, opened 2026-05-18
Rip out `core::arch` wrappers

Shnatsel:
Remove the wrappers for intrinsics inherited from `pulp` now that #214 provides a way to access intrinsics without us having to wrap every single one.

Part of https://github.com/linebender/fearless_simd/issues/166

Supersedes #209


LaurenzV:
Should this be merged before #214 or after?

Shnatsel:
I would prefer merging this after #214. To keep CI passing I had to create some editing conflicts in samples, I'd rather resolve them here by simply overriding them with the samples from #214.

LaurenzV:
I approved the other PR, not sure if you wanted to wait for input from Daniel though.

Shnatsel:
Thanks! I'll wait a few days to let Daniel and the others chime in. This isn't blocking any follow-up work as far as I know. 

If it holds anything up, please let me know and I'll go ahead and merge it.

Shnatsel:
Now that #214 is in, this should also be good to go. Daniel has explicitly stated he supports removing this: https://github.com/linebender/fearless_simd/pull/212#pullrequestreview-4305703483

I'd appreciate an approval on this PR

DJMcNab:
I've not checked yet, as I'm on mobile. There aren't two levels of tokens now, right, or are there?
Would you think we'll tear that out at some point?

Shnatsel:
There aren't two levels, no. The tokens were simply in the same file as all the macros to generate wrappers.

LaurenzV (approved):
LGTM, just two things but can/should be follow-ups:
1) Is there even any point in keeping the SSE3/FMA/etc. tokens since those aren't supported natively anyway?
2) We probably should remove the subfolders and put everything into a single file per architecture or even just everything into one file.

LaurenzV:
(Removing it from the queue just to make sure you saw my comment, feel free to re-add)

DJMcNab:
I mean these structures:

```rust
pub struct Avx2 {
    pub avx2: crate::core_arch::x86::Avx2,
}
```

I'm on mobile, so it's not easy to browse your branch, but I don't think those are collapsed. It's fine for that to be another follow-up, fwiw

Shnatsel:
Ah, good catch! I totally missed that the token structs are wrapped and are indeed duplicated. 

I'll remove them before re-enabling auto-merge. Thanks!

Shnatsel:
Committing file deletions to git is hard :sweat_smile:

fearless_simd/src/lib.rs
```diff
@@ -417,8 +412,6 @@ impl Level {
     /// because if Fearless SIMD gets support for an instruction set which is a superset of Neon,
     /// this method will return a value even if that "better" instruction set is available.
     ///
-    /// This can be used in combination with the `safe_wrappers` feature to gain checked access to
-    /// the level-specific SIMD capabilities.
```
DJMcNab, line 421: Possibly we should point to the kernel macro here. This isn't urgent

PR #217, opened 2026-05-18
Add from_array/to_array to SIMD vectors for std::simd compatibility

Shnatsel:
Convenience/compatibility wrapper around the existing `SimdFrom` and `From` implementations.

This closes the API gap with `std::simd`. I also find this more discoverable and readable.

This is not applied to mask types because in std::simd they operate on `[bool; N]` rather than integer types such as `[i64; N]`, so that would be an API conflict.


Shnatsel:
Marking this as a draft until the implications for generics are better understood.

Also the API mismatch between masks and regular vectors is tricky. I don't exactly want to expose APIs via arrays of bools because that's slower than the current approach.

Shnatsel:
I'll revisit this later, closing to remove from the list of PRs for review

PR #218, opened 2026-05-18
Make mask types opaque

Shnatsel:
This is necessary for eventual AVX-512 support. Part of #179.

This does not add any AVX-512 stuff yet, just lays the groundwork by abstracting away the internal representation.

Contrary to what the description of #196 said, we don't actually need i64 vectors in the public API so long as we're not providing direct conversions between mask types and integer vector types, which aren't available even on main.

Summary of changes:

- Add `SimdMask<S>` trait independent from `SimdBase<S>`.
- Remove integer-vector-style APIs from masks:
  - `Deref`
  - Indexing
  - `Bytes`
  - public `SimdSplit` / `SimdCombine`
  - public `slide` / `slide_within_blocks`
  - public byte conversions
- Remove scalar bit-op overloads for masks, so masks support mask-to-mask `& | ^ !` but not `mask & -1`.

This was extricated from a larger changeset I had locally that also added some std::simd API compatibility functions, but that was getting too complicated to review. I'm happy to add more APIs if there's desire and review capacity for them.

A port of vello to this API can be found [here]( https://github.com/linebender/vello/compare/main...Shnatsel:opaque-masks?expand=1).


Shnatsel:
The test failure from a completely unrelated part is spooky. I checked it for UB under miri with various CPU feature combinations, but it passes there, as well as on real hardware (locally and in CI) and on Intel's emulator on CI.

I'm pinning it on a Rosetta bug, since real hardwware and 2 other emulators all pass.

When running x86 code on the macos-latest image, it runs through Rosetta. We need to switch over to `macos-26-intel` runner for x86 macos to get real x86 hardware.

Shnatsel:
Porting vello was trivial: https://github.com/linebender/vello/compare/main...Shnatsel:opaque-masks?expand=1

So this is ready for review.

fearless_simd/src/generated/avx2.rs
```diff
@@ -890,82 +890,10 @@ impl Simd for Avx2 {
...
-    fn load_array_ref_mask8x16(self, val: &[i8; 16usize]) -> mask8x16<Self> {
-        mask8x16 {
-            val: unsafe { core::mem::transmute_copy(val) },
-            simd: self,
-        }
-    }
-    #[inline(always)]
     fn as_array_mask8x16(self, a: mask8x16<Self>) -> [i8; 16usize] {
         unsafe { core::mem::transmute::<__m128i, [i8; 16usize]>(a.val.0) }
     }
```
LaurenzV, line 896: Why can we still keep this method? Wouldn't this also expect that the inner representation is backed by actual integers?

LaurenzV: And similarly, how would `load_array_mask8x16` be implemented for AVX512 if it takes the array as an integer array?

fearless_simd/src/generated/simd_trait.rs
```diff
@@ -392,36 +396,12 @@ pub trait Simd:
     fn widen_u8x16(self, a: u8x16<Self>) -> u16x16<Self>;
     #[doc = "Reinterpret the bits of this vector as a vector of `u32` elements.\n\nThe total bit width is preserved; the number of elements changes accordingly."]
     fn reinterpret_u32_u8x16(self, a: u8x16<Self>) -> u32x4<Self>;
-    #[doc = "Create a SIMD vector with all elements set to the given value."]
+    #[doc = "Create a SIMD mask with all lanes set from the given signed integer mask value."]
     fn splat_mask8x16(self, val: i8) -> mask8x16<Self>;
```
LaurenzV: How would `splat_mask` be implemented for AVX-512 since the representation of a single entry is always just a single bit?

I guess more fundamentally, the question is how we fundamentally want to allow construction of masks while preventing hidden footguns due to the exact internals of the SIMD level. For example, I imagine a `from_bitmask` method would be fast for AVX-512 since its the native representation but slow on NEON since you need to expand the bitmask manually, while a `from_array`, as it is now, is the best for NEON but would require manually encoding the mask into a single bitmask. I'm not sure what the best thing to do is here, but probably we can take inspiration from portable SIMD?

fearless_simd_gen/src/mk_simd_trait.rs
```diff
@@ -80,14 +83,13 @@ pub(crate) fn mk_simd_trait() -> TokenStream {
             type i32s: SimdInt<Self, Element = i32, Block = i32x4<Self>, Mask = Self::mask32s, Bytes = <Self::u32s as Bytes>::Bytes> + SimdCvtTruncate<Self::f32s>
                 + core::ops::Neg<Output = Self::i32s>;
             /// A native-width SIMD mask with 8-bit lanes.
-            type mask8s: SimdMask<Self, Element = i8, Block = mask8x16<Self>, Bytes = <Self::u8s as Bytes>::Bytes> + Select<Self::u8s> + Select<Self::i8s> + Select<Self::mask8s>;
+            type mask8s: SimdMask<Self, Element = i8> + SimdFrom<i8, Self> + Select<Self::u8s> + Select<Self::i8s> + Select<Self::mask8s>;
```
LaurenzV: Why do we need to add `SimdFrom`?

LaurenzV (commented):
Haven't finished looking through it, just some first comments.

fearless_simd/src/generated/avx2.rs
Shnatsel: We can't keep `Deref` into an integer slice/array, but explicit conversion into an owned instance is perfectly fine and still useful. So conversions to/from the integer array representation are left intact in the API on purpose. They're the cheapest way to create a mask right now, and are decently cheap even on AVX-512 (a single instruction).

fearless_simd/src/generated/simd_trait.rs
Shnatsel: It would check if the value is 0, and treat any other value as 1.

I've actually started this work by taking inspiration from `std::simd`, and I have a local branch that is a much more complete mirror of `std::simd` APIs. This PR is a stripped down, MVP version of it.

`std::simd` only implements construction of masks from bits represented as a `u64` and from arrays of booleans; not from [i32; 4]. Their trade-off is greater portability but worse performance. I implemented that at first, then rolled back the removal of integer-based APIs because the assembly for bool arrays looked gnarly.

Now that you point it out, I agree it's probably better to mirror std::simd and make `splat` operate on booleans. I guess writing `== 0` is not too much to ask of the users. I'll make that change to better align with `std::simd` in this case.

fearless_simd_gen/src/mk_simd_trait.rs
Shnatsel: It's for `splat`. This requirement used to be inherited from `SimdBase` which defines `SimdFrom<Self::Element, S>` but now that we no longer implement `SimdBase` on masks we need to require it explicitly.

fearless_simd/src/generated/simd_trait.rs
Shnatsel: I've experimented with this and looked at the assembly; it does add a `neg` instruction in the hot path for non-constant inputs on x86, so it's technically less efficient, but that's probably worth it for the nicer API.

fearless_simd_gen/src/mk_simd_trait.rs
Shnatsel: I've reintroduced `.splat()` now that it matches `std::simd` API, and removed this bound now that it's no longer necessary (generic code can directly use .splat())

Shnatsel:
I'd like to add [to_bitmask()](https://doc.rust-lang.org/stable/std/simd/struct.Mask.html#method.to_bitmask)/from_bitmask() as well as [`test`](https://doc.rust-lang.org/stable/std/simd/struct.Mask.html#method.test) and [set](https://doc.rust-lang.org/stable/std/simd/struct.Mask.html#method.set) to match `std:simd`, but I'll do it in a follow-up PR to avoid complicating review of this one.

LaurenzV:
Will try to review the rest after RustWeek!

fearless_simd_gen/src/types.rs
```diff
@@ -178,8 +178,8 @@ impl VecType {
         if self.scalar == ScalarType::Mask {
             let scalar_bits = self.scalar_bits;
             format!(
-                "A SIMD mask of {len} {scalar_bits}-bit elements.\n\n\
-                When created from a comparison operation, and as it should be used in a [`Self::select`] operation, each element will be all ones if it's \"true\", and all zeroes if it's \"false\".",
+                "A SIMD mask of {len} logical lanes corresponding to {scalar_bits}-bit vector elements.\n\n\
+                The storage representation of this type is intentionally opaque. For compatibility with existing APIs, it may be converted to and from signed integer lanes where false is encoded as all zeroes (integer value 0) and true is encoded as all ones (integer value -1).",
```
LaurenzV, line 182: Maybe we should add a note that the conversion to/from signed integers might not be zero-cost for all levels though?

fearless_simd_gen/src/ops.rs
```diff
@@ -955,9 +989,10 @@ const INT_OPS: &[Op] = &[
 // `concat!` macro.
 macro_rules! mask_reduce_blurb {
     () => {
-        "Behavior on mask elements that are not all zeroes or all ones is unspecified. It may vary depending on architecture, feature level, the mask elements' width, the mask vector's width, or library version.\n\n\
-        The behavior is also not guaranteed to be logically consistent if mask elements are not all zeroes or all ones. `any_true` may not return the same result as `!all_false`, and `all_true` may not return the same result as `!any_false`.\n\n\
-        The [`select`](crate::Select::select) operation also has unspecified behavior for mask elements that are not all zeroes or all ones. That behavior may not match the behavior of this operation."
+        "Masks may be converted to and from signed integer lane arrays for compatibility with older APIs. For those conversions, false is encoded as all zeroes (integer value 0) and true is encoded as all ones (integer value -1).\n\n\
```
LaurenzV, line 992: Same here

LaurenzV (approved):
LGTM, I haven't tried to understand the new codegen logic very thoroughly, but the new generated code seems good to me and I agree that making the mask representation opaque is the right move towards getting AVX-512 working. It's a bit unfortunate that there is now more code duplication in the codegen code, but not a big deal.

LaurenzV:
It seems like you might have to rebase or open a new PR to make CI work correctly again.

fearless_simd_gen/src/types.rs
Shnatsel: I'll describe the efficiency in a follow-up where I add other conversion paths

Shnatsel:
I've reduced the duplication in the generation of splat(). 

There is more refactoring we could do, but I haven't found the right cut points for shared logic yet. The conditions end up being more awkward and hard to follow when they're not defined at the leaf.

PR #219, opened 2026-05-18
Run x86 apple job on x86 hardware

Shnatsel:
`macos-latest` is ARM, and runs x86 binaries through rosetta. We need to specify the Intel runner explicitly to get x86 hardware.

Motivation: https://github.com/linebender/fearless_simd/pull/218#issuecomment-4482607401

Our test suite is now so thorough that it caught a Rosetta bug :sunglasses:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,7 +202,7 @@ jobs:
       matrix:
         platform:
           - { target: aarch64-apple-darwin, os: macos-latest }
-          - { target: x86_64-apple-darwin, os: macos-latest }
+          - { target: x86_64-apple-darwin, os: macos-26-intel }
           - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
           - { target: x86_64-pc-windows-msvc, os: windows-latest }
           - { target: i686-pc-windows-msvc, os: windows-latest }


LaurenzV (approved):
I guess the question is how long those will be supported by GitHub, but for now LGTM!

Shnatsel:
They're maintained until fall 2027, but Apple is [removing Rosetta in 2028](https://github.com/Stremio/stremio-service/issues/111) anyway.

Shnatsel:
Hmm, this looks like a flaw in github merge queue implementation: main has a CI check that this branch doesn't, so auto-merge is never going to happen because one of the checks will never execute.

Someone with greater access than me will need to bypass merge queue somehow, e.g. disable it, merge this PR and then re-enable it. Or perhaps the required checks are listed explicitly somewhere?

LaurenzV:
I don’t have these permissions I think

LaurenzV:
@DJMcNab Do you have the permissions for this?

DJMcNab:
Only on desktop, and I'm on mobile atm. Do ping me again if I don't sort this today.

raphlinus:
I've tried to do this with `gh pr merge 219 --admin` but it tells me "Changes must be made through the merge queue". I think we need to disable, merge, then re-enable.

PR #220, opened 2026-05-19
Add slice op soundness tests

Shnatsel:
The vast majority of our operations accept fixed-size inputs with the lengths encoded in type system, so they are pretty hard to mess up. We also have Address Sanitizer on CI verifying that all the raw intrinsics actually stick to those sizes.

The only exception are from_slice/store_slice APIs, which accept variable-length slices. This PR adds tests verifying that slices that are too short are correctly rejected.

### DIFF
--- a/fearless_simd_tests/Cargo.toml
+++ b/fearless_simd_tests/Cargo.toml
@@ -9,6 +9,8 @@ description = "Safer and easier SIMD"
 readme = "README.md"
 rust-version.workspace = true
 publish = false
+# Tests are collected through tests/mod.rs; files beside it are modules, not separate integration crates.
+autotests = false
 
 [[test]]
 name = "tests"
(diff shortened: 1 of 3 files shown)


LaurenzV:
Instead of doing this unwinding thing, can we not simply change the `simd_test` macro to accept an additional attribute `should_panic` that propagates it to the test?

Shnatsel:
We could, but that makes the proc macro even more complex and magical, and I'm trying to keep things simple and stupid to make them easier to maintain in the long run. This already uses a macro to reduce duplication so I felt I've already spent all the complexity budget.

Plus I didn't expect this to be used often (there are literally two methods where this is even needed) so I didn't think it was worth complicating the common path for the sake of this.

Shnatsel:
I've heard that `cargo nextest` might be able to run `#[should_panic]` tests on WASM which doesn't have unwinding, since it runs each test in a separate process. But turns out that's not true. Instead of running the test and checking if it terminated, `cargo nextest` simply skips `#[should_panic]` tests on targets without unwinding.

LaurenzV (approved):

PR #221, opened 2026-05-19
Mention access to platform-specific intrinsics in the README

Shnatsel:


LaurenzV (approved):

fearless_simd/README.md
```diff
@@ -109,6 +109,10 @@ E.g. We might want names for these, e.g.:
 TODO: Talk about writing versions of functions which can be called in other `S: Simd` functions.
 -->
 
+# Platform-specific intrinsics
+
+If the portable APIs are not enough, you can safely invoke platform-specific intrinsics via the [`kernel!()`](kernel) macro.
```
DJMcNab, line 114: I think this needs a `[kernel]: https://docs.rs/...` link reference definition.
I'm happy to add that myself, although I am liable to forget about this.

Shnatsel: I just followed the existing conventions for intra-doc links. If we change it we should change it across the board.

The current approach is great for rustdoc/docs.rs, but probably show up as broken in the README. I'm surprised `cargo rdme` doesn't handle this.

DJMcNab: I don't think I explained this clearly (which is why I suggested doing it myself). To clarify though, I don't see that this follows the existing pattern...
On lines 31 to 39, there are existing link reference definitions, but one needs to be added for this.

Shnatsel: I don't think I understand, so I'd appreciate a follow-up PR.

PR #223, opened 2026-05-20
For `Token::new_unchecked`, use `target_feature(enable = )` attributes

DJMcNab:
This removes some instances of `unsafe` for having target features enabled. It also allows the use of target features 1.1 to transfer into Fearless SIMD safely.

A few unanswered questions:
- Should we rename `new_unchecked` to `new`?
- Does it even make sense for these to be const?
- Does this prevent inlining of these functions?

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -20,15 +20,16 @@ pub struct Avx2 {
     _private: (),
 }
 impl Avx2 {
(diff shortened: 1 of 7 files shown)


DJMcNab:
Current state: I'd claim that the clippy failure is a clippy bug, but I might be able to gut check with Ralf at some point this week.

Shnatsel:
Superseded by #293

PR #224, opened 2026-05-20
Fix some doc links

DJMcNab:
Follow-up from #221

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The other packages are as follows:
 
 ## History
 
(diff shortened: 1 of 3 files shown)


Shnatsel (approved):
There are other broken links that could be given the same treatment:

Image

and

Image

Just to name a few.

We might as well fix them for consistency?

Shnatsel:
Removing the queue to give a chance to brush up more links in the same PR, but you're welcome to merge as-is if you prefer.

Shnatsel (approved):
That's an improvement, but the links are still somewhat patchy:

Image
Image

That's not a blocker and can be addressed in a follow-up PR.

Shnatsel:
LGTM. Thanks!

DJMcNab:
Fwiw, that `Simd` one is kind of intentional. We don't have a defined format for how often we should link to the same link.

PR #225, opened 2026-05-20
Make our READMEs consistent

DJMcNab:
Document how to run `cargo rdme` in the lib.rs files.
Also change the base level now that we use level 1 headings in our docs


Shnatsel (approved):
Looks good, thanks!

PR #227, opened 2026-05-23
Remove `nextest` to speed up CI

Shnatsel:
`cargo nextest`'s claim to fame is faster performance in some cases. However, its model of one process per test maps especially poorly to `fearless_simd` test structure.

We have over a thousand very small tests, so process spawning overhead really adds up. Locally `cargo test` is 4x faster than `cargo nextest` on my x86 machine.

`cargo nextest` overhead gets particularly egregious for WASM tests, where it starts up a whole new instance of `wasmtime` for every single test, which adds up to over 3 minutes to run the test suite even on my beefy machine. `cargo test` completes in just a few seconds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,14 +374,16 @@ jobs:
 
       - name: cargo test
         run: |
-          cargo test --lib --bins --tests --workspace --locked --all-features --target wasm32-wasip1 \
+          cargo test --workspace --locked --all-features --no-fail-fast --target wasm32-wasip1 \
             --config 'target.wasm32-wasip1.rustflags = "-Ctarget-feature=+simd128"' \
+            --config 'target.wasm32-wasip1.rustdocflags = "-Ctarget-feature=+simd128"' \
             --config 'target.wasm32-wasip1.runner = "wasmtime"'
 
       - name: cargo test (w/ relaxed-simd)
         run: |
-          cargo test --lib --bins --tests --workspace --locked --all-features --target wasm32-wasip1 \
+          cargo test --workspace --locked --all-features --no-fail-fast --target wasm32-wasip1 \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -215,21 +211,13 @@ jobs:
...
         with:
           save-if: ${{ github.event_name != 'merge_group' }}
 
-      - name: cargo nextest
-        run: cargo nextest run --workspace --locked --all-features --no-fail-fast --target ${{ matrix.platform.target }}
-
-      - name: cargo test --doc
-        run: cargo test --doc --workspace --locked --all-features --no-fail-fast --target ${{ matrix.platform.target }}
+      - name: cargo test
+        run: cargo test --workspace --locked --all-features --no-fail-fast --target ${{ matrix.platform.target }}
```
LaurenzV, line 220: Are you sure this will test the docs? Since this was removed.

.github/workflows/ci.yml
```diff
@@ -384,28 +372,16 @@ jobs:
...
-            --config 'target.wasm32-wasip1.runner = "wasmtime"'
-
-      - name: cargo nextest (w/ relaxed-simd)
-        run: |
-          cargo nextest run --workspace --locked --all-features --no-fail-fast --target wasm32-wasip1 \
-            --config 'target.wasm32-wasip1.rustflags = "-Ctarget-feature=+simd128,+relaxed-simd"' \
-            --config 'target.wasm32-wasip1.runner = "wasmtime"'
-
-      - name: cargo test --doc
+      - name: cargo test
```
LaurenzV, line 375: Same for those.

LaurenzV (approved):

.github/workflows/ci.yml
Shnatsel: Yes, Cargo runs doctests by default. You can verify that by e.g. running `cargo test` under Address Sanitizer and watch them fail to link.

PR #228, opened 2026-05-24
PoC: x86 CPU levels behind feature flags

Shnatsel:
Codex-driven sketch of what it would take to put x86 CPU levels behind Cargo feature flags. It's not as bad as I feared.

Part of #92

---

SSE4.2 and AVX2 are now independent Cargo features, enabled by default.

Key changes:
- Added `sse4_2` and `avx2` features in `Cargo.toml`.
- Gated x86 backend modules so disabled backends are not compiled at all in `generated.rs`.
- Gated `Level` variants, detection, `baseline`, accessors, and dispatch arms in `lib.rs` and `macros.rs`.
- Kept exported macros correct cross-crate by passing feature-state booleans from cfg-selected helper macros, not by relying on downstream `#[cfg(feature = ...)]`.
- Updated `kernel!` so `Sse4_2`/`Avx2` token kernels require the matching crate feature.
- Fixed the compiler-discovered AVX2-only case: AVX2 internally needs 128-bit x86 arch conversions that previously came from the SSE module. The generator now emits those from AVX2 under `#[cfg(not(feature = "sse4_2"))]` via `level.rs` and `mk_x86.rs`.

Verification passed:
- `cargo check -p fearless_simd` across default, `std`, `std,avx2`, `std,sse4_2`, `--all-features`
- same feature matrix under `RUSTFLAGS='-Ctarget-cpu=x86-64-v3'`
- `cargo clippy -p fearless_simd ... -- -D warnings` across default, all-features, fallback-only, AVX2-only, SSE-only
- `cargo test -p fearless_simd --lib`
- `cargo test -p fearless_simd --lib --no-default-features --features std`
- `cargo test -p fearless_simd_tests`
- downstream temp-crate smoke checks for `dispatch!` with dependency features `std`, `std,avx2`, and `std,sse4_2`

No broad dead-code suppression was needed.


Shnatsel:
Closing in favor of #258

PR #229, opened 2026-05-24
PoC: add no-std libm tests to CI

Shnatsel:


DJMcNab:
What needs to happen here? It looks like all the tests pass (which is good news)

Shnatsel:
I don't remember. I quickly slapped this together and got distracted doing something else.

So a thorough review, I guess?

PR #231, opened 2026-05-24
Add AVX-512 support

### LINKED ISSUE #179: Mask types for AVX-512

Shnatsel:
Yes, really. It's all here. In one humongous PR. Sorry :sweat_smile: 

This is probably best reviewed commit-by-commit. The first commit is still big because the history was getting really messy with changes and rollbacks, and squashing it made it less of a mess.

This also touches other backends in three ways:

1. `set_mask()` is now a backend method so it could be specialized per-level
2. Changes to mask conversion routines to support different internal representations bled into other levels. It occasionally adds an intermediate array but it gets optimized out in practice.
3. `transmute_copy()` is wrapped into `checked_transmute_copy()` and the raw version disallowed after I almost had a horrible accident with it. ~~This could be its own PR but I wanted the insurance right away.~~ This was split and shipped in v0.5.0

Everything changed here should be covered by tests. I've expanded test coverage where it was lacking.

Closes #179


LaurenzV:
I think it would indeed be great to have a custom PR for 3.

Shnatsel:
It will cause a lot of conflicts if I try to split it, but I have it isolated to its own commit at least: f08f7e6dcd3ac5ddaaa8d6f035dc7ab38be46db0

Shnatsel:
I've run miri on the entire test suite (minus the exhaustive tests that have opt-outs) and filtered out the tests that fail due to unsupported intrinsics. SSE4.2, AVX2 and AVX-512 all pass.

Shnatsel:
I've researched whether the instruction set we chose is forward-compatible with Intel's upcoming AVX10. It is: according to the [Intel AVX10 architecture specification revision 7.0](https://cdrdv2-public.intel.com/919693/361050-007-intel-avx10.2-arch-spec.pdf), all AVX10 CPUs include the AVX-512 features from Ice Lake (our target) as well as Sapphire Rapids (higher than our target but doesn't add anything particularly useful).

Shnatsel:
I've run Vello benchmarks on Zen4, which doesn't even have native 512-bit vectors, and it slashes the end-to-end rendering benchmarks by about 15%!

<details><summary>Full benchmark run</summary>
<p>

```
$ cargo bench --bench main render_strips/ -- --load-baseline=avx512-8442ef44 --baseline=main-8442ef44
    Finished `bench` profile [optimized] target(s) in 0.09s
     Running benches/main.rs (/home/shnatsel/Code/vello/target/release/deps/main-efa34b2edd34e856)
render_strips/Ghostscript_Tiger_simd
                        time:   [125.64 µs 125.77 µs 125.94 µs]
                        change: [-16.343% -16.180% -16.011%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 50 measurements (18.00%)
  3 (6.00%) high mild
  6 (12.00%) high severe
render_strips/coat_of_arms_simd
                        time:   [1.3414 ms 1.3442 ms 1.3470 ms]
                        change: [-14.185% -14.081% -13.961%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 50 measurements (10.00%)
  5 (10.00%) high mild
render_strips/Saimaa_Canal_(map)_simd
                        time:   [3.6695 ms 3.6715 ms 3.6744 ms]
                        change: [-11.730% -11.510% -11.303%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 50 measurements (20.00%)
  2 (4.00%) high mild
  8 (16.00%) high severe
render_strips/heraldry_simd
                        time:   [883.81 µs 883.88 µs 883.97 µs]
                        change: [-15.411% -15.239% -15.078%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 50 measurements (14.00%)
  1 (2.00%) high mild
  6 (12.00%) high severe
```

</p>
</details>

check_targets.sh
```diff
@@ -15,6 +15,8 @@ cargo check -p fearless_simd --target aarch64-linux-android  --features force_su
 cargo check -p fearless_simd --target aarch64-linux-android
 
 # x86_64, at all supported static SIMD levels.
+RUSTFLAGS=-Ctarget-cpu=icelake-server cargo check -p fearless_simd --target x86_64-unknown-linux-gnu
+RUSTFLAGS=-Ctarget-cpu=icelake-server cargo check -p fearless_simd --target x86_64-unknown-linux-gnu --features force_support_fallback
```
LaurenzV, line 19: Why not just set the AVX512 feature flags here, like below? Also, do we need to update the commands below to activate all feature flags that were added to SSE4.2/AVX2 a while ago?

Cargo.toml
```diff
@@ -44,10 +44,11 @@ clippy.collection_is_never_read = "warn"
 clippy.default_trait_access = "warn"
 clippy.dbg_macro = "warn"
 clippy.debug_assert_with_mut_call = "warn"
+clippy.disallowed_methods = "deny"
 clippy.doc_markdown = "warn"
 clippy.fn_to_numeric_cast_any = "warn"
 clippy.infinite_loop = "warn"
-clippy.large_stack_arrays = "warn"
+clippy.large_stack_arrays = "allow"             # appears to be buggy as of 1.93, fixed in 1.95. TODO: re-enable
```
LaurenzV, line 51: Why would changing the MSRV from 1.88 to 1.89 impact this then?

fearless_simd_tests/tests/mod.rs
```diff
@@ -5,13 +5,53 @@
...
 use fearless_simd::*;
 use fearless_simd_dev_macros::simd_test;
 
 mod harness;
+#[cfg(not(miri))] // too slow
 mod soundness;
 
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+fn x86_detects_icelake_avx512() -> bool {
+    std::arch::is_x86_feature_detected!("adx")
```
LaurenzV: What about f16c, which we use for AVX2?

fearless_simd_tests/tests/mod.rs
```diff
@@ -5,13 +5,53 @@
...
 
 use fearless_simd::*;
 use fearless_simd_dev_macros::simd_test;
 
 mod harness;
+#[cfg(not(miri))] // too slow
 mod soundness;
 
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+fn x86_detects_icelake_avx512() -> bool {
```
LaurenzV: It would also be good to add a comment how this list was derived. I presume using `rustc --print=cfg --target x86_64-unknown-linux-gnu -C target-cpu=icelake-server`?

fearless_simd_tests/tests/mod.rs
```diff
@@ -62,6 +102,53 @@ fn supports_highest_level() {
...
+    );
+    assert!(
+        level.as_sse4_2().is_some(),
+        "AVX-512 should downgrade to an SSE4.2 proof"
+    );
+}
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[test]
+fn avx512_masks_are_compact() {
```
LaurenzV: Up to you, but this seems a bit superfluous to test.

fearless_simd_tests/tests/harness/lm_generated/mod_256.rs
```diff
@@ -315,6 +377,92 @@ fn trunc_f32x8_special_values<S: Simd>(simd: S) {
...
+        ],
+    );
+    assert_eq!(
+        *a.to_int_precise::<i32x8<_>>(),
+        [-10, 0, i32::MAX, i32::MIN, i32::MAX, i32::MIN, 42, 0]
+    );
+}
+
+#[simd_test]
+fn cvt_u32_f32x8_rounding<S: Simd>(simd: S) {
```
LaurenzV, line 423: COuld we also add a i32 one with rounding, checking rounding across both positive and negative?

fearless_simd_tests/tests/harness/lm_generated/mask_roundtrip_x86.rs
LaurenzV, line 1: I'm not super convinced that this module carries its weight, could you maybe elaborate on why you added this? It seems like `mask_roundtrip` should be enough to test that all behavior from the public API is correct and consistent across different backends, why do we need this module in addition?

check_targets.sh
LaurenzV: I guess it makes sense to keep it shorter, but the invocations below probably need to be updated (in a follow-up), no? Since they are missing the other target features we require. Or am I missing something?

LaurenzV:
(Sorry, meant to push to my private branch 😅 )

Shnatsel:
No worries, I have a local backup. I'm glad you're looking into this!

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -597,8 +716,218 @@ fn signed_literal(value: u64, bits: u32) -> TokenStream {
...
+        "simd_eq" => "cmpeq",
+        "simd_lt" => "cmplt",
+        "simd_le" => "cmple",
+        "simd_ge" => "cmpge",
+        "simd_gt" => "cmpgt",
+        _ => unreachable!(),
+    }
+}
+
+fn avx512_float_compare_predicate(method: &str) -> i32 {
```
LaurenzV, line 790: Could we maybe add a link to where those are defined?

fearless_simd_tests/tests/harness/lm_generated/extended_512.rs
LaurenzV, line 1: Should we have some more tests for `simd_le`/`simd_ge`,`select`, and in general some bit-mask ops? As far as I can tell those aren't really represented here right?

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -1204,12 +1984,30 @@ impl X86 {
                 |token| quote! { #set_intrinsic(a.into(), b.into()).simd_into(#token) },
             )
         } else {
-            let method_sig = op.simd_trait_method_sig(vec_ty);
             generic_block_combine(method_sig, combined_ty, self.max_block_size())
         }
     }
 
     pub(crate) fn handle_zip(&self, op: Op, vec_ty: &VecType, select_low: bool) -> TokenStream {
+        if *self == Self::Avx512 && vec_ty.scalar != ScalarType::Mask && vec_ty.n_bits() >= 256 {
```
LaurenzV, line 1994: Can this method even be called with masks? If not it seems like the sceond condition can just be omitted.

LaurenzV: Same also for interleave/deinterleave etc. Some other positions as well.

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -1527,6 +2473,116 @@ impl X86 {
...
+            });
+        }
+
+        if *self == Self::Avx512
+            && matches!(vec_ty.n_bits(), 128 | 256)
+            && vec_ty.scalar == ScalarType::Unsigned
+            && target_scalar == ScalarType::Float
+            && vec_ty.scalar_bits == 32
+        {
+            // We cannot emit the intrinsics for the conversion instructions
```
LaurenzV, line 2554: Let's orefix with a TODO then so we don't forget aout this.

fearless_simd/src/kernel_macros.rs
```diff
@@ -153,13 +153,27 @@ macro_rules! __fearless_simd_kernel_dispatch {
...
 
+    (
+        Avx512,
+        $($body:tt)*
+    ) => {
+        $crate::__fearless_simd_kernel_impl! {
+            @cfg any(target_arch = "x86", target_arch = "x86_64");
+            @token_ty $crate::Avx512;
+            @kernel_attrs #[target_feature(
+                enable = "adx,aes,avx512bitalg,avx512bw,avx512cd,avx512dq,avx512f,avx512ifma,avx512vbmi,avx512vbmi2,avx512vl,avx512vnni,avx512vpopcntdq,bmi1,bmi2,cmpxchg16b,fma,gfni,lzcnt,movbe,pclmulqdq,popcnt,rdrand,rdseed,sha,vaes,vpclmulqdq,xsave,xsavec,xsaveopt,xsaves"
```
LaurenzV, line 164: I really wish there were a more readable and easier-verifiable way for this 😓

fearless_simd_gen/src/generic.rs
```diff
@@ -375,10 +378,23 @@ pub(crate) fn generic_from_array(
     } else {
         quote! { val }
     };
-
     // There are architecture-specific "load" intrinsics, but they can actually be *worse* for performance. If they
     // lower to LLVM intrinsics, they will likely not be optimized until much later in the pipeline (if at all),
     // resulting in substantially worse codegen. See https://github.com/linebender/fearless_simd/pull/185.
+    //
+    // Safety: The native vector type backing any implementation will be:
```
LaurenzV: Isn't this the wrong place for a safety comment? There isn't actually any unsafe here, shouldn't this be in the transmute module (and I think we already have a similar comment there).

fearless_simd/src/generated/avx2.rs
```diff
@@ -11457,16 +11589,15 @@ impl<S: Simd> From<u8x32<S>> for __m256i {
 impl<S: Simd> SimdFrom<__m256i, S> for mask8x32<S> {
     #[inline(always)]
     fn simd_from(simd: S, arch: __m256i) -> Self {
-        Self {
-            val: crate::transmute::checked_transmute_copy(&arch),
-            simd,
-        }
+        let lanes: [i8; 32usize] = crate::transmute::checked_transmute_copy(&arch);
```
LaurenzV, line 11592: In the future, could this be avoided by specializing the `SimdFrom` impls for specific backends instad of making them generic over `Simd`?

LaurenzV (commented):
So, disclaimer, I have not actually tried to deeply understand how most of the more complex operations are implemented, I'm relying on my trust and the extensive tests here. 🙏 I don't have the time to try to validate all of this manually.

I did try to read through all code though, so just some comments here and there. For me, it's just important that we land some form of https://github.com/linebender/fearless_simd/pull/228 before making a new release.

Apart from that, some other things:
- I think the mk_86 code in `fearless_simd_gen` is getting pretty convoluted with all of the special casing for AVX512... I'm wondering if there is room for improvement in the future, but no idea.
- I'm also wondering whether in the future there is some way of having all of the required ffeatures be auto-generated so they only need to be defined once in fearless_simd_gen, but also something for the future not now.

check_targets.sh
Shnatsel: oh yeah I missed that script, it does need updating for v2/v3 targets, good call

fearless_simd_tests/tests/mod.rs
Shnatsel: It is implied by `avx512f`, you can verify it by running:

`rustc --print=cfg --target x86_64-unknown-linux-gnu -C target-feature=+avx512f`

This will print `target_feature="f16c"` among other things.

Cargo.toml
Shnatsel: I believe Clippy is run on MSRV instead of latest stable on CI.

We could probably switch it to latest stable now that it respects the crate MSRV, but that would randomly cause CI to fail on main because of Clippy adding new lints in later releases, and dealing with that is rather miserable in my experience.

fearless_simd/src/kernel_macros.rs
Shnatsel: I never managed to come up with one. We can't use variables here, and a declarative macro doesn't help either because we write these out in different contexts in slightly different ways, so even the strings we insert aren't the same.

fearless_simd/src/generated/avx2.rs
Shnatsel: I guess? I didn't see the point on complicating the generator further for the sake of this, since it's just two transmutes in a row anyway and optimizes into a by-value transmute.

Shnatsel:
> I think the mk_86 code in fearless_simd_gen is getting pretty convoluted with all of the special casing for AVX512... I'm wondering if there is room for improvement in the future, but no idea.

I agree. However, much of it is genuinely shared with the other levels, e.g. all the basic math operations emitting mostly the same intrinsics with 256 swapped for 512, so I'm not quite sure what the right cut points would be.

fearless_simd_tests/tests/harness/lm_generated/mask_roundtrip_x86.rs
Shnatsel: The complexity stems from AVX-512 having a different mask representation. I wanted to verify that the implementation of mask conversion into native x86 types worked correctly while I was fiddling with the implementation. It went through several iterations and I wanted to keep it covered by tests while I was experimenting.

Other platforms do not need the same tests because they don't have different mask representation depending on the simd level.

fearless_simd_tests/tests/harness/lm_generated/extended_512.rs
Shnatsel: Those are covered in fearless_simd_tests/tests/harness/lm_generated/mod_512.rs

The tests are getting out of hand, with 512-bit tests spread across 3 different files. They could use a refactor.

fearless_simd_gen/src/mk_x86.rs
Shnatsel: AFAIK it's not called right now but it could be in the future, so I don't see a reason to remove this guard.

fearless_simd_gen/src/generic.rs
Shnatsel: This comment is outdated, left over from before I made `checked_transmute_copy` fully safe. I'll change it.

Shnatsel:
I think I've addressed everything, so please take another look. Thanks again!

skewballfox:
> > I think the mk_86 code in fearless_simd_gen is getting pretty convoluted with all of the special casing for AVX512... I'm wondering if there is room for improvement in the future, but no idea.
> 
> I agree. However, much of it is genuinely shared with the other levels, e.g. all the basic math operations emitting mostly the same intrinsics with 256 swapped for 512, so I'm not quite sure what the right cut points would be.

just observing. If the test can be structured in a way to have the same inputs and outputs, you could alias the simd type and some const and include the same module twice. I recently tried this when I was trying to test api compability between two structs for different representations of complex tensors
```rust
#[allow(unused)]
macro_rules! gen_tests {
    ($variant:ident, $($ty:ty),*) => {
        $(
            paste::paste! {
                mod [<$variant:snake>] {

                    mod basic {
                        type TestTensor<const D: usize> = $ty;
                        include!("complex/basic.rs");
                    }
                    mod numeric {
                        type TestTensor<const D: usize> = $ty;
                        include!("complex/numeric.rs");
                    }
                    mod ops {
                        type TestTensor<const D: usize> = $ty;
                        include!("complex/ops.rs");
                    }
                }
            }
        )*
    };
}

//gen_tests!(split, burn_tensor::SplitTensor<D,burn_tensor::Complex> );
#[cfg(feature = "flex")]
gen_tests!(interleaved, burn_tensor::Tensor<D,burn_tensor::Complex>);
```

fearless_simd_tests/tests/harness/lm_generated/extended_512.rs
Shnatsel: I've confirmed that `simd_eq`, `simd_lt`, `simd_le`, `simd_ge`, and `simd_gt` are all covered for all widths. Is there anything in particular you think is missing?

LaurenzV: RIght, I guess I missed thsoe!

fearless_simd_gen/src/mk_x86.rs
LaurenzV: Right, in the end we'll end up in the unreachable section anyway.

fearless_simd/src/generated/avx2.rs
LaurenzV: > It occasionally adds an intermediate array but it gets optimized out in practice.

My thinking was more to ensure that this never regresses, but yeah it's not high priority.

LaurenzV (approved):
LGTM, however, I'll leave it up to you to decide but I'd still suggest considering removing the `mask_roundtrip_x86` tests. It's great that they helped during development, but in the end I think they test something we don't really care about? What we care about is that the public API of masks works as expected and iyields consistent results with other backends, which we already have plenty tests for.

Basically, I feel like (especially with our LLM-generated tests) we already have very good test coverage and already a lot of boilerplate code, so I'd like to avoid unnecessarily bloating our already very big test suite. 🙂

Shnatsel:
For AVX-512 the conversion is genuinely non-trivial. It is currently implemented via a roundtrip through `[T; N]` but it wasn't always like that and is not guaranteed to remain like that.

I do have gripes with the test suite, but I do believe testing these codepaths provides genuine value, especially with `kernel!` in the mix where you might reasonably want to turn a mask into a native type and back.

Shnatsel:
Thanks for the review! I'm excited to get this over the finish line!

PR #232, opened 2026-05-25
Safe transmute_copy()

Shnatsel:
"we have `bytemuck` at home"

These are the `transmute_copy()` safety improvements from #231 taken one step further: on top of checking the sizes match at compile time and that both types are `Copy`, we also add a marker trait to types that are safe to transmute and require it in `checked_transmute_copy()`, making that function fully safe to call.

The design follows the `bytemuck` crate closely. We could just use `bytemuck`, but it would pull in `syn` and other proc macro machinery, and we would still have to write the `impl_aligned_simd_pod!` macro because bytemuck rightfully refuses to derive `Pod` on generic wrappers like `Aligned128<T>` because they may introduce padding depending on the `T`. So the amount of code it saves us is minimal.

Rebasing #231 on top of this would be hellish, but a merge commit probably wouldn't be too bad.

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -154,6 +154,7 @@ mod kernel_macros;
 mod macros;
 mod support;
 mod traits;
+mod transmute;
 
 pub use generated::*;
 pub use traits::*;
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -131,6 +131,7 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 // These lints shouldn't apply to examples.
 #![warn(clippy::print_stdout, clippy::print_stderr)]
+#![cfg_attr(not(test), deny(clippy::disallowed_methods))]
 // Targeting e.g. 32-bit means structs containing usize can give false positives for 64-bit.
 #![cfg_attr(target_pointer_width = "64", warn(clippy::trivially_copy_pass_by_ref))]
 // END LINEBENDER LINT SET
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -131,10 +131,10 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 // These lints shouldn't apply to examples.
 #![warn(clippy::print_stdout, clippy::print_stderr)]
-#![cfg_attr(not(test), deny(clippy::disallowed_methods))]
 // Targeting e.g. 32-bit means structs containing usize can give false positives for 64-bit.
 #![cfg_attr(target_pointer_width = "64", warn(clippy::trivially_copy_pass_by_ref))]
 // END LINEBENDER LINT SET
+#![cfg_attr(not(test), deny(clippy::disallowed_methods))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![allow(non_camel_case_types, reason = "TODO")]
 #![expect(clippy::unused_unit, reason = "easier for code generation")]
--- a/fearless_simd/src/transmute.rs
+++ b/fearless_simd/src/transmute.rs
@@ -106,8 +106,16 @@ impl_aligned_simd_pod!(
     Aligned512<[u32; 16]>,
 );
 
+// the `const` is just to only use a single cfg annotation, nothing to do with const evalulation
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
 const _: () = {
+    // SAFETY: std docs clearly state:
+    // > The in-memory representation of this type is the same as the one of an equivalent array
+    // > (i.e. the in-memory order of elements is the same,
+    // and there is no padding between two consecutive elements);
+    // > however, the alignment is different and equal to the size of the type.
+    // at https://doc.rust-lang.org/stable/core/arch/x86_64/struct.__m256.html and the other structs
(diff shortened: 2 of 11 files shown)


LaurenzV:
Thanks, not expecting you to rebase, merge commit is fine!

fearless_simd/src/lib.rs
```diff
@@ -131,6 +131,7 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 // These lints shouldn't apply to examples.
 #![warn(clippy::print_stdout, clippy::print_stderr)]
+#![cfg_attr(not(test), deny(clippy::disallowed_methods))]
```
LaurenzV: Let's move this below `// END LINEBENDER LINT SET` since it's not part of the set to my understanding?

fearless_simd/src/transmute.rs
```diff
@@ -0,0 +1,235 @@
...
+#[allow(
+    unnameable_types,
+    reason = "This must be `pub` to avoid `private_bounds` warnings on the generated `ArchTypes` trait, but the containing module remains private"
+)]
+#[allow(
+    unreachable_pub,
+    reason = "This must be `pub` to avoid `private_bounds` warnings on the generated `ArchTypes` trait, but the containing module remains private"
+)]
+pub unsafe trait SimdPod: Copy {}
+
```
LaurenzV, line 48: Let's maybe also add a comment that links to https://docs.rs/bytemuck/latest/src/bytemuck/pod.rs.html.

fearless_simd/src/transmute.rs
```diff
@@ -0,0 +1,235 @@
...
+    Aligned512<[i8; 64]>,
+    Aligned512<[i16; 32]>,
+    Aligned512<[i32; 16]>,
+    Aligned512<[i64; 8]>,
+    Aligned512<[u8; 64]>,
+    Aligned512<[u16; 32]>,
+    Aligned512<[u32; 16]>,
+);
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
LaurenzV, line 111: Could you add a comment that the const is just to avoid repeating the cfg, that confused me a bit.

fearless_simd_gen/src/generic.rs
```diff
@@ -376,19 +376,7 @@ pub(crate) fn generic_from_array(
     // lower to LLVM intrinsics, they will likely not be optimized until much later in the pipeline (if at all),
     // resulting in substantially worse codegen. See https://github.com/linebender/fearless_simd/pull/185.
     let expr = quote! {
-        // Safety: The native vector type backing any implementation will be:
-        // - A `#[repr(simd)]` type, which has the same layout as an array of scalars
-        // - An array of `#[repr(simd)]` types
-        // - For AArch64 specifically, a `#[repr(C)]` tuple of `#[repr(simd)]` types
-        //
```
LaurenzV, line 383: I think it would be valuable to retain those comments for the `SimdPod` implementation of the types.

LaurenzV (commented):
LGTM, also nice so see the `unsafe` being more reduced to a single module.

We actually did use bytemuck initially, but got rid of it precisely to avoid pulling in bytemuck, with the argument that `fearless_simd` is low-level enough that we should just do our own thing here. It's a trade-off, but I think it's good to avoid extra dependencies here.

LaurenzV (approved):

fearless_simd/src/transmute.rs
Shnatsel: The link is already there in the doc comment, there's just a gap between it and the trait because of all the lint suppression noise

LaurenzV: Ah right

Shnatsel:
transient DNS resolution error, retrying

DJMcNab:
> We actually did use bytemuck initially, but got rid of it precisely to avoid pulling in bytemuck, with the argument that fearless_simd is low-level enough that we should just do our own thing here. It's a trade-off, but I think it's good to avoid extra dependencies here.

@LaurenzV can you link to where that decision was made? I can only find https://github.com/linebender/fearless_simd/pull/120, which removed the use of bytemuck only because we had the unsafe code for these cases already; in that thread, Valerie also would have been happy implementing this in terms of bytemuck. Alternatively, the mood in [#simd > ✔ Using &#96;bytemuck&#96;](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/.E2.9C.94.20Using.20.60bytemuck.60/with/539411260) also doesn't match your analysis. Fwiw, I'd be loosely in favour of moving to bytemuck completely, but equally I don't think it's urgent.

> We could just use bytemuck, but it would pull in syn and other proc macro machinery

I don't see how that follows, fwiw. We would only depend on bytemuck without using the `derive` feature, which then doesn't bring in Syn. That's the solution we use in Color and Peniko, so it would bring better uniformity with the rest of our crates.

LaurenzV:
> @LaurenzV can you link to where that decision was made? 

I'm not sure if it's written down anywhere, maybe I remember wrong. But I feel pretty confident that at some point I heard something along the lines of "fearless_simd is going to be a crate deep in the depenency tree, so we should avoid additional dependencies if possible" being said somewhere, but I can't back it up right now. I guess it wouldn't be a big deal, but I think for fearless_simd specifically it makes sense to just use our own solution here (unlike color or peniko), but if other's prefer bytemuck I'm not gonna block it!

PR #234, opened 2026-05-26
Replace by-value transmutes with checked_transmute_copy

Shnatsel:
A big safety win - a lot less bespoke `unsafe` to worry about. 

No difference in codegen in spot-checked assembly.

### Why not wrap `transmute` instead of using `transmute_copy`?

It's not possible to create a generic wrapper for by-value transmutes. rustc rejects it because it cannot prove they have the same sizes for all possible instantiations.

This is a genuine fundamental limitation. Even `bytemuck` implements its by-value transmute in terms of `transmute_copy`: https://github.com/Lokathor/bytemuck/blob/164cedda0eae131bc6cb67902599f4ec253642ca/src/lib.rs#L126

So we are on a well-trodden path for the optimizer, at least.

### DIFF
--- a/fearless_simd_gen/src/generic.rs
+++ b/fearless_simd_gen/src/generic.rs
@@ -213,11 +213,15 @@ pub(crate) fn generic_op(op: &Op, ty: &VecType) -> TokenStream {
             block_count,
         } => {
             let split_len = (block_size * block_count) as usize / (ty.scalar_bits * 2);
+            let ty_rust = ty.rust();
             quote! {
                 #method_sig {
                     let (chunks, _) = src.as_chunks::<#split_len>();
-                    unsafe {
-                        core::mem::transmute([self.#do_half(&chunks[0]), self.#do_half(&chunks[1])])
+                    let lo = self.#do_half(&chunks[0]);
+                    let hi = self.#do_half(&chunks[1]);
+                    #ty_rust {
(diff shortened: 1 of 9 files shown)


fearless_simd_gen/src/generic.rs
```diff
@@ -397,21 +401,30 @@ pub(crate) fn generic_as_array<T: ToTokens>(
...
-                // Not only do these all have the same layout as a flat array of the corresponding scalars, but they
-                // wrap primitives where all bit patterns are valid (ints and floats).
-                core::mem::transmute::<#ref_tok #native_ty, #ref_tok [#rust_scalar; #num_scalars]>(#ref_tok a.val.0)
+    match kind {
+        RefKind::Value => quote! {
+            #method_sig {
+                crate::transmute::checked_transmute_copy::<#native_ty, [#rust_scalar; #num_scalars]>(&a.val.0)
+            }
+        },
+        RefKind::Ref | RefKind::Mut => {
```
LaurenzV, line 413: Will we have something for transmuting (mutable) references as well in the future?

LaurenzV (approved):

fearless_simd_gen/src/generic.rs
LaurenzV: Ah, nevermind, there is a follow-up PR 😄

PR #235, opened 2026-05-26
Implement safe reference casts

Shnatsel:
Stacked on top of #234

reimplements bytemuck::cast/cast_mut but more clearly zero-cost and failing at build time rather than at runtime

then uses those functions to expunge the remaining unchecked `transmute` calls

and bans unchecked transmute via clippy

### DIFF
--- a/fearless_simd/src/transmute.rs
+++ b/fearless_simd/src/transmute.rs
@@ -245,6 +245,7 @@ pub(crate) fn checked_transmute_copy<Src: SimdPod, Dst: SimdPod>(src: &Src) -> D
     clippy::disallowed_methods,
     reason = "This is the central checked wrapper around transmute"
 )]
+#[allow(dead_code, reason = "Not all backends use this function")]
 pub(crate) fn checked_cast_ref<Src: SimdPod, Dst: SimdPod>(src: &Src) -> &Dst {
     const {
         assert!(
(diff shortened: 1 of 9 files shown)


fearless_simd/src/transmute.rs
```diff
@@ -237,3 +237,59 @@ pub(crate) fn checked_transmute_copy<Src: SimdPod, Dst: SimdPod>(src: &Src) -> D
...
+)]
+#[allow(dead_code, reason = "Not all backends use this function")]
+pub(crate) fn checked_cast_ref<Src: SimdPod, Dst: SimdPod>(src: &Src) -> &Dst {
+    const {
+        assert!(
+            size_of::<Src>() == size_of::<Dst>(),
+            "checked_cast_ref requires source and destination to have the same size"
+        );
+        assert!(
+            align_of::<Src>() >= align_of::<Dst>(),
```
LaurenzV, line 259: I guess might be obvious to others but it wasn't to me, this works because the alignment is always a power of 2, right? Otherwise, it wouldn't for example work if you could have an alignment of 5 and 4. Could you maybe add a comment linking to that https://doc.rust-lang.org/stable/reference/type-layout.html#size-and-alignment?

LaurenzV (approved):

fearless_simd/src/transmute.rs
Shnatsel: Added, thanks

DJMcNab:
I'm not sure what "more clearly zero cost" here means, but "failing at build time rather than compile time" is exactly bytemuck's [`must_cast`](https://docs.rs/bytemuck/latest/bytemuck/fn.must_cast.html)

PR #236, opened 2026-05-27
Remove unsafe from generated calls to intrinsics

Shnatsel:
Uses the `kernel!` macro added in https://github.com/linebender/fearless_simd/pull/214 to eliminate most of the `unsafe` in calls to intrinsics.

This is a proof-of-concept to evaluate the impact on performance, compile time, etc. It is not intended to be merged as-is.

### DIFF
--- a/fearless_simd_gen/src/mk_neon.rs
+++ b/fearless_simd_gen/src/mk_neon.rs
@@ -83,14 +83,14 @@ impl Level for Neon {
             OpSig::Splat => {
                 let expr = neon::expr(method, vec_ty, &[quote! { val }]);
                 let normalize_mask = integer_lane_mask_splat_arg(vec_ty);
-                quote! {
-                    #method_sig {
-                        unsafe {
-                            #normalize_mask
-                            #expr.simd_into(self)
-                        }
-                    }
-                }
+                op.simd_trait_kernel_method(
+                    self.token(),
+                    vec_ty,
+                    quote! {
+                        #normalize_mask
+                        #expr.simd_into(self)
+                    },
+                )
             }
             OpSig::Shift => {
                 let dup_type = vec_ty.cast(ScalarType::Int);
@@ -109,26 +109,14 @@ impl Level for Neon {
                     vec_ty,
                     &[quote! { a.into() }, quote! { #dup_intrinsic ( #shift ) }],
                 );
-                quote! {
-                    #method_sig {
-                        unsafe {
-                            #expr.simd_into(self)
-                        }
-                    }
-                }
+                op.simd_trait_kernel_method(self.token(), vec_ty, quote! { #expr.simd_into(self) })
             }
             OpSig::Unary => {
                 let args = [quote! { a.into() }];
 
                 let expr = neon::expr(method, vec_ty, &args);
 
-                quote! {
-                    #method_sig {
-                        unsafe {
-                            #expr.simd_into(self)
-                        }
-                    }
-                }
+                op.simd_trait_kernel_method(self.token(), vec_ty, quote! { #expr.simd_into(self) })
             }
             OpSig::LoadInterleaved {
                 block_size,
@@ -187,17 +175,17 @@ impl Level for Neon {
                     let id2 =
                         Ident::new(&format!("vcombine_{}", target_scalar_ty), Span::call_site());
 
-                    quote! {
-                        #method_sig {
-                            unsafe {
-                                let converted: #arch = a.into();
-                                let low = #id1(converted.0);
-                                let high = #id1(converted.1);
+                    op.simd_trait_kernel_method(
+                        self.token(),
+                        vec_ty,
+                        quote! {
+                            let converted: #arch = a.into();
+                            let low = #id1(converted.0);
+                            let high = #id1(converted.1);
 
-                                #id2(low, high).simd_into(self)
-                            }
-                        }
-                    }
+                            #id2(low, high).simd_into(self)
+                        },
+                    )
                 } else {
                     let arch = self.arch_ty(&target_ty);
                     let id1 = Ident::new(&format!("vmovl_{}", vec_scalar_ty), Span::call_site());
@@ -205,16 +193,16 @@ impl Level for Neon {
                     let id3 =
                         Ident::new(&format!("vget_high_{}", vec_scalar_ty), Span::call_site());
 
-                    quote! {
-                        #method_sig {
-                            unsafe {
-                                let low = #id1(#id2(a.into()));
-                                let high = #id1(#id3(a.into()));
+                    op.simd_trait_kernel_method(
+                        self.token(),
+                        vec_ty,
+                        quote! {
+                            let low = #id1(#id2(a.into()));
+                            let high = #id1(#id3(a.into()));
 
-                                #arch(low, high).simd_into(self)
-                            }
-                        }
-                    }
+                            #arch(low, high).simd_into(self)
+                        },
+                    )
                 }
             }
             OpSig::Binary => {
@@ -263,13 +251,7 @@ impl Level for Neon {
                     }
                 };
 
-                quote! {
-                    #method_sig {
-                        unsafe {
-                            #expr
-                        }
-                    }
-                }
+                op.simd_trait_kernel_method(self.token(), vec_ty, expr)
             }
             OpSig::Ternary => {
                 let args = match method {
@@ -291,13 +273,7 @@ impl Level for Neon {
                     let neg = simple_intrinsic("vneg", vec_ty);
                     expr = quote! { #neg(#expr) };
                 }
-                quote! {
-                    #method_sig {
-                        unsafe {
-                            #expr.simd_into(self)
-                        }
-                    }
-                }
+                op.simd_trait_kernel_method(self.token(), vec_ty, quote! { #expr.simd_into(self) })
             }
             OpSig::Compare => {
                 let args = [quote! { a.into() }, quote! { b.into() }];
@@ -306,13 +282,11 @@ impl Level for Neon {
                 let scalar_bits = vec_ty.scalar_bits;
                 let reinterpret_str = format!("vreinterpret{opt_q}_s{scalar_bits}_u{scalar_bits}");
                 let reinterpret = Ident::new(&reinterpret_str, Span::call_site());
-                quote! {
-                    #method_sig {
-                        unsafe {
-                            #reinterpret(#expr).simd_into(self)
-                        }
-                    }
-                }
+                op.simd_trait_kernel_method(
+                    self.token(),
+                    vec_ty,
+                    quote! { #reinterpret(#expr).simd_into(self) },
+                )
             }
             OpSig::Select => {
                 let opt_q = neon::opt_q(vec_ty);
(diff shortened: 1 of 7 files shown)


Shnatsel:
The initial assembly comparison on the fearless_simd_test binary is promising. 

The default multi-CGU build is too noisy to evaluate on size, but `-C codegen-units=1` there is very little change: the binary is only 16 bytes larger, the total number of `call` instructions is identical before and after.

Both default and `-C codegen-units=1` builds do not contain any `kernel` symbols.

DJMcNab:
Just to clarify my understanding, this should fix #124, right?

Shnatsel:
Yes, indeed. It would make the compiler verify it for us.

Shnatsel:
vello benchmarks (vello commit eae31cf ) this branch vs main on commit a01002e7

[safe_kernels_poc_bench_x86_zen4.txt](https://github.com/user-attachments/files/28430041/safe_kernels_poc_bench_x86_zen4.txt)

[safe_kernels_poc_bench_aarch64_m4.txt](https://github.com/user-attachments/files/28430055/safe_kernels_poc_bench_aarch64_m4.txt)

mostly noise-ish (some improvements and regressions on of equal magnitude) but a handful of large regressions stands out

LaurenzV:
That’s sad, any idea what’s going on?

Shnatsel:
I dug into the images/overlapping regression, which is quite severe (takes 2.5x longer).

**TL;DR:** It's a code layout change that hits a bad case in my CPU's frontend by accident. Layout varies per binary so it may not reproduce outside the benchmarking binary, and appear or disappear due to completely unrelated changes in other parts of the codebase.

There is no failed inlining or anything obvious like that. All the code still ends up inlined into `vectorize_avx2`. The assembly differences are also minimal.

`perf stat` shows one metric jumping up: `resyncs_or_nc_redirects` jumps from 1.7M to 164M.

It seems it's just a code layout change causing pathological frontend behavior. LLVM has, entirely by accident, changed the code layout in such a way that it causes op cache thrashing in the frontend.

I don't have the expertise to analyze this deeper myself. I asked codex to research it, this is what it came up with after a bit of back and forth:

<details><summary>Codex explanation</summary>
<p>

The important distinction is: **we did not make a tail instruction slow. We changed where the hot loop lands in the frontend’s fetch/op-cache geometry.**

The concrete assembly-level change I can point to is layout/packing:

```text
current:
  loop head: 0x4d0940  ; 64-byte aligned
  backedge:  0x4d0f42  ; offset 0x02 inside a 64-byte block

baseline:
  loop head: 0x4d8360  ; offset 0x20 inside a 64-byte block
  backedge:  0x4d8977  ; offset 0x37 inside a 64-byte block
```

The loop body is almost the same code, but the safety-branch codegen is about `0x15` bytes shorter in this specialization and has slightly different register allocation/scheduling. That small shift changes how the ~1.5 KiB hot loop is split across fetch blocks, op-cache entries, and predictor/redirect metadata.

Why would that restart over and over? Because every 16-byte output chunk hits the same taken backedge. If the target/body layout lands in a bad Zen 4 frontend case, then each iteration repeatedly asks the frontend to fetch from a stream it is not serving cleanly from the op cache. The counters match that:

```text
op-cache misses:          470M current vs 152M baseline
resyncs_or_nc_redirects:  164M current vs 1.7M baseline
branch redirects:         basically unchanged
branch misses:            not the story
```

So the plausible mechanism is:

1. The backedge retires.
2. The frontend redirects to the loop head.
3. In current layout, that redirect/op-cache lookup/fetch path repeatedly fails or needs resync.
4. The core sits underfed.
5. `perf` samples land on the tail/backedge because those are the nearby retired instructions.

What “caused it” in the assembly is not one bad instruction; it is the changed **code position and byte layout** of the hot loop. The `kernel!` wrapper change in `fearless_simd` perturbed inlining/codegen enough to move instructions and shrink/repack the loop, and Zen 4 is apparently landing on a pathological frontend case.

I would treat this as a strong hypothesis, not a fully proven root cause, until a no-op layout perturbation confirms it. The clean proof would be: add harmless padding/alignment that changes only addresses, not semantics, and see whether `resyncs_or_nc_redirects` and runtime swing back.

</p>
</details> 

`RUSTFLAGS='-Cllvm-args=-x86-align-branch-boundary=64 -Cllvm-args=-x86-align-branch=fused+jcc+jmp -Cllvm-args=-x86-pad-max-prefix-size=5'` which purely changes padding does reduce the op cache misses from 470.7M to 183.6M, and `resyncs_or_nc_redirects` from 163.7M to 105.0M, which supports codex's hypothesis.

Using `--profile=instrument` that adds debug info also cuts the regression in half. So it's clearly layout-sensitive.

`-C target-cpu=znver3` (last AMD with only AVX2, without AVX-512) fixes the regression completely. `-C target-cpu=znver4` improves further, but that enables AVX-512.

**Conclusion:** I think this is worth pursuing despite the handful of large regressions, since they may be isolated to the benchmarking binary, and are sensitive to unrelated changes to the code so they will be broken by something later anyway.

Shnatsel:
on https://github.com/smu160/PhastFT on Zen4 this also causes sharp but very isolated regressions, on top of some minor (<=10%) gains/losses.

In parallel mode, the sizes f32/8388608 and f64/4194304 completely collapse:

```
c2c_forward_f32/PhastFT DIT/8388608
                        time:   [44.797 ms 45.453 ms 46.209 ms]
                        thrpt:  [181.54 Melem/s 184.56 Melem/s 187.26 Melem/s]
                        thrpt:  [1.3526 GiB/s 1.3751 GiB/s 1.3952 GiB/s]
                 change:
                        time:   [+101.44% +104.72% +108.40%] (p = 0.00 < 0.05)
                        thrpt:  [−52.015% −51.153% −50.357%]
                        Performance has regressed.

c2c_inverse_f32/PhastFT DIT/8388608
                        time:   [61.591 ms 62.571 ms 63.539 ms]
                        thrpt:  [132.02 Melem/s 134.06 Melem/s 136.20 Melem/s]
                        thrpt:  [1007.3 MiB/s 1022.8 MiB/s 1.0148 GiB/s]
                 change:
                        time:   [+153.45% +158.45% +162.77%] (p = 0.00 < 0.05)
                        thrpt:  [−61.944% −61.308% −60.544%]
                        Performance has regressed.

c2c_forward_f64/PhastFT DIT/4194304
                        time:   [44.631 ms 45.638 ms 46.654 ms]
                        thrpt:  [89.903 Melem/s 91.903 Melem/s 93.977 Melem/s]
                        thrpt:  [1.3397 GiB/s 1.3695 GiB/s 1.4004 GiB/s]
                 change:
                        time:   [+97.208% +102.22% +107.59%] (p = 0.00 < 0.05)
                        thrpt:  [−51.828% −50.549% −49.292%]
                        Performance has regressed.
```

It is related specifically to the in-memory size, not the number of elements, since the f64 test at half the size collapses also. But that only happens on that size and only in multi-threaded mode.

Both single-threaded and multi-threaded modes also regress on the size just above it. But the size just above is completely unaffected in both, and all other sizes are unaffected in single-threaded.

Single-threaded:

```
c2c_forward_f32/PhastFT DIT/16777216
                        time:   [156.82 ms 157.70 ms 158.64 ms]
                        thrpt:  [105.76 Melem/s 106.39 Melem/s 106.98 Melem/s]
                        thrpt:  [806.87 MiB/s 811.66 MiB/s 816.22 MiB/s]
                 change:
                        time:   [+30.549% +31.228% +31.907%] (p = 0.00 < 0.05)
                        thrpt:  [−24.189% −23.797% −23.400%]
                        Performance has regressed.

c2c_forward_f64/PhastFT DIT/8388608
                        time:   [156.33 ms 157.21 ms 158.15 ms]
                        thrpt:  [53.043 Melem/s 53.361 Melem/s 53.661 Melem/s]
                        thrpt:  [809.38 MiB/s 814.22 MiB/s 818.81 MiB/s]
                 change:
                        time:   [+29.807% +30.500% +31.457%] (p = 0.00 < 0.05)
                        thrpt:  [−23.929% −23.371% −22.963%]
                        Performance has regressed.
```

Multi-threaded:

```
c2c_forward_f64/PhastFT DIT/8388608
                        time:   [63.184 ms 63.674 ms 64.209 ms]
                        thrpt:  [130.65 Melem/s 131.74 Melem/s 132.76 Melem/s]
                        thrpt:  [1.9468 GiB/s 1.9631 GiB/s 1.9783 GiB/s]
                 change:
                        time:   [+17.890% +18.958% +20.112%] (p = 0.00 < 0.05)
                        thrpt:  [−16.745% −15.937% −15.175%]
                        Performance has regressed.
```

Shnatsel:
I re-measured PhastFT and I can no longer reproduce the multi-threaded regression. Same setup, same compiler version, same Cargo.lock. FFTs are prone to running into [cache associativity issues](https://igoro.com/archive/gallery-of-processor-cache-effects/) due to their power-of-2 strides in memory access, so I suspect that run's in-memory data layout was just unlucky.

The single-treaded benchmark that appeared to regress also seems to be plainly unsable, varying a lot from run to run. I get 1.3 GiB/s, 1.1 GiB/s, or 0.9 GiB/s from run to run with no code changes.

That is to say, **PhastFT does not show any regressions that would block landing this.**

LaurenzV:
Will try it in vello on my x86 machine soon.

LaurenzV:
Unfortunately seeing pretty bad slowdowns on NEON. 😔 

image

LaurenzV:
Seems like codegen simply ends up worse because of all the inner functions. :( I guess they aren't optimized as they should be.

Shnatsel:
I don't think I've seen these `render_strips` benchmarks. How do I run them? Do they require some extra files to be downloaded?

LaurenzV:
You need to put any SVG into `sparse_strips/vello_bench/data`, then they will be registered automatically to the benchmarks. Here are `coat_of_arms` and `heraldry`, paris_30k is too big to upload unfortunately.

coat_of_arms
heraldry

Shnatsel:
That's fascinating. These benchmarks are completely unchanged on x86 (Zen4), even with https://commons.wikimedia.org/wiki/File:Saimaa_Canal_(map).svg added in for good measure.

But Aarch64 has NEON in the baseline instruction set and really doesn't need inlining for good codegen. So seeing it break but not x86 is unexpected.

I'll investigate, see what's up with that.

LaurenzV:
Are you able to reproduce it on your end?

LaurenzV:
But on the upside, I can confirm that I'm not really seeing any regressions for x86 AVX2. Some are a bit slower, some a bit faster but overall it's within ~10% range.

Shnatsel:
I am unable to reproduce the ARM regressions on my M4 McMini as of commit [8878920](https://github.com/linebender/fearless_simd/pull/236/commits/88789203933b82275351ca2cd1f5fcb6a942cf0b)

ghostscript tiger, coat_of_arms, heraldry and [Saimaa_Canal_(map).svg](https://commons.wikimedia.org/wiki/File:Saimaa_Canal_(map).svg) are all completely unchanged in the `render_strips` benchmarks.

I did wonder if it could be an incremental compilation bug/difference, but even after a `cargo clean` the numbers are completely stable (within 1%).

What hardware are you benchmarking this on? If it's a laptop, the difference could be simply due to thermal throttling if this branch is measured second.

I've been measuring on Vello from git on commit https://github.com/linebender/vello/commit/440b5a1c18ed9e9ac67931ed2a61f633a7572a05

I've upgraded to rustc 1.96 in case this interacts poorly with a rustc/LLVM version, but still no regressions.

Shnatsel:
I managed to reproduce the regression on rustc 1.90 to 1.92 inclusive (albeit ghostscript tiger improves by 7-10%).

Rust 1.93 and up do not show the same regression.

fearless_simd_gen/src/mk_neon.rs
```diff
@@ -109,26 +107,18 @@ impl Level for Neon {
...
-                        unsafe {
-                            #expr.simd_into(self)
-                        }
-                    }
-                }
+                self.kernel_method(op, vec_ty, |token| {
+                    quote! { #expr.simd_into(#token) }
+                })
             }
             OpSig::LoadInterleaved {
```
LaurenzV: What about load/store interleaved?

LaurenzV (approved):
Nice! 🎉🎉 After upgrading to Rust 1.96, I was not able to reproduce any noticeable slowdown (just some noise +/-10% sometimes) on any of the Vello benchmarks on NEON. I also played with the interactive Vello binary on WASM and didn't observe any slowdown. Given this, I think we can merge this without too much worries about regressions.

As mentioned in private message, it should be possible to make `to/from_bitmask` use this mechanism as well. Unfortunately, it doesn't work for NEON's `load/store_interleaved`.

PR #237, opened 2026-05-27
Replace remaining NEON unsafe loads with checked_transmute_copy

Shnatsel:
Follow-up to #233 which did this for x86

Leaves `vld4q_*` untouched because those are deinterleaving loads that genuinely need an intrinsic.

### DIFF
--- a/fearless_simd/src/generated/neon.rs
+++ b/fearless_simd/src/generated/neon.rs
@@ -780,7 +780,9 @@ impl Simd for Neon {
     #[inline(always)]
     fn from_bitmask_mask8x16(self, bits: u64) -> mask8x16<Self> {
         unsafe {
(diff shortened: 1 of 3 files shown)


LaurenzV (approved):
Maybe we should move them out of the unsafe block then?

LaurenzV:
Also, it doesn't seem like any WASM ones changed.

Shnatsel:
Ah, sorry, I was confused by the way github presented the diff. I thought the added impl was for wasm, but it wasn't.

I didn't move them out of the unsafe block to reduce conflicts with #236 but I can do that if you prefer.

LaurenzV:
Ah ok, no it's fine to leave them there! Just rename the PR maybe

PR #238, opened 2026-05-28
Make SIMD stores safe

Shnatsel:
Mirrors #232 but for stores

Performance is unchanged or potentially improved: the old codegen emitted `ptr::copy_nonoverlapping` which lowered into a `memcpy` in the LLVM IR, while this approach lowers into an unaligned store directly, which is the same thing as platform-specific SIMD instructions lower to. (Yes, platform-specific SIMD intrinsics lower into a generic op!)


LaurenzV (approved):
I haven't verified the memcpy claim but gonna trust you on this! The code makes sense to me.

PR #240, opened 2026-06-01
Rewrite kernel! macro examples in a way that is amenable to rustfmt

Shnatsel:
rustfmt treats macros called like `kernel!( ... )` as potentially containing Rust code and tries to format them, but `kernel!{ ... }`  is treated as opaque.

Use the auto-formattable form in all examples so that rustfmt doesn't break for crate users.


LaurenzV (approved):
TIL!

PR #241, opened 2026-06-12
Download Intel SDE from Github Releases mirror to fix CI

Shnatsel:
The old version is no longer available so switch to this new version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,8 +226,9 @@ jobs:
       # doc tests take a long time to run in the emulator and are already tested elsewhere, so exclude them with --lib --bins --tests
       # the rest of the arguments match the native test jobs
       CARGO_TEST_ARGS: --lib --bins --tests --workspace --locked --all-features
-      # SDE package name (for both the tarball and the folder within)
-      SDE_PKG: sde-external-10.8.0-2026-03-15-lin
+      # SDE package name without the platform suffix.
+      SDE_PKG: sde-external-10.8.0-2026-03-15
+      SDE_SHA256: 50b320cd226acef7a491f5b321fc1be3c3c7984f9e27a456e64894b5b0979dd3
     strategy:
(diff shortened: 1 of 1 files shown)


Shnatsel:
Ah, it's not that. I think Intel just always returns a zero-sized file for wget and curl downloads now. We'll need to find a way to store the binary on our CI instead.

Shnatsel:
Github Actions cache expires too quickly, so I've used Github Releases to store a mirror of Intel SDE. 
The CI job now verifies SHA-256 hash of the downloaded file to avoid supply chain attacks.

.github/workflows/ci.yml
```diff
@@ -247,30 +248,36 @@ jobs:
           save-if: ${{ github.event_name != 'merge_group' }}
 
       - name: install Intel Software Development Emulator
-        run: curl "https://downloadmirror.intel.com/873619/${SDE_PKG}.tar.xz" | tar -Jx
+        run: |
+          curl -fsSL -o "${SDE_PKG}-lin.tar.xz" "https://github.com/Shnatsel/intel-sde-mirror/releases/download/sde-${SDE_PKG#sde-external-}/${SDE_PKG}-lin.tar.xz"
```
LaurenzV, line 252: What does `lin` stand for?

LaurenzV (approved):

.github/workflows/ci.yml
LaurenzV: Ah, Linux 😂 nvm

DJMcNab:
Is there an easy way for us to validate that the checksum corresponds with a release from Intel? That's relatively low-priority for me, as the CI isn't very privileged (afaik), but it would be best practise.

Shnatsel:
There is a .sig file so you could bring in a GPG verifier, store the public key in Github Actions job in base64 or something, and verify the signature.

PR #242, opened 2026-06-16
Add top 5 uncredited contributors to AUTHORS

Shnatsel:
The README says

> Please feel free to add your name to the [AUTHORS](https://github.com/linebender/fearless_simd/blob/main/AUTHORS) file in any substantive pull request.

But it seems everyone just ignored it.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,3 +6,8 @@
 # source control.
 Google LLC
 Raph Levien
+Laurenz Stampfl
+valadaptive
+Sergey "Shnatsel" Davidoff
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):

DJMcNab:
Our general process for this has been to ask people, rather than just doing this unilaterally. At least for myself, the primary reason for not being on here was that my contributions until the end of January were all owned by Google, and I don't think my contributions since reach the level.

We should revert this, and do it with proper consent.

PR #243, opened 2026-06-16
Document the motivation for custom transmute wrappers

Shnatsel:
They've grown to be more than just 'bytemuck at home'

### DIFF
--- a/fearless_simd/src/transmute.rs
+++ b/fearless_simd/src/transmute.rs
@@ -1,9 +1,9 @@
 // Copyright 2026 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! This all serves a small set of checked transmute and cast functions.
+//! Safe wrappers for SIMD loads/stores and reference casts.
 //!
-//! It fulfills the purpose of both `safe_unaligned_simd` and `bytemuck` crates.
+//! This fulfills the purpose of both `safe_unaligned_simd` and `bytemuck` crates.
 //! The implementation is bytemuck-like, but far smaller than either of those crates,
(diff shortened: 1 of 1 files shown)


fearless_simd/src/transmute.rs
```diff
@@ -1,10 +1,17 @@
 // Copyright 2026 the Fearless_SIMD Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! We have bytemuck at home
+//! This all serves a small set of checked transmute and cast functions.
```
LaurenzV: I think it would be better to keep a concise top-title (like "Transmuting between SIMD types and wrappers" and then start with the description in the next paragraph.

LaurenzV (approved):

fearless_simd/src/transmute.rs
```diff
@@ -1,10 +1,17 @@
...
+//! Safe wrappers for SIMD loads/stores and reference casts.
 //!
-//! This all serves a small set of checked transmute and cast functions;
-//! if we find this growing in complexity we should probably just use the real bytemuck
+//! This fulfills the purpose of both `safe_unaligned_simd` and `bytemuck` crates.
+//! The implementation is bytemuck-like, but far smaller than either of those crates,
+//! mostly by virtue of supporting less features (e.g. no by-value transmute).
+//!
+//! Unlike bytemuck, this verifies that sizes match at compile time,
+//! so we won't accidentally ship always-panicking code even if it's not covered by tests.
```
DJMcNab, line 11: As I noted in #232, with bytemuck's 'must' functionality, this isn't really the case.

PR #244, opened 2026-06-16
Remove `unsafe` from generated helpers and const-generic functions

### LINKED ISSUE #124: Make sure we're not using any unsupported intrinsics in the various x86 levels

Shnatsel:
Follow-up to https://github.com/linebender/fearless_simd/pull/236

Didn't even need to extend `kernel!` to support const generics in the end.

The only remaining `unsafe` is now in `load/store_interleaved` which can be eliminated on x86 and WASM but not NEON. That's addressed separately in https://github.com/linebender/fearless_simd/pull/245

Fixes #124

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -152,20 +152,19 @@ impl Simd for Avx2 {
     }
     #[inline(always)]
     fn slide_f32x4<const SHIFT: usize>(self, a: f32x4<Self>, b: f32x4<Self>) -> f32x4<Self> {
-        unsafe {
-            if SHIFT >= 4usize {
-                return b;
-            }
-            let result = dyn_alignr_128(
-                self.cvt_to_bytes_f32x4(b).val.0,
-                self.cvt_to_bytes_f32x4(a).val.0,
-                SHIFT * 4usize,
-            );
-            self.cvt_from_bytes_f32x4(u8x16 {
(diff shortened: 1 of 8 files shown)


DJMcNab (approved):
Reviewing with whitespace hidden helps a lot here. I can approve that this only removes `unsafe`s, without changing logic.

PR #245, opened 2026-06-16
Safer load/store_interleaved

Shnatsel:
Expunge unsafe code from load_interleaved and store_interleaved implementations on x86 and WASM.

NEON is implemented via a call to an intrinsic, so it can't be made any safer.

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -8676,7 +8676,7 @@ impl Simd for Avx2 {
                 );
             }
         );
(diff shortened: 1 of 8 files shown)


LaurenzV (approved):
LGTM, if you want to be really thorough it might be good to rerun the `pack_block` benchmark on Vello to see whether this causes any change on x86, since it uses those methods. But you don't have to, I'll redo the benchmarks again anyway before bumping fearless_simd in vello, so if it causes issue I'll notice it then. 😄

PR #246, opened 2026-06-17
Revert "Add top 5 uncredited contributors to AUTHORS (#242)"

Shnatsel:
This reverts commit 4f885d94ea27085339a94ff3dafe69aaea6d9f4b

as requested in https://github.com/linebender/fearless_simd/pull/242#issuecomment-4725028210

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,8 +6,3 @@
 # source control.
 Google LLC
 Raph Levien
-Laurenz Stampfl
-valadaptive
-Sergey "Shnatsel" Davidoff
-Benjamin Saunders
-Daniel McNab


LaurenzV (approved):

DJMcNab:
Thank you. Sorry to make this difficult - it's just a weird aspect of where the technical meets the social.

But basically I did get advice from Google that I shouldn't have my name here, as this is especially marking copyright ownership. The whole authors file thing is a weird holdover from when we had much greater Google involvement. Otheres would likely to have different reasons (the risk of deadnaming is another).

I really appreciate the urge to give people credit, but as I said in the other thread, my general position would be that people should have a say on whether they are added here.

PR #247, opened 2026-06-17
Better explain the motivation for bytemuck-at-home

Shnatsel:
I figured it's best to put it directly in the doc comment instead of explaining in a PR or on Zulip.

Thanks for the detailed reviews!


DJMcNab (approved):
/shrug

This isn't especially far up my priority list. I'd lean towards removing the first and last sentences here here, and just say that we've already done the work to validate this, but not enough to block on that.

PR #248, opened 2026-06-17
Populate changelog for the next release

Shnatsel:
Even if we don't ship soon, it's good to have this written down.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,23 @@ You can find its changes [documented below](#041-2026-05-16).
 
 This release has an [MSRV][] of 1.88.
 
+### Added
+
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):
I haven't double-checked all the links, but this matches my recollection of changes made. If you want I can make a new release in the evening.

Shnatsel:
Please publish `0.5.0-rc.1`, which is a pre-release. If everything is going smoothly, we can publish the real 0.5.0 in a few days.

LaurenzV:
Hmm I'm not sure, I don't think we have precedent for this. 😅 It should be fine to just do the release and do a patch release afterwards in case issues show up.

Shnatsel:
Well, I'm certainly not opposed to my work being shipped sooner! :sweat_smile:

PR #252, opened 2026-06-21
Update README

Shnatsel:
Remove "experimental" warning from the README, brag about our offerings, and add an example for every abstraction level.

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -11,7 +11,7 @@
 //!
 //! Zero dependencies, from-scratch build time under 1 second, safe public APIs, and [very little](https://gist.github.com/Shnatsel/61fc294987a1e051ce3835c97dc0fc19) `unsafe` under the hood.
 //!
-//! ## Automatic vectorization
+//! # Automatic vectorization
 //!
 //! Put the code to vectorize in an `#[inline(always)]` function generic over [`Simd`].
 //!
@@ -33,7 +33,7 @@
 //! assert_eq!(values, [2, 4, 6, 8, 10]);
 //! ```
 //!
-//! ## Portable SIMD
+//! # Portable SIMD
 //!
 //! Use the vector types for explicit lane-wise operations while staying generic over the SIMD level:
 //!
@@ -60,7 +60,7 @@
 //!
 //! You can also use fixed-size types such as [u32x8] instead of using the hardware's native SIMD width.
 //!
-//! ## Explicit intrinsics
+//! # Explicit intrinsics
 //!
 //! If you need access to raw intrinsics, [`kernel!`][kernel] creates a function where they can be called safely:
 //!
@@ -99,7 +99,7 @@
 //! intrinsics with the other approaches, using high-level code most of the time and dropping down to
 //! hardware-specific intrinsics only when necessary.
 //!
-//! ## Inlining
+//! # Inlining
 //!
 //! Fearless SIMD relies heavily on Rust's inlining support to create functions which have the given target features enabled.
 //!
@@ -112,7 +112,7 @@
 //! [The article describing the design](https://gist.github.com/Shnatsel/61fc294987a1e051ce3835c97dc0fc19#the-abi-would-like-a-word) covers why this is the
 //! case. There's also Q&A on [Zulip](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/inlining/with/546913433).
 //!
-//! ## Instruction set support
+//! # Instruction set support
 //!
 //! - x86/x86-64: [v2](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (SSE4.2), [v3](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (AVX2)
 //! - Aarch64: Baseline [NEON](https://en.wikipedia.org/wiki/Arm_architecture_family#Advanced_SIMD_(Neon))
@@ -120,7 +120,7 @@
 //!
 //! A scalar fallback is also provided for platforms, so your code still works even if SIMD is not available.
 //!
-//! ## WebAssembly
+//! # WebAssembly
 //!
 //! WASM SIMD doesn't have feature detection, and so you need to compile two versions of your bundle for WASM, one with SIMD and one without,
 //! then select the appropriate one for your user's browser. This can be done via [the `wasm-feature-detect`
@@ -140,7 +140,7 @@
 //! that builds it once with the `RUSTFLAGS` specified, and once without. [Cargo currently does not allow specifying compiler flags
 //! per-profile.](https://github.com/rust-lang/cargo/issues/10271)
 //!
-//! ### Relaxed SIMD
+//! ## Relaxed SIMD
 //!
 //! Fearless SIMD can make use of the [relaxed SIMD](https://github.com/WebAssembly/relaxed-simd/blob/main/proposals/relaxed-simd/Overview.md)
 //! WebAssembly instructions, if the requisite target feature is enabled. These instructions can return implementation-dependent results
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -179,6 +179,10 @@
 #![cfg_attr(not(test), deny(clippy::disallowed_methods))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![allow(non_camel_case_types, reason = "TODO")]
+#![expect(
+    clippy::doc_markdown,
+    reason = "No, Clippy, I'm not going to put 'x86/x86_64' in documentation into backticks."
+)]
 #![expect(clippy::unused_unit, reason = "easier for code generation")]
 #![no_std]
 
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -114,7 +114,7 @@
 //!
 //! ## Instruction set support
 //!
-//! - x86/x86_64: [v2](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (SSE4.2), [v3](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (AVX2)
+//! - x86/x86-64: [v2](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (SSE4.2), [v3](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (AVX2)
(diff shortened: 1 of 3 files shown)


fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
+//! Zero dependencies, from-scratch build time under 1 second, and >99.8% safe Rust under the hood!
 //!
-//! Additionally, there are types for packed vectors of a specific width and element type (such as [`f32x4`]).
-//! Fearless SIMD does not currently support vectors of less than 128 bits.
-//! These vector types implement some standard arithmetic traits (i.e. they can be added together using
-//! `+`, multiplied by a scalar using `*`, among others), which are implemented as efficiently
-//! as possible using SIMD instructions.
-//! These can be created in a SIMD context using the [`SimdFrom`] trait, or the
-//! [`from_slice`][SimdBase::from_slice] associated function.
+//! ## Automatic vectorization
```
DJMcNab: Nit:
```suggestion
//! # Automatic vectorization
```
Edit: Oh, you seem to have done this throughout the file. I'm interested to hear why this was needed!

fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
 //!
-//! To call a function with the best available target features and get the associated `Simd`
-//! implementation, use the [`dispatch`] macro:
+//! Put the code to vectorize in an `#[inline(always)]` function generic over [`Simd`].
+//!
+//! This will generate several implementations for different SIMD levels and select the best one at runtime:
```
DJMcNab, line 18: Personally, I'd mention `dispatch` here. Something like:

```suggestion
//! You can then call this function using the `dispatch` macro.
//! This will generate several implementations for different SIMD levels and select the best one at runtime:
```

fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
-//! If you are writing an application, you should create this once (using [`Level::new`]), and pass it to any function which wants to use SIMD.
-//! This type stores which instruction sets are available for the current process, which is used
-//! in the macro to dispatch to the most optimal variant of the supplied function for this process.
+//! If you need a specific intrinsic, [`kernel!()`][kernel] creates a safe wrapper from a SIMD token:
 //!
-//! # Inlining
+//! ```rust
+//! use fearless_simd::{prelude::*, Level, u32x4};
+//!
+//! fearless_simd::kernel!(
```
DJMcNab, line 70: It's a bit surprising to me that we don't need an explicit `#[cfg(target_arch = "aarch64")]` here.

fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
-//! 2) [`dispatch`] is used to invoke the given function with the target features associated with the supplied [`Level`].
-//! 3) The function or closure passed to [`dispatch`] should be `#[inline(always)]`.
-//!    The performance of the SIMD implementation may be poor if that isn't the case. See [the section on inlining for details](#inlining)
+//! ## Explicit intrinsics
 //!
-//! The first parameter to [`dispatch`] is the [`Level`].
-//! If you are writing an application, you should create this once (using [`Level::new`]), and pass it to any function which wants to use SIMD.
-//! This type stores which instruction sets are available for the current process, which is used
-//! in the macro to dispatch to the most optimal variant of the supplied function for this process.
+//! If you need a specific intrinsic, [`kernel!()`][kernel] creates a safe wrapper from a SIMD token:
```
DJMcNab: I would like for us to mention in prose somewhere that this allows using any safe intrinsic from `core::arch` without unsafe. That is, this sentence as written to me hints that you have a to create a wrapper for the intrinsic yourself.

fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
-//! for some train of thought explanation.
-//!
-//! <!--
-//! TODO: Also have concrete examples of each of these.
-//!
-//! TODO: This is a really subtle point, and we do need there to be a well-written explanation available.
-//! E.g. We might want names for these, e.g.:
-//!
-//! # Kernels vs not kernels
+//! [The article describing the design](https://shnatsel.medium.com/safe-simd-in-rust-even-on-the-inside-c6f1ff381828) covers why this is the
```
DJMcNab: Do we have an answer to the (account creation) paywall introduced by the use of medium? This is only based on the reddit comments, but it seems especially insidious as it is triggered semi-arbitrarily.

fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
-//! -->
+//! ## Instruction set support
 //!
-//! # Platform-specific intrinsics
+//! - x86/x86-64: [v2](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (SSE4.2), [v3](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) (AVX2)
+//! - Aarch64: Baseline [NEON](https://en.wikipedia.org/wiki/Arm_architecture_family#Advanced_SIMD_(Neon))
+//! - WebAssembly: [128-bit packed SIMD](https://github.com/WebAssembly/spec/blob/main/proposals/simd/SIMD.md), [relaxed SIMD](https://github.com/WebAssembly/relaxed-simd/blob/main/proposals/relaxed-simd/Overview.md)
 //!
-//! If the portable APIs are not enough, you can safely invoke platform-specific intrinsics via the [`kernel!()`][kernel] macro.
+//! A scalar fallback is also provided for platforms, so your code still works even if SIMD is not available.
```
DJMcNab, line 121: I don't think that this first clause is grammatical. Perhaps `all platforms` or `all operations`?

fearless_simd/src/lib.rs
```diff
@@ -4,81 +4,123 @@
...
+//! `fearless_simd` takes `unsafe` out of SIMD.
 //!
-//! Fearless SIMD exposes safe SIMD with ergonomic multi-versioning in Rust.
+//! No matter what level of abstraction you're after, be it autovectorization and multiversioning, or portable SIMD, or safe access to raw
+//! intrinsics and nothing more, `fearless_simd` has you covered!
 //!
-//! Fearless SIMD uses "marker values" which serve as proofs of which target features are available on the current CPU.
-//! These each implement the [`Simd`] trait, which exposes a core set of SIMD operations which are implemented as
-//! efficiently as possible on each target platform.
+//! Zero dependencies, from-scratch build time under 1 second, and >99.8% safe Rust under the hood!
```
DJMcNab: I don't think that the ">99.8% safe Rust" clause reads especially well. I think it's trying to express two things:
- That we're sound.
- That you can use the same abstractions to access the same functionality safely.

For the former, that's the default expectation for a Rust library, and so I wouldn't choose to use it as an "entry-point marketing" thing. I think a prose section about how we maintain soundness is valuable, but luckily that's already covered by your article.

For the latter, this is a valuable differentiator. I would phrase it as something like "choose your own level of abstraction" instead, but that is already kind of covered by the sentence beforehand anyway.

(As a note, this whole section isn't something I'd normally expect to find in Linebender's tone of voice, but I'm happy for this to be in your style).

DJMcNab (commented):
Generally I'm happy with this, but there are a few points from the review feedback I would like to re-validate before we land this.

Thanks!

fearless_simd/src/lib.rs
Shnatsel: First, it's the obviously correct heading structure given the level 1 header in the README.

Second, it looks a little nicer when rendered.

But it's not a hill I'm willing to die on so if you have strong opinions on header levels please go ahead and change them.

fearless_simd/src/lib.rs
Shnatsel: I think it's already covered in prose in the inlining section in detail. I don't think duplicating only a part of it here would be beneficial.

fearless_simd/src/lib.rs
Shnatsel: Good call, I'll rewrite that part.

fearless_simd/src/lib.rs
Shnatsel: What I am trying to convey is that very little `unsafe` is used internally, so the library is easy to audit, and doesn't have a whole lot that can go wrong.

This is a crucial differentiator because all other SIMD abstractions except `jxl-simd` have thousands of unsafe blocks.

fearless_simd/src/lib.rs
Shnatsel: `kernel!` takes care of it internally. That's a benefit of matching on hardcoded identifiers, as opposed to full-blown token types. The drawback is that kernel! doesn't accept aliases of `Avx2` etc.

fearless_simd/src/lib.rs
Shnatsel: I've changed it, please take another look.

fearless_simd/src/lib.rs
Shnatsel: Good catch, fixed!

fearless_simd/src/lib.rs
Shnatsel: I've mirrored the article on github gist and link to that instead.

fearless_simd/src/lib.rs
DJMcNab: Could you point specifically to this header? In my GitHub diff viewer, it seems to go from h1 straight to h3...

Cargo rdme adds a level to the headings, so as to allow both to be idiomatic.

fearless_simd/src/lib.rs
DJMcNab: Sure. But I'm saying that it's at least a little bit surprising as a user to declare a function which has no `cfg` block, but just isn't there on some platforms.

This isn't the place for this discussion though.

fearless_simd/src/lib.rs
DJMcNab: My point was more that the prose here to me reads that just writing the generic function is what makes the multiple implementations, which just isn't the case. But I don't care too much; this is still better than before.

fearless_simd/src/lib.rs
DJMcNab: It looks like you tried to fix it in 946abb0eed4ccf46720f870f2daa0e755098d471, but then undid that in a0366aa727f09be7acff4e1806883bfe75d15219?

DJMcNab (approved):
Barring the two blocking concerns (header levels and ungrammatical sentence), this looks good to me. Thanks!

I especially really like the idea of having examples for each level.

fearless_simd/src/lib.rs
Shnatsel: Oops, apparently so! Thanks for catching that!

PR #97, opened 2025-09-28
64 bit ints

dbuch:
I began drafting out a basic 64-bit integer support.
A few tests has been added to assist development a few more are likely to be added.

I didn't check-in the generated code from `cargo run -p fearless_simd_gen` yet since i think its easier to review w/o.

Whats next:

1. test on other architectures than x86
2. Iron out undetected mistakes
3. Add more tests

This is tested with `cargo test -p fearless_simd_tests` and `RUSTFLAGS=-Ctarget-feature=+simd128 wasm-pack test --headless --firefox` (note: i couldn't get --chrome to work on my)

Both passes with the additional tests after generated code has been updated.

PR #253, opened 2026-06-23
Add 64-bit integer vectors and operations on them

Shnatsel:
There are two unrelated changes:

1. The documentation on load/store_interleaved_128 is fixed to match the implementation, which in turn mirrors the NEON intrinsics
2. Some generic fallback methods are refactored to match the fallback's fully unrolled style as opposed to `array::from_fn()` calls

Supersedes #97


Shnatsel:
The documentation for load/store_interleaved_128 was misleading. Both formulations are valid for 32-bit elements but the 8- and 16-bit elements already behaved differently, following the NEON vld4/vst4 semantics rather than our documented semantics. This misled me into generalizing the op to 64-bit numbers incorrectly.

I've changed the implementation back to vld4/vst4 semantics in subsequent commits and updated documentation.

fearless_simd_gen/src/generic.rs
```diff
@@ -309,8 +312,78 @@ pub(crate) fn generic_op(op: &Op, ty: &VecType) -> TokenStream {
...
+        quote! { if a[#idx] #op b[#idx] { true_lane } else { false_lane } }
+    });
+
+    quote! {
+        let a: [#scalar; #len] = a.into();
+        let b: [#scalar; #len] = b.into();
+        let true_lane: #mask_scalar = !0;
+        let false_lane: #mask_scalar = 0;
+        let result: [#mask_scalar; #len] = #items;
+        result.simd_into(#simd)
```
Shnatsel: This basically duplicates the scalar fallback code but I didn't want to do a big refactoring here that would change the scalar fallback level.

But that refactoring might be worth it considering that https://github.com/linebender/fearless_simd/issues/256 also needs it.

LaurenzV:
I'm curious, do you think this will overall impact the compile time for the crate a lot, even if none of the 64-bit stuff is used? Have you done any measurements?

Shnatsel:
It really shouldn't. This is all generic code, so it is not actually instantiated and doesn't turn into MIR or LLVM IR until something actually calls it.

The downside of generics is that if we call the same function 5 times you get 5 different instantiations of it so 5x the IR for LLVM to chew through, but in our case we want all the intrinsics inlined anyway so this is unavoidable, generics or not.

Shnatsel:
Marking this as draft until it's updated to latest main. Some refactoring should also be possible now that the SSE2 PR has laid the groundwork for nicer scalar fallback.

fearless_simd_gen/src/generic.rs
Shnatsel: Done

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -1358,6 +1355,13 @@ impl X86 {
             });
         }
 
+        if vec_ty.scalar_bits == 64
+            && matches!(vec_ty.scalar, ScalarType::Int | ScalarType::Unsigned)
+            && method != "simd_eq"
+        {
+            return fallback_method(op, vec_ty);
+        }
+
```
Shnatsel, line 1364: I don't like that this fires only when it's not AVX-512 but this is not explicitly under an `else` or a non-AVX-512 branch. I want to refactor all the early returns out of the generator, but that's best left to a follow-up PR, I don't want to cram even more unrelated changes into this one.

Shnatsel:
This should now be ready for review.

fearless_simd_tests/tests/harness/int64.rs
LaurenzV, line 1: What about `rotate/shift_elements_left/right` and `swizzle`?

fearless_simd_gen/src/generic.rs
```diff
@@ -431,8 +436,9 @@ fn right_offset_arms(
         .collect()
 }
 
-pub(crate) fn scalar_binary(f: TokenStream) -> TokenStream {
-    quote! { core::array::from_fn(|i| #f(a[i], b[i])).simd_into(self) }
+pub(crate) fn unrolled_array(len: usize, item: impl FnMut(usize) -> TokenStream) -> TokenStream {
+    let items = (0..len).map(item).collect::<Vec<_>>();
+    quote! { [#(#items),*] }
```
LaurenzV, line 441: What's the motivation for this change? Did you verify whether this actually gives better assembly? Is that something that could be reverted for now to make the diff easier to read? Up to you.

fearless_simd/src/generated/simd_types.rs
```diff
@@ -392,8 +392,25 @@ impl<S: Simd> SimdBase<S> for i8x16<S> {
         block
     }
     #[inline(always)]
-    fn from_fn(simd: S, f: impl FnMut(usize) -> i8) -> Self {
-        simd.load_array_i8x16(core::array::from_fn(f))
+    fn from_fn(simd: S, mut f: impl FnMut(usize) -> i8) -> Self {
```
LaurenzV, line 395: Piggybacking on my other comment, does this really yield better results than before? Seems a bit suspicious. 😄

fearless_simd/src/generated/fallback.rs
LaurenzV, line 1: I think we need to use `wrapping_neg` for negation, no? Also maybe add a test with `i64::MIN` to ensure it works.

LaurenzV (approved):
Once again mostly relying on tests here, haven't tried to verify the actual logic of each operation. If it's possible, I do think it would be nice to split out the `from_fn` change, just to keep the diff clean (not like anyone else will ever read it, but still 🙂).

fearless_simd_tests/tests/harness/int64.rs
Shnatsel: that's covered in https://github.com/linebender/fearless_simd/pull/275/changes#diff-f8c0fbfc8f4c516a0abcc54128c32013ac7838e28285085fef7800fa7cce64c3

so not this PR but the follow-up that expands test coverage, stacked on top of this one

fearless_simd_gen/src/generic.rs
Shnatsel: I noticed that these fallbacks are not the same style as the actual `Fallback` implementation and went ahead and changed it, but I didn't verify that the fully unrolled implementations produce better assembly, I just assumed that the fallbacks are that way for a reason.

fearless_simd/src/generated/simd_types.rs
Shnatsel: There were some issues with it historically: https://github.com/rust-lang/rust/issues/108765

But I guess it doesn't matter as much today. There's all sorts of considerations like panic safety and such, but measuring this rigorously is quite difficult, since you can put arbitrary functions in there and the properties probably vary depending on what you put there.

fearless_simd/src/generated/fallback.rs
Shnatsel: Yes, it should. That's an issue with the current code, not restricted to 64-bit ints.

fearless_simd_gen/src/generic.rs
Shnatsel: I've checked it for shl op and it seems LLVM unrolls those loops anyway, so it thinks that's profitable but I don't want to bet on LLVM preserving this optimization in the future so I'll keep the manual unrolling.

As far as I can tell these two formulations emit literally identical assembly, at least for shl as the op. I was surprised and suspected an error in my setup, but I double-checked and the results appear to hold.

I've inspected the optimized LLVM IR before lowering into assembly and it seems it gets recognized as a vector operation in both cases.

```
%values = bitcast <2 x i64> %a to <16 x i8>
%shifts = bitcast <2 x i64> %b to <16 x i8>
%masked = and <16 x i8> %shifts, splat (i8 7)
%result = shl <16 x i8> %values, %masked
```

fearless_simd/src/generated/fallback.rs
Shnatsel: I've changed the scalar fallback behavior in this PR, and updated the tests in #275

PR #257, opened 2026-06-28
Merge LLM-generated 512-bit tests into a single file

Shnatsel:
This is a simple copy-paste job to make the tests a little less scattered.

There's certainly more to be done, but this looks like an uncontroversial starting point.


LaurenzV (approved):
I have not double checked that it's actually the same

PR #258, opened 2026-06-29
Multiversioning behind feature flags for SIMD levels

Shnatsel:
A different, better take on #228. It doesn't disable the tokens themselves and still lets you use `kernel!` with whatever features you please. The cargo features only control multiversioning via the `dispatch!` macro for functions generic over `Simd`.

~~TODO: better documentation, incl. on opting out specific functions instead of crate-wide, advice on features for library authors, advice on reducing binary size without degrading performance via codegen-units=1 and lto=true~~ Done

### DIFF
--- a/fearless_simd/examples/disable_avx2_for_one_function.rs
+++ b/fearless_simd/examples/disable_avx2_for_one_function.rs
@@ -23,6 +23,8 @@ fn disable_avx2<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) {
 
 #[inline(always)]
 fn sigmoid<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) {
+    // verify that this function is never called with AVX2
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     debug_assert!(!matches!(simd.level(), Level::Avx2(_)));
 
     // If this function calls anything else with its `simd`,
--- a/fearless_simd/examples/disable_avx2_for_one_function.rs
+++ b/fearless_simd/examples/disable_avx2_for_one_function.rs
@@ -1,0 +1,46 @@
+// Copyright 2024 the Fearless_SIMD Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! This example demonstrates a SIMD function that's never compiled for AVX2.
+//!
+//! This can be useful if benchmarks show a specific insruction set regressing performance.
+
+use fearless_simd::{Level, dispatch, prelude::*};
+
+#[inline(always)]
+fn disable_avx2<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) {
+    let level = simd.level();
+    match level {
+        Level::Avx2(_) => {
+            // downgrade AVX2 to SSE4.2
+            let simd_sse4_2 = level.as_sse4_2().unwrap();
+            sigmoid(simd_sse4_2, x, out)
+        }
+        _ => sigmoid(simd, x, out),
+    }
+}
+
+#[inline(always)]
+fn sigmoid<S: Simd>(simd: S, x: &[f32], out: &mut [f32]) {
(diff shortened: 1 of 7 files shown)


fearless_simd/examples/disable_avx2_for_one_function.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2024 the Fearless_SIMD Authors
```
LaurenzV:
```suggestion
// Copyright 2026 the Fearless_SIMD Authors
```

Probably?

fearless_simd/src/macros.rs
```diff
@@ -140,19 +107,154 @@ macro_rules! dispatch {
...
+    ($token:expr, $simd:pat => $op:expr) => {{
+        /// Convert the `Simd` value into an `impl Simd`, which enforces that
+        /// it is correctly handled.
+        // TODO: Just make into a `pub` function in fearless_simd itself?
+        #[inline(always)]
+        fn launder<S: $crate::Simd>(x: S) -> impl $crate::Simd {
+            x
+        }
+
+        let __fearless_simd_token = $token;
```
LaurenzV, line 136: Do we need this intermediate variable, can we not just use `$token` directly?

LaurenzV (approved):
Nice, I like it! 

I wonder whether we should also change it such that `Level::fallback`always exists, and instead the `force_support_fallback` feature is also implemented in the dispatch macro instead. For consistency.

Also, I imagine in the future we will want something similar to neon once we add an `fp16` level.

fearless_simd/src/macros.rs
Shnatsel: We can, but it would change the semantics.

With the `token` being an `expr` (an arbitrary expression) it is possible to call this macro with things other than just a single "simd" variable. If we put `$token` under both arms instead of binding it to a variable, that expression would be evaluated twice. But if we bind the result to a variable first we only evaluate the expression once, which is more efficient and also less surprising.

DJMcNab:
Could you expand a bit on the expected user flow? The documentation doesn't seem to have much guidance for actual users. Is the thinking that that libraries treat these like `no_std`, and have passthrough features?

This seems like it makes it either exceptionally hard to actually disable the levels in practise, or else exceptionally easy to accidentally have users of a library not take advantage of SIMD.

I definitely like not feature flagging out the different actual levels though - that makes a lot of sense.

I'd almost be inclined to treat this in the same way that e.g. configuring c compilation is handled; i.e. with global environment variables and us using a build script to read them.

Shnatsel:
For libraries, passthrough with features of its own makes sense.

I think you're right about the features being difficult to disable. But I don't like the global variable approach either, too unconventional and magical.

We could make another macro that accepts a list of multiversioning targets, like `dispatch_selective!` or some such. That would let a library select the set they think is beneficial and disable some that don't; but I don't think that is very useful, and this can already be achieved today with the in-tree example.

I think what we really want is a global toggle. A good solution for that would be `cfg` values, e.g. `fearless_simd_disable_avx2`, which are global for the whole tree (I ran into issues with colliding cfg values deep in my dependencies). We want everything enabled by default, since there is no downside to it other than binary size, but whoever builds the final binary still can opt out of select features if they really know what they're doing - I want that for benchmarks, at the very least.

I'll prototype the cfg approach tomorrow.

DJMcNab:
I'm interested to hear more on why you'd want to disable this for benchmarking - that's a new use case I'm not familiar with. I image it can't be for benchmarking each level individually, because you'd achieve that by passing down an explicit `Level`.

Shnatsel:
RustFFT only supports AVX2 but with the addition of AVX-512 in fearless_simd it will gain an AVX-512 capability. I need to run both on AVX2 in an apples-to-apples comparison to understand what part of the performance difference is attributable to a different instruction set and which to different algorithms.

But you're right, for that use case I should just override the Level value.

PR #259, opened 2026-06-29
Remove Level::try_detect()

Shnatsel:
This function is buggy and useless: it returns `fallback` on Aarch64 where NEON presence is guaranteed, and doesn't take `-C target-cpu` or `-C target-feature` into account either. `Level::baseline()` does the same thing but has a better API and actually works.


Shnatsel:
Nevermind, I got confused, please ignore this.

PR #261, opened 2026-06-30
Widen inputs in standalone examples

Shnatsel:
Now that AVX-512 is supported, on machines with it the old 256-bit wide inputs would be too small and the output was always all zeroes.


LaurenzV:
Why, shouldn't it just fall back to SSE?

LaurenzV:
Ah, nvm, the examples use the native width/

LaurenzV (approved):

PR #262, opened 2026-06-30
Release 0.6.0

Shnatsel:
While we didn't make any API-breaking changes, previously the widest native vector width was 256 bits, and downstream code could implicitly depend on it. #261 demonstrates this failure mode. So I think calling it 0.6.0 is prudent.


Shnatsel:
Let's hold this off for a bit until the design concern raised in https://github.com/linebender/fearless_simd/pull/258#issuecomment-4849011562 is addressed

LaurenzV (approved):

PR #265, opened 2026-07-01
Add dynamic swizzles for 128-bit u8/i8 vectors

LaurenzV:
Swizzles are a whole can of worm and I do not intend to figure out how we should implement this generically right now. 😄 I think in the long term, we will mostly want to mirror portable_simd where we have `swizzle` for static swizzles (unfortunately this is currently a bit hard to implement due to current limitations of const generics) and `swizzle_dyn` for dynamic ones.

In any case, for now I just want to have a simple method we can use in vello to convert RGBA vectors into BGRA [during packing](https://github.com/linebender/vello/blob/492beec0a811c44eb7064d50e95dd2cf582dcc46/sparse_strips/vello_cpu/src/fine/lowp/mod.rs#L373-L376). We know the indices statically so that would be the best, but as mentioned this one's harder to implement so for now I decided to just go with dynamic swizzles instead. In order to keep things simple, I'm limiting this implementation
1) Is only done for 128-bit vectors, since doing it for vectors that exceed the native bit-width of the level is probably a bit tricky.
2) Only for u8/i8, since there's a discussion to be had on how to handle indexing for larger types.

So despite the complexity of swizzles, I hope adding this is non-controversial. The only controversial part is probably how to deal with OOB indices, but in the interest of getting the best performance I decided to not explicitly zero out invalid ones on x86, instead documenting that the result for those is unspecified.

### DIFF
--- a/fearless_simd/src/generated/fallback.rs
+++ b/fearless_simd/src/generated/fallback.rs
@@ -1165,6 +1165,14 @@ impl Simd for Fallback {
         .simd_into(self)
     }
     #[inline(always)]
+    fn swizzle_dyn_i8x16(self, a: i8x16<Self>, idxs: u8x16<Self>) -> i8x16<Self> {
+        let mut dest = [Default::default(); 16];
+        for (i, idx) in idxs.val.0.iter().copied().enumerate() {
+            dest[i] = a.val.0.get(idx as usize).copied().unwrap_or(0);
+        }
+        dest.simd_into(self)
+    }
+    #[inline(always)]
     fn reinterpret_u8_i8x16(self, a: i8x16<Self>) -> u8x16<Self> {
         a.bitcast()
     }
@@ -1766,6 +1774,14 @@ impl Simd for Fallback {
         result[0..16usize].copy_from_slice(&a.val.0);
         result[16usize..32usize].copy_from_slice(&b.val.0);
         result.simd_into(self)
+    }
+    #[inline(always)]
+    fn swizzle_dyn_u8x16(self, a: u8x16<Self>, idxs: u8x16<Self>) -> u8x16<Self> {
+        let mut dest = [Default::default(); 16];
+        for (i, idx) in idxs.val.0.iter().copied().enumerate() {
+            dest[i] = a.val.0.get(idx as usize).copied().unwrap_or(0);
+        }
+        dest.simd_into(self)
     }
     #[inline(always)]
     fn widen_u8x16(self, a: u8x16<Self>) -> u16x16<Self> {
--- a/fearless_simd/src/traits.rs
+++ b/fearless_simd/src/traits.rs
@@ -156,6 +156,11 @@ pub trait SimdCvtFloat<T: Seal>: Seal {
     fn float_from(x: T) -> Self;
 }
 
+/// Dynamic swizzle for byte vectors.
+pub trait SimdSwizzleDyn<S: Simd>: SimdBase<S> + Seal {
+    fn swizzle_dyn(self, idxs: crate::u8x16<S>) -> Self;
+}
+
 /// Concatenation of two SIMD vectors.
 ///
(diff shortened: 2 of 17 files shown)


fearless_simd/src/generated/fallback.rs
```diff
@@ -1165,6 +1165,14 @@ impl Simd for Fallback {
         .simd_into(self)
     }
     #[inline(always)]
+    fn swizzle_dyn_i8x16(self, a: i8x16<Self>, idxs: u8x16<Self>) -> i8x16<Self> {
+        let mut dest = [Default::default(); 16];
+        for (i, idx) in idxs.val.0.iter().copied().enumerate() {
+            dest[i] = a.val.0.get(idx as usize).copied().unwrap_or(0);
+        }
+        dest.simd_into(self)
+    }
```
Shnatsel, line 1174: Fallback implementations are usually manually unrolled so that platforms we don't support via intrinsics would have a better shot at autovectorization, see e.g. https://github.com/linebender/fearless_simd/blob/df33f0c5389011fb1341d62c0bff3b39f0d9301b/fearless_simd/src/generated/fallback.rs#L1009-L1114

I'd prefer to have this fully unrolled style applied here as well.

fearless_simd/src/generated/fallback.rs
```diff
@@ -1768,6 +1776,14 @@ impl Simd for Fallback {
...
     }
     #[inline(always)]
+    fn swizzle_dyn_u8x16(self, a: u8x16<Self>, idxs: u8x16<Self>) -> u8x16<Self> {
+        let mut dest = [Default::default(); 16];
+        for (i, idx) in idxs.val.0.iter().copied().enumerate() {
+            dest[i] = a.val.0.get(idx as usize).copied().unwrap_or(0);
+        }
+        dest.simd_into(self)
+    }
+    #[inline(always)]
```
Shnatsel, line 1786: same here

fearless_simd/src/generated/fallback.rs
Shnatsel: Separately, instead of `get().unwrap_or(0)` I would prefer `get((idx % 16) as usize )` which should be easier for the compiler to autovectorize and matches the current x86 behavior. But that's ultimately up to you, I'm not going to block this PR on that.

fearless_simd/src/traits.rs
```diff
@@ -156,6 +156,11 @@ pub trait SimdCvtFloat<T: Seal>: Seal {
     fn float_from(x: T) -> Self;
 }
 
+/// Dynamic swizzle for byte vectors.
+pub trait SimdSwizzleDyn<S: Simd>: SimdBase<S> + Seal {
+    fn swizzle_dyn(self, idxs: crate::u8x16<S>) -> Self;
+}
+
```
Shnatsel, line 163: The way this trait is defined locks us into u8x16 vectors for indices. Making it more general later is a semver-breaking change.

Something like `SimdSwizzleDyn<Idxs, S>` or an associated type `Indices` would be more future-proof, implemented today with `u8x16<S>` only for some types.

`Indices` sounds like it would be useful on `SimdBase` but I guess we can move that to `SimdBade` later once the implementation matures.

Shnatsel (commented):
The semver hazard is a concern. The rest is less critical.

But I am excited to see this, because this gives us a real edge over `std::simd` which requires `-Z build-std` for dynamic swizzles to perform well: https://doc.rust-lang.org/stable/std/simd/struct.Simd.html#method.swizzle_dyn

And we also can get better performance by not adhering to the "out of bounds is 0" rule, making it opt-in via a bitmask for the users that need it instead.

Shnatsel:
What we could easily provide for all vector sizes is swizzling within 128-bit blocks, which maps very naturally to all widths.

LaurenzV:
Superseded by #266.

PR #266, opened 2026-07-02
Implement swizzle_dyn_within_blocks, add tests

Shnatsel:
This addresses my criticism of #265

This maps to cheap hardware instructions on x86 and also maps really well to split+combine on blocks on NEON and WASM, so unlike #265 this can be implemented for all widths.

The API follows the `std::simd` [`swizzle_dyn`](https://doc.rust-lang.org/stable/std/simd/struct.Simd.html#method.swizzle_dyn), which uses byte-level index vectors. This also makes the implementation really straightforward, since everything is treated as byte vectors internally, so it's implemented for all types too.

The scalar fallback is branch-free thanks to `% len` which should be faster than the branchy fallback in #265 and also more amenable to autovectorization on obscure platforms we don't have intrinsics for.

### DIFF
--- a/fearless_simd_gen/src/mk_simd_trait.rs
+++ b/fearless_simd_gen/src/mk_simd_trait.rs
@@ -168,6 +168,8 @@ fn mk_simd_base() -> TokenStream {
             type Mask: SimdMask<S, Element = <Self::Element as SimdElement>::Mask>;
             /// A 128-bit SIMD vector of the same scalar type.
             type Block: SimdBase<S, Element = Self::Element>;
+            /// A same-width byte vector used to dynamically index bytes in this vector.
+            type Indices: SimdBase<S, Element = u8>;
             /// The array type that this vector type corresponds to, which will
             /// always be `[Self::Element; Self::N]`. It has the same layout as
             /// this vector type, but likely has a lower alignment.
--- a/fearless_simd_gen/src/mk_simd_trait.rs
+++ b/fearless_simd_gen/src/mk_simd_trait.rs
@@ -168,8 +168,6 @@ fn mk_simd_base() -> TokenStream {
             type Mask: SimdMask<S, Element = <Self::Element as SimdElement>::Mask>;
             /// A 128-bit SIMD vector of the same scalar type.
             type Block: SimdBase<S, Element = Self::Element>;
-            /// A same-width byte vector used to dynamically index bytes in this vector.
-            type Indices: SimdBase<S, Element = u8>;
             /// The array type that this vector type corresponds to, which will
             /// always be `[Self::Element; Self::N]`. It has the same layout as
             /// this vector type, but likely has a lower alignment.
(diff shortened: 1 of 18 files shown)


Shnatsel:
And just as I'm done placating Clippy, CI is stuck. Ugh.

fearless_simd_gen/src/mk_simd_trait.rs
```diff
@@ -168,6 +168,8 @@ fn mk_simd_base() -> TokenStream {
             type Mask: SimdMask<S, Element = <Self::Element as SimdElement>::Mask>;
             /// A 128-bit SIMD vector of the same scalar type.
             type Block: SimdBase<S, Element = Self::Element>;
+            /// A same-width byte vector used to dynamically index bytes in this vector.
+            type Indices: SimdBase<S, Element = u8>;
```
LaurenzV: Could we not just use `Self::Bytes` instead?

fearless_simd_gen/src/ops.rs
```diff
@@ -587,6 +599,14 @@ const BASE_OPS: &[Op] = &[
         },
         "Like `slide`, but operates independently on each 128-bit block.",
     ),
+    Op::new(
+        "swizzle_dyn_within_blocks",
+        OpKind::BaseTraitMethod,
+        OpSig::SwizzleDynWithinBlocks,
+        "Dynamically swizzle this vector's bytes independently within each 128-bit block.\n\n\
+        The `indices` operand is a same-width byte vector. For each output byte, index values `0..=15` select the corresponding byte from the same 128-bit input block.\n\n\
+        Out-of-range index behavior is target-native: on x86 this follows `pshufb`, while NEON and WASM table swizzles produce zero.",
```
LaurenzV: Let's not specify more than just "is target-dependent", people shouldn't start relying on target-specific guarantees in the documentation so that we still have the flexibility to change it in the future if we want to.

fearless_simd_tests/tests/harness/mod.rs
```diff
@@ -3794,6 +3794,406 @@ fn bitcast_native<S: Simd>(simd: S) {
...
+    let result = do_swizzle::<S, u8x16<S>>(value, indices);
+
+    assert_eq!(
+        *result,
+        [15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
+    );
+}
+
+#[simd_test]
+fn swizzle_dyn_within_blocks_x86_pshufb_oob<S: Simd>(simd: S) {
```
LaurenzV: As per above suggestion to not make any target-specific guarantees, let's turn this into a test that simply checks that it doesn't panic (like was done in my original PR) instead of checking the actual output.

LaurenzV (commented):
LGTM overall!

fearless_simd_gen/src/mk_simd_trait.rs
Shnatsel: To clarify, you mean renaming it to `Self::Bytes`? I don't think this type exists right now.

LaurenzV: `SimdBase` requires the type to implement the `Bytes` trait, which has the `Bytes` type that basically does the same as the `Indices` type you introduced, I think. But maybe it's still worth keeping them separately for semantic reasons.

Shnatsel: Ah! You are correct.

No, it's not worth keeping them separate. Even the scalable SIMD on SVE and RISC-V is only defined up to 2048 bytes in length, which is still fully addressable with u8 indices even for general swizzles.

Shnatsel: Done. Thanks again for the thorough review!

LaurenzV (approved):

Issue #256, opened 2026-06-28
Add an SSE2 level for use with the `kernel!` macro


Shnatsel:
SSE2 is interesting because it is part of the baseline x86_64 instruction set, so it can be used without any runtime dispatch and the associated overhead.

I've [prototyped](https://github.com/Shnatsel/fearless_simd/tree/sse2-backup) a dedicated SSE2 level in the generator as a replacement for the `Fallback` level on x86, but most operations already autovectorize just fine, so it wasn't worth the additional complexity in the generator for the sake of a handful of operations for [2% of users](https://firefoxgraphics.github.io/telemetry/#view=system).

However, there are use cases for the SSE2 level with the `kernel!` macro. For example, `image-webp` crate [wants a safe way to access SSE2 intrinsics and safe loads/stores](https://github.com/image-rs/image-webp/issues/163). So perhaps we can provide a minimal implementation that falls through to either SSE4.2 (where only SSE2 instructions are used) or fallback, without writing any SSE2-specific intrinsic code.

LaurenzV:
Well, that once again sounds like we should have https://github.com/linebender/fearless_simd/pull/108 no? 😄

Shnatsel:
Sadly that alone isn't enough, we also need safe loads/stores at the very least. And right now those are exposed through types such as `f32x4`.

Shnatsel:
Done in #270

PR #270, opened 2026-07-12
Add SSE2 backend

Shnatsel:
Fulfills #256

SSE2 is guaranteed to be present in the baseline on x86_64, and the Rust i686 targets also guarantee its presence. You need to use the tier-2 i586 target to opt out. So this replaces Fallback on x86 and does not bloat the binary, even on 32-bit targets.

Not having SSE4.2 is quite rare these days, but this is still useful in contexts where dynamic dispatch is expensive. For example, this unlocks access to explicit SIMD in the `png` crate which uses SSE2 with autovectorization without runtime dispatch, but autovectorization sometimes regresses on newer rustc versions, so explicit SIMD is desirable.

To reduce the maintenance burden, most operations on SSE2 level reuse either the SSE4.2 path or the scalar fallback. Only a few select operations get dedicated SSE2 paths. I feel those exceptions are justified because autovectorization is still quite fickle. For example, individual calls to min() and max() autovectorize into identical assembly, but without intrinsics this falls apart and turns from 8 instructions into 181 - I've even experimented with alternative formulations of the fallback level to no avail:

```rust
#[inline(always)]
fn minmax_u8<S: Simd>(
    simd: S,
    a: &[u8; 16],
    b: &[u8; 16],
    min_out: &mut [u8; 16],
    max_out: &mut [u8; 16],
) {
    let a = u8x16::from_slice(simd, a);
    let b = u8x16::from_slice(simd, b);
    a.min(b).store_slice(min_out);
    a.max(b).store_slice(max_out);
}
```


CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ You can find its changes [documented below](#060-2026-07-10).
 
 ## [Unreleased]
 
+### Added
+
+- Added an `Sse2` level as the x86/x86_64 baseline, with runtime dispatch upgrading through SSE4.2, AVX2, and AVX-512 where available.
+
+### Changed
+
+- On x86_64 targets with static SSE2 support, `Level::baseline()` now returns `Sse2` instead of `Fallback`.
```
LaurenzV: I thought it's all x86_64 targets?

Also, what about x86_32? Does this mean we will compile 6 different levels for those now? I wonder whether we should exclude anything above SSE2 on 32-bit since there likely aren't any CPU's supporting it anyway (in a follow-up if at all)?

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -19,16 +19,19 @@ use quote::{ToTokens as _, format_ident, quote};
 
 #[derive(Clone, Copy, PartialEq, Eq)]
 pub(crate) enum X86 {
+    Sse2,
     Sse4_2,
     Avx2,
     Avx512,
 }
 
+pub(crate) const SSE2_FEATURES: &str = "fxsr,sse,sse2";
```
LaurenzV, line 28: I thought that our goal was to get rid of fallback completely for x86_64 (except if `force_fallback` is enabled). But it seems like there are targets that support sse2 but not fxsr?

```
rustc --print cfg --target x86_64-unknown-trusty 
debug_assertions
panic="abort"
target_abi=""
target_arch="x86_64"
target_endian="little"
target_env=""
target_feature="crt-static"
target_feature="sse"
target_feature="sse2"
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="64"
target_has_atomic="8"
target_has_atomic="ptr"
target_os="trusty"
target_pointer_width="64"
target_vendor="unknown"
```

Am I missing something here?

CHANGELOG.md
LaurenzV: Ah soryr, only now saw that you mentioned it. 😅 SO am I understanding correctly that all 32-bit targets supported by Rust support SSE2?

fearless_simd_gen/src/mk_x86.rs
Shnatsel: The tier-1 x86 targets such as `i686-unknown-linux-gnu` support SSE2 and fxsr. You can verify it with `rustc --print=cfg --target i686-unknown-linux-gnu`

Tier-2 targets such as `i586-unknown-linux-gnu` get SSE2 via runtime detection, like they would get SSE4.2.

`x86_64-unknown-trusty` is tier 3 so I don't care what happens there.

LaurenzV: > x86_64-unknown-trusty is tier 3 so I don't care what happens there.

Well, even if it's Tier-3 I'd rather not introduce unsoundness. Maybe it's just a bug in the print feature, idk, but still, better safe than sorry.

So I would suggest that either we remove "fxsr" from the level, or we explicitly check that feature wherever we construct it in addition to "sse2", for example here:

https://github.com/Shnatsel/fearless_simd/blob/3843c4dc6893c6c977887c116a1c766534aba909/fearless_simd/src/lib.rs#L408-L410

And in all other locations.

Shnatsel: Ah, good call, I missed that `fxsr` is not implied by `sse2` and forgot about that since I wrote the code. I'll fix it. Good catch, thanks!

Shnatsel: `fxsr` allows for fast restore of SSE registers, so we do want it in the SSE2 level as it's very much relevant to SSE2 performance. I'll add checks for it everywhere.

Shnatsel: To be fair, there is no actual hardware with SSE2 but not fxsr that ever shipped. fxsr was added in Pentium 3 alongside SSE, while SSE2 was added in Pentium 4.

Shnatsel: Nevermind, upon further research it seems that `fxsr` is used by the OS for context switching, not by the userland code itself when entering or exiting functions. I'll drop `fxsr` instead of plumbing it into the SSE2 level.

Shnatsel:
I went back and forth on this a bunch, but eventually decided to add fxsr checks everywhere since it's part of the baseline for x86_64 and i686 targets, and there are some uses for those instructions inside kernel! functions. 

The tier3 target is probably just misconfigured, so now we disable SIMD on it instead of making potentially unsound calls.

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -686,6 +722,134 @@ fn signed_literal(value: u64, bits: u32) -> TokenStream {
...
+
+/// Invert an SSE2 mask expression.
+///
+/// SSE2 has no vector NOT intrinsic, so we synthesize an all-ones vector by
+/// comparing zero with itself and XOR the mask with that.
+fn sse2_not_mask_expr(mask: TokenStream) -> TokenStream {
+    quote! {
+        {
+            let zero = _mm_setzero_si128();
+            let all_ones = _mm_cmpeq_epi8(zero, zero);
```
LaurenzV: Why not just use `_mm_set1_epi8(-1)`?

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -1304,6 +1451,13 @@ impl X86 {
             }
         }
 
+        if *self == Self::Sse2
+            && vec_ty.scalar == ScalarType::Float
+            && matches!(method, "floor" | "ceil" | "round_ties_even" | "trunc")
```
LaurenzV, line 1462: I feel like at some point we should introduce constants for the names... But also not for this PR.

fearless_simd/src/generated/sse2.rs
```diff
@@ -0,0 +1,13647 @@
...
+        kernel(self, a, b, c)
+    }
+    #[inline(always)]
+    fn min_i8x16(self, a: i8x16<Self>, b: i8x16<Self>) -> i8x16<Self> {
+        crate::kernel!(
+            #[inline(always)]
+            fn kernel(token: Sse2, a: i8x16<Sse2>, b: i8x16<Sse2>) -> i8x16<Sse2> {
+                _mm_or_si128(
+                    _mm_and_si128(_mm_cmpgt_epi8(a.into(), b.into()), b.into()),
+                    _mm_andnot_si128(_mm_cmpgt_epi8(a.into(), b.into()), a.into()),
```
LaurenzV: Hopefully it's optimized away, but maybe the imask should be stored in a variable first?

LaurenzV (approved):
Once again mostly relying on our existing tests for correctness, but LGTM overall!

fearless_simd_gen/src/mk_x86.rs
Shnatsel: Yeah -1 should work and be more direct. Thanks!

Shnatsel:
The remarks are addressed, merging. I appreciate the in-depth review as always!

PR #272, opened 2026-07-13
Cache the feature detection

reinerp:
Feature detection takes a lot of instructions on x86 (https://rust.godbolt.org/z/xW1rGs3v3), so we might as well cache it.

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -370,65 +370,65 @@ impl Level {
         }
         #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
         {
-            if x86_detects_icelake_avx512() {
-                return unsafe { Self::Avx512(Avx512::new_unchecked()) };
+            use core::sync::atomic::{AtomicU8, Ordering};
+
+            static CACHE: AtomicU8 = AtomicU8::new(0);
+            mod code {
+                pub(super) const SSE4_2: u8 = 1;
+                pub(super) const AVX2: u8 = 2;
+                pub(super) const AVX512: u8 = 3;
+                pub(super) const FALLBACK: u8 = 4;
             }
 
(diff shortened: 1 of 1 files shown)


LaurenzV:
Hmm, I might remember wrong but I thought this is already cached internally....

LaurenzV:
Ah nvm, you basically mean caching the union of all detected features.

raphlinus:
My personal feeling is that this is not very important, as in general we want people to pass the level around explicitly for fine-grained function calls. Ideally, we can do feature detection more efficiently at an even finer grain than atomics, but Rust doesn't have a good way to express it. But I'm open to an argument that doing an atomic is marginally better than relying on the feature detection macro, especially when there are lots of features to detect.

See https://internals.rust-lang.org/t/better-codegen-for-cpu-feature-detection/22083 for more discussion.

DJMcNab:
In a correctly operating program, this should have O(1) callers. I worry that this kind of optimisation could actually be a net de-optimisation, as people would be more likely to incorrectly run feature detection at a narrower scope...

I'd definitely be in favour of seeing some numbers for how much time this saves per call, how much it costs for the first call (hopefully negligible). This is at least fairly standard in similar libraries...

Shnatsel:
I don't think there's a better way to cache it, if an application needs it. Like, you could put a resolved `Level` into a thread local or into a `OnceLock`, but both of those are more expensive than a direct relaxed atomic load.

For comparison, `#[multiversion]` does an atomic load and dispatch every time it enters a function annotated with it, and hardly anyone notices. On x86 this is literally a single regular load, a relaxed atomic is that cheap.

I want to take a closer look at the implementation later. There are further things we can do to maybe improve robustness, such as putting the numeric conversion onto the `Level` enum as either a private or doc(hidden) function. I also want to research how this interacts with semver.

I see the actual feature detection is outlined into a #[cold] function already, nice!

Shnatsel:
We've eventually merged a slightly different implementation of this in #278, so I'm going to go ahead and close this. Thank you for getting the ball rolling!

PR #274, opened 2026-07-13
Add shift_elements_* and rotate_elements_* functions to match std::simd API

### LINKED ISSUE #269: Add discoverable shift and rotate functions
[`slide`](https://docs.rs/fearless_simd/latest/fearless_simd/trait.SimdBase.html#tymethod.slide) can already be used to implement all kinds of shifts and rotates, but if you search the documentation for `shift` or `rotate`, nothing comes up. It is not discoverable enough, [making people think it doesn't exist](https://www.reddit.com/r/rust/comments/1uuewck/fearless_simd_v06_brings_avx512_and_ways_to/ox4iu49/).

Providing direct equivalents to `std::simd`'s [`shift_elements_left`](https://doc.rust-lang.org/stable/std/simd/struct.Mask.html#method.shift_elements_left) and [`rotate_elements_right`](https://doc.rust-lang.org/stable/std/simd/struct.Mask.html#method.rotate_elements_right) would make that more discoverable.

If that's a problem (const generic code explosion?), we should at least provide search aliases.

Shnatsel:
Adds convenience functions that lower into `slide` behind the scenes, matching the `std::simd` API and improving discoverability.

No new intrinsic/platform-specific code is added, since this simply lowers into calls to the existing slide() op.

Closes https://github.com/linebender/fearless_simd/issues/269

### DIFF
--- a/fearless_simd_tests/tests/harness/mod.rs
+++ b/fearless_simd_tests/tests/harness/mod.rs
@@ -4438,6 +4438,777 @@ fn slide_f32x4<S: Simd>(simd: S) {
 }
 
 #[simd_test]
+fn rotate_elements_f32x4<S: Simd>(simd: S) {
+    let a = f32x4::from_slice(simd, &[0.0, 1.0, 2.0, 3.0]);
+    assert_eq!(*a.rotate_elements_left::<3>(), [3.0, 0.0, 1.0, 2.0]);
+    assert_eq!(*a.rotate_elements_right::<3>(), [1.0, 2.0, 3.0, 0.0]);
+}
+
+#[simd_test]
+fn shift_elements_f32x4<S: Simd>(simd: S) {
+    let a = f32x4::from_slice(simd, &[0.0, 1.0, 2.0, 3.0]);
+    assert_eq!(*a.shift_elements_left::<3>(99.0), [3.0, 99.0, 99.0, 99.0]);
+    assert_eq!(*a.shift_elements_right::<3>(99.0), [99.0, 99.0, 99.0, 0.0]);
+}
+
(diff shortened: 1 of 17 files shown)


LaurenzV:
Did you notice any change regarding compile times? Also, should we maybe just replace the existing slide method then?

Shnatsel:
It may still be worth exposing slide, since it's more flexible.

`cargo build --tests` which does instantiate these functions goes from ~8.5s to ~9.3s, and `cargo build --tests --release` from ~11.5s to ~12.5s on my x86 machine, so there is a compilation time hit if you do use them.

We're noticing the build time hit because we have instantiated these functions for all 24 vector types and all instruction sets. For dependents there is no change in build times until they actually use these functions, because all the newly added functions all generic and don't get instantiated into code unless they are actually used.

If I take main and comment out all the slide tests, the compilation time drops from ~8.5s to ~8s. So using `slide` directly costs exactly the same as using any of these newly added methods.

fearless_simd_tests/tests/harness/mod.rs
LaurenzV, line 1: We really should start splitting up the tests into different files. 😅

LaurenzV (approved):
Can't say that I'm super convinced that we need really those wrapper methods, but I guess aiming for SIMD-compatibility with std::simd isn't a bad idea. 😄

Shnatsel:
Well, the alternative is to alias `slide` to `rotate_elements` and `shift_elements` in search, but for us the implementation complexity is minimal, and for actual users the compile-time impact is the same regardless, so I think the extra bit of compatibility is worth it.

PR #275, opened 2026-07-18
Reorganize tests, expand test coverage

Shnatsel:
~~Stacked on top of #253 which adds tests for 64-bit integer types~~ Rebased onto main after that was merged

The first commit moves existing tests around. Tests that were haphazardly strewn about multiple files are now organized by operation, one file = one operation. This makes the tests much easier to navigate and makes the missing test coverage readily apparent. This also reduces the from-scratch build time with `cargo build --tests` from 19s to 14s on my machine.

The two subsequent commits add 7k lines of tests to fill in coverage gaps. This was done with Python scripts and a bit of fixup. 

The generated tests are slightly different from the existing patterns in that they use `array::from_fn` as opposed to handwritten inputs and outputs. If that's a problem, one could probably put up an LLM to bringing them into the same style as the pre-existing tests, if desired.


LaurenzV:
Ready for review?

Shnatsel:
Yes. I have rebased on main and addressed the review comment about tests from #263.

LaurenzV (commented):
I haven't checked every test in detail but I did skim over each file. With that said, Codex found some more stuff, I checked the first two and think they are reasonable, I haven't checked the other three yet.

Image

LaurenzV (approved):

PR #276, opened 2026-07-19
Add swizzle_dyn_precise, mirroring swizzle_dyn from `std::simd`

Shnatsel:
I wondered how `std::simd` lowers `swizzle_dyn` - not the limited within_blocks version we added in #266 but the full arbitrary version - and the answer turned out to be "poorly". So I got [nerd-sniped](https://xkcd.com/356/), badly. It's 4:30 AM as I write this.

The in-tree std::simd is trivially suboptimal, not even taking the most obvious optimization avenues like using 512-bit table lookups on NEON. The NEON gap is [fixed upstream](https://github.com/rust-lang/portable-simd/blob/7d497cca160ae6062acc1a2db838667f83c0b58e/crates/core_simd/src/swizzle_dyn.rs#L127-L182) but not synced into std yet.

They also lower into a slow, branchy scalar fallback for all cases that don't have a native shuffle op available. There is a better way: if all you have is 128-bit vectors, you can get a 256-bit vector shuffle with four 128-bit shuffles and a couple bitwise ORs.

This PR fixes all of that:

1. Implements all the native ops, and reuses the [tuned AVX2 impl for 256-bit shuffles](https://github.com/rust-lang/portable-simd/blob/7d497cca160ae6062acc1a2db838667f83c0b58e/crates/core_simd/src/swizzle_dyn.rs#L194-L225) from std::simd (with attribution)
2. Adds the "4 small shuffles make a big shuffle" decomposition for SSE4.2 and WASM
3. ~~Adds a further optimized 512-bit shuffle on AVX2 which is better than the generic decomposition~~

I've written [a toy SIMD base64 decoder](https://github.com/Shnatsel/toy-simd-base64) using 512-bit shuffles to benchmark this. On Zen4 the specialized AVX2 implementation improves the benchmark by 5%, but it'll never use it because it has AVX-512. On Zen2, where the specialized 512-bit AVX2 is supposed to help the most, there was no difference on benchmarks on real hardware at all, so I went ahead and deleted the specialized 512-bit AVX2 implementation.

Hand-written unit tests are included. Tests on random data that check the optimized SIMD impls against a trivial scalar impl were used to verify the implementation but aren't included in the final PR to avoid extra deps and slow tests (they're preserved in history for posterity but reverted).

I've implemented the *_precise variant first because it is genuinely important for some algorithms, mirrors the std::simd implementation exactly, and the "out of bounds is zero" property actually allows for cool optimizations for non-native vector widths.

This probably deserves a blog post.


Shnatsel:
I've also submitted similar changes upstream to `std::simd`:

- https://github.com/rust-lang/portable-simd/pull/540
- https://github.com/rust-lang/portable-simd/pull/542

One PR is already merged, the other is awaiting review.

Shnatsel:
The implementation has evolved somewhat since the PR description was written, here's a blog post that goes into detail and is more representative of the current state: https://shnatsel.github.io/improving-std-simd-swizzle-dyn/

Shnatsel:
anxsan12 [suggested](https://github.com/rust-lang/portable-simd/pull/542#issuecomment-5077221668) a further optimization that turned out to be worthwhile, so I applied it. 

It is also merged upstream: https://github.com/rust-lang/portable-simd/pull/543

Shnatsel:
@DJMcNab could you take a quick look?

All the complex parts (x86 implementations, 2x vector width trick) have already been reviewed upstream in `std::simd`:

 - https://github.com/rust-lang/portable-simd/pull/540
 - https://github.com/rust-lang/portable-simd/pull/543
 - https://github.com/rust-lang/portable-simd/pull/542
 - https://github.com/rust-lang/portable-simd/pull/545

WASM and NEON implementations are trivial.

There are unit tests and [random testing](github.com/linebender/fearless_simd/pull/276/commits/896f20908a10d97199ddfea43d63c6976f7c4b7d) to ensure correctness, so I'm quite confident in this code.

fearless_simd/src/generated/sse2.rs
LaurenzV, line 1: It's a bit unfortunate that it makes the file that much longer. :/ Have you checked whether for loops create similar assembly? If so, maybe that would be preferable (same for `Fallback`).

LaurenzV (approved):
Only took a brief look.

fearless_simd/src/generated/sse2.rs
Shnatsel: The original motivation for fully unrolled forms is https://github.com/linebender/fearless_simd/issues/29#issuecomment-3577305034

In a nutshell, LLVM is good at autovectorizing things on x86. But other architectures aren't nearly as sophisticated, and many generally useful optimizations still only exist in the x86 backend. The default Aarch64 cost model in LLVM also has very little loop unrolling, despite ample ILP on recent Aarch64 chips.

I'll double-check what the fallback for this particular op does on SSE2 and on something obscure like POWER.

Shnatsel: I didn't look at the fallback very closely until you asked, but I did manage to find a better fallback formulation after a while. It even improves large swizzles on SSE4.2, which is a somewhat relevant target. Thanks!

PR #278, opened 2026-07-23
Cache the feature detection, take 2

Shnatsel:
This is inspired by #272 by @reinerp

The difference from #272 is in that we don't introduce a custom numbering scheme for the levels that unsafe code depends on. We just use `LazyLock<Level>`  as the cache without rolling our own on top of an `AtomicU8`.

llvm-mca shows identical performance for both (a handful of cycles). `LazyLock<Level>` is larger but easily fits within a cache line so the cost is the same. The only failure mode is that LazyLock could cross a cache line boundary, but wrapping it in Aligned128 prevents it. I've verified that makes the size of the final `Aligned128<LazyLock<Level>>` is 128 bits and therefore cannot cross a cache line at 128 bit alignment.

Arguably frequent calls to `Level::new()` are a performance pitfall, but not doing any caching and making it expensive is an even bigger pitfall because it's always free on Aarch64 and WASM, so if anyone develops for those platforms they'll find that exact same code unbearably slow on x86.

### DIFF
--- a/fearless_simd/src/lib.rs
+++ b/fearless_simd/src/lib.rs
@@ -259,7 +259,8 @@ pub mod x86 {
 use std::sync::LazyLock;
 
 #[cfg(all(feature = "std", any(target_arch = "x86", target_arch = "x86_64")))]
-static X86_LEVEL: LazyLock<Level> = LazyLock::new(detect_x86_level);
+static X86_LEVEL: support::Aligned128<LazyLock<Level>> =
+    support::Aligned128(LazyLock::new(detect_x86_level));
 
 // Sourced from `rustc --print=cfg --target x86_64-unknown-linux-gnu -C target-cpu=icelake-server`
 // and pruned against the features implied by `avx512f` which can be viewed via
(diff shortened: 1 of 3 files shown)


fearless_simd/src/lib.rs
```diff
@@ -352,97 +413,23 @@ impl Level {
     /// handling the `None` case as they deem fit (probably panicking).
     /// This strategy avoids users of the library inadvertently using the fallback level,
     /// even if the requisite target features are available.
-    ///
-    /// If you are on an embedded device where these macros are not supported,
-    /// you should construct the relevant variants yourself, using whatever
-    /// way your specific chip supports accessing the current level.
-    ///
```
DJMcNab, line 359: Is this a drive by fix to remove duplicate docs?

Or is it the case that we don't actually support manually constructing the level yet - I'd believe that...

DJMcNab (approved):

fearless_simd/src/lib.rs
Shnatsel: We don't support creating levels unsafely directly. We only support creating tokens such as `Avx2`: https://docs.rs/fearless_simd/latest/fearless_simd/struct.Avx2.html

But once you have a token you can make the appropriate `Level` manually. I'll update the docs to make it more clear.

Shnatsel: Oh, we do, it's just counter-intuitive. I've updated the docs to make them more clear (hopefully).

PR #279, opened 2026-07-24
Optimize shrv_i64 for AVX2

Shnatsel:
LLVM already autovectorizes this in isolation, but autovec can easily break when inlined into larger functions. 

This enshrines the autovec formulation, but gains some performance by dropping 2 instructions for handling shifts of size 64 or greater (i.e. shifting out of bounds entirely), which are inconsistent across backends today anyway.

Tests used in development but omitted to avoid conflicting with the tests refactor PR #275. Ideally #275 should be merged first to ensure test coverage.

Gains for other sizes are possible but blocked on suboptimal codegen by rustc: https://github.com/linebender/fearless_simd/issues/280

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -4723,11 +4723,18 @@ impl Simd for Avx2 {
     }
     #[inline(always)]
     fn shrv_i64x2(self, a: i64x2<Self>, b: i64x2<Self>) -> i64x2<Self> {
-        [
-            i64::shr(a[0usize], &b[0usize]),
-            i64::shr(a[1usize], &b[1usize]),
-        ]
-        .simd_into(self)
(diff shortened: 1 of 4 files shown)


Shnatsel:
I believe out-of-bounds shifts are just inconsistent across the board. Simple shift left by a constant (not vectored) panics in debug mode on fallback, wraps in release; x86 and NEON zero; WASM wraps. Vectored shifts left either zero or debug-panic.

More to the point, vectored shifts right debug-panic or release-wrap in the scalar fallback, also used on SSE and NEON, and sign-fill on NEON, AVX2 and AVX-512.

I am of the opinion that we should expose the cheapest native op, and if someone needs the bounding semantics they can apply their own AND mask onto the shift values.

LaurenzV (approved):
Gonna trust your experiments here!

PR #282, opened 2026-07-24
Add a comment documenting slide optimization choice

Shnatsel:
I spent some time trying to optimize slide from permutex2var into more specific intrinsics, but the result was the same (literally identical assembly) or worse in the case of floats due to https://github.com/rust-lang/rust/issues/159842

So document the rationale for keeping current implementation instead.

### DIFF
--- a/fearless_simd_gen/src/mk_x86.rs
+++ b/fearless_simd_gen/src/mk_x86.rs
@@ -2643,6 +2643,11 @@ impl X86 {
             let level = self.token();
             let ty = vec_ty.rust();
             let vec = quote! { #ty<#level> };
+            // This deliberately expresses slides of every lane width as byte permutations.
+            // LLVM folds the constant indices to the same instructions as the equivalent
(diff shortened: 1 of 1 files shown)


fearless_simd_gen/src/mk_x86.rs
```diff
@@ -2643,6 +2643,11 @@ impl X86 {
             let level = self.token();
             let ty = vec_ty.rust();
             let vec = quote! { #ty<#level> };
+            // This deliberately expresses slides of every lane width as byte permutations.
```
LaurenzV, line 2646: Should we mark this as a TODO in case there are plans to fix it upstream, so we don't forget about it?

LaurenzV (approved):

fearless_simd_gen/src/mk_x86.rs
Shnatsel: We already produce identical assembly, so no point in changing it.

PR #283, opened 2026-07-26
Document that the result of shifting out of bounds varies by platform

Shnatsel:
Now that I've researched it for https://github.com/linebender/fearless_simd/pull/279#issuecomment-5074075264, might as well document it.

I also had Codex write tests to probe for other platform divergences and it didn't find anything that isn't already documented, other than differences in how the signalling bit in `NaN`s is propagated or not in some ops. But the standard library [notes](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.from_bits) that you "very likely" don't care if a NaN is signalling or not, so I didn't bother documenting that.


fearless_simd_gen/src/ops.rs
```diff
@@ -1006,29 +1006,33 @@ const INT_OPS: &[Op] = &[
...
         This operation is not implemented in hardware on all platforms. On WebAssembly, and on x86 platforms without AVX2, this will use a fallback scalar implementation.",
     ),
     Op::new(
         "shr",
         OpKind::Overloaded(CoreOpTrait::Shr),
         OpSig::Shift,
         "Shift each element right by the given number of bits.\n\n\
-        For unsigned integers, zeros are shifted in on the left. For signed integers, the sign bit is replicated.",
+        For unsigned integers, zeros are shifted in on the left. For signed integers, the sign bit is replicated.\n\n\
+        When shifting out of bounds (e.g. shifting a 32-bit value by 32 or more), the result is implementation-defined and may vary by platform.",
```
LaurenzV, line 1027: Since you mentioned that it might panic in debug mode for fallback, maybe expicitly adding "or panic" might be a good idea?

LaurenzV (approved):

fearless_simd_gen/src/ops.rs
Shnatsel: Should we just make it consistently panic in debug mode, then?

LaurenzV: For the other backends? I don't think we have to, just would be good to clarify that "implementation-defined" can also include panicking.

Shnatsel: I thought it might be good idea to panic in debug mode here, since doing out-of-bounds shifts and relying on the result is a footgun. But std has explicit wrapping shifts while we don't, so it's probably best not to force panics on people.

We could make fallback wrap instead of panicking; I think that's the best approach, considering the above.

LaurenzV: Wrap sounds good to me.

PR #284, opened 2026-07-27
Make bitcast() work in generic contexts, remove reinterpret_* methods

Shnatsel:
Motivated by [#simd > Missing reinterpret_u64_&#42; methods on Simd trait](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/Missing.20reinterpret_u64_.2A.20methods.20on.20Simd.20trait/with/612860358)

- Make `bitcast()` work in generic contexts on variable-width vectors
    - All SIMD vector types now use `Self::u8s` as their `Bytes` representation, which avoids arcane bounds
    - Add doctest and test coverage for generic bitcasts
- Remove type-specific reinterpret_* methods which blow up the API surface quadratically
- Remove intrinsics-backed `cvt_to/from_bytes` implementations, use safe transmute to implement them instead
- Alias `reinterpret` and `transmute` to `bitcast` to make it more discoverable via search


LaurenzV (approved):
Nice! Bonus points for maybe letting AI code up some examples that use reinterpret before vs. after and making sure there are no obvious regressions in assembly, but here I'm less concerned than in #286, so it should be fine.

Shnatsel:
I had Codex do that before opening the PR and the assembly is identical, but self-contained examples always optimize well, so that's neither here nor there.

I see Vello has a few reinterpret calls, so I'll port it to this PR, disassemble the relevant functions and compare before and after.

Shnatsel:
This optimizes fine, save for a register allocation perturbation in one function on AVX-512 only: https://github.com/linebender/vello/pull/1773

PR #286, opened 2026-07-27
Make conversion to/from arrays generic

Shnatsel:
Move conversion to/from array into a separate trait, `SimdArray`. Unifies conversion handling between vectors and masks. `SimdArrayRef` with in-place conversions to references is added as an extension for vectors only (not masks).

In most cases `store_array` no longer needs the length to be explicitly qualified; in the few places where it is needed, it can be done through the usual turbofish synatx, same as `From` and `TryFrom`.

This improves generic programming:

- Generic code can now convert to/from arrays
- Native-width types such as `S::f32s` can now be converted to/from arrays

Despite `N` now living in `SimdArray`, it is still accessible through `SimdBase` through supertrait relationship.

Pairs well with #285 which makes arrays more useful in general. Also conflicts with it, but _c'est la vie,_ I'll resolve conflicts in one when the other lands.

This also drops 204 code-generated wrappers around various safe transmute functions, reducing the size of the generated code.


LaurenzV:
Looks nice in principle, but we should benchmark to make sure it doesn't cause regressions. It has happened before that changing how arrays are loaded/stored resulted in differences.

Shnatsel:
I tried converting Vello to this and ended up disliking how this ends up getting used in practice.

Being able to load and store masks and vectors without knowing which one it is just isn't very useful, but it makes the docs harder to navigate and requires changes to the existing user code in some cases (`<Self as SimdBase<_>>::from_slice(simd, slice)` now needs to be rewritten as `<Self as SimdArray<_>>::from_slice(simd, slice)`, extra trait to import if not using the prelude).

I think the upsides of this can be achieved without requiring code changes from users, and without complicating and fragmenting the trait hierarchy. I'll experiment with that.

Shnatsel:
Closing in favor of #292

PR #288, opened 2026-07-28
Fix byte-wise shl to not saturate when overflowing u8

### LINKED ISSUE #287: u8 shl with overflowing value produces incorrect result on x86
Trivial reproducer:

```
u8x64::splat(s, 3) << 7 == u8x64::splat(s, 128)
```

fails at HEAD on x86, the shift sets all bytes to 255.

The cause is that the generated Shl implementation widens to 16-bit to execute the shift before shrinking back to 8-bit, to work around the lack of 8-bit shift instructions on x86. The `packus` instruction used to shrink back to 8-bit uses saturating arithmetic, so if the shift pushed some bits into the upper byte of a u16, it gets clamped to 255 instead of truncating. The pack instruction has the same semantics on all x86 vector extensions, so this affects all the SSE and AVX generations.

The implementation of `shlv` has the correct logic already (apply an explicit mask before packing), so it just needs to be applied to `shl` as well.

I have the fix implemented and will send a PR for it shortly.

danderson:
Fixes #287

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -969,6 +969,9 @@ impl Simd for Avx2 {
                 let hi_16 = _mm_unpackhi_epi8(val, _mm_cmpgt_epi8(_mm_setzero_si128(), val));
                 let lo_shifted = _mm_sll_epi16(lo_16, shift_count);
                 let hi_shifted = _mm_sll_epi16(hi_16, shift_count);
+                let byte_mask = _mm_set1_epi16(0x00ff);
+                let lo_shifted = _mm_and_si128(lo_shifted, byte_mask);
+                let hi_shifted = _mm_and_si128(hi_shifted, byte_mask);
                 _mm_packs_epi16(lo_shifted, hi_shifted).simd_into(token)
             }
         );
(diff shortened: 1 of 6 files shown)


Shnatsel (approved):
Thank you for the PR!

This seems to fix the issue, but I believe a cheaper formulation is possible. Instead of widening, shifting, masking, and packing, mask each byte before shifting:

```rust
let mask_byte = 0xff_u32.wrapping_shr(shift) as i8;
let byte_mask = set1_epi8(mask_byte);
let result = sll_epi16(and(value, byte_mask), shift_count);
```

Clearing the upper `shift` bits of every byte prevents bits crossing between adjacent bytes during the 16-bit shift. This eliminates widening and packing entirely.

I like the refactoring in https://github.com/danderson/fearless_simd/commit/cc39d001f26f6805859018b83c686df62a17f119 and I'd be happy to merge that. Please open a PR for that. I can then do additional optimization work on top of the refactored state.

danderson:
Ah yes, masking before the shift is indeed much better. I sent #290 with the refactor. If you like I can also implement the performance optimization later today.

Another potential optimization on Avx512 could be using the GFNI affine transform instruction, which can do 8-bit native shifts in a single instruction. One question for that is instruction timing, I think GFNI takes a few cycles to execute so it's possible it's not faster than the multi-instruction and+sll form.

GFNI was also not available on early AVX512 CPUs, but for the CPUs that fearless_simd supports I think it's always available.

Shnatsel:
I'll handle optimizing SSE4 and AVX2 myself. I'd appreciate help with AVX-512.

llvm-mca should be able to tell which form is cheaper. It's important to measure on several representative processors - ice lake, sapphire rapids, zen 4, zen 5.

danderson:
ok, I'll hack on an avx512 version later. I haven't used llvm-mca before so it might take me a bit longer to make the PR, but I'll make sure to check different silicon versions.

Issue #289, opened 2026-07-28
i8 shl produces incorrect result


danderson:
Reproducer:

```
i8x64::splat(s, 1) << 7 == i8x64::splat(s, -128)
```

Works on the fallback impl, but results in 127 on sse2/sse4/avx2/avx512.

The result looks suspiciously like saturation arithmetic to me again, as in #287 but keeping the sign bit unchanged. I haven't dug into the Shl implementation for the signed types yet, just filing this so I don't forget.

danderson:
Hmm, I think the issue there is that, if we're trying to match the semantics of rust primitives, shl on signed types should be implemented exactly like shl on unsigned types. The current generated implementation adds a bunch of logic to handle sign extension and sign-aware packing, which is necessary for shr but does the wrong thing on shl.

I think the minimum necessary change is to use an unsigned pack instruction for shl on signed types. A more invasive change would be factoring out the 8b shift impls into a pair of helpers, one for shl and one for shr, because their logic is diverging enough that it's probably easier to read as two separate fns.

Hacking on this now to see if I can get a nice fix, but I'll be asleep soon so if someone else beats me to it, that's fine.

danderson:
https://github.com/linebender/fearless_simd/commit/b53d28f16f08ac089d7714b70311c2cd54515493 is the minimal fix that makes the new regression test pass. It depends on the fix in #288 to not produce a different incorrect result, so I'll PR this (or a larger refactor) once the first one is merged.

danderson:
And https://github.com/danderson/fearless_simd/commit/cc39d001f26f6805859018b83c686df62a17f119 is the refactor that splits out the 8-bit shr/shl helpers, and cleans up the generated code. Same as the other, I can PR this one after #288 is dealt with.

Shnatsel:
Fixed in https://github.com/linebender/fearless_simd/pull/290

PR #290, opened 2026-07-28
Fix byte-wise shl on i8 to match rust semantics

danderson:
And factor out the x86 8-bit shift kernels into helpers, since their logic is diverging between left/right shifts.

Updates #289 

---

The main logic change in generated code is the use of `packus` instead of `packs` for signed shl implementations. The other diffs are just reorganizing of the shift+and chain into a more compact form.


Shnatsel (approved):
I've already looked at the diff when this refactor was mentioned in #288, looks good

danderson:
pushed update to fix rustfmt CI failure.

PR #291, opened 2026-07-28
Faster 8-bit full-vector shifts

Shnatsel:
I had GPT-5.6 Sol suggest various formulations of these operations and measure them on llvm-mca, then picked the best imlementation for each op.

On top of the usual unit tests I've added ~~(and then reverted)~~ random tests that generate 100k random vectors, apply every possible shift and compare the result against a scalar reference.

I've only measured the AVX-512 path on tigerlake since my llvm-mca is too old to have Zen 4/5 or Sapphire Rapids. @danderson  I'd appreciate modeling on other CPUs for AVX-512 or benchmarks from real algorithms using this.

### DIFF
--- a/fearless_simd_tests/tests/harness/ops/shr.rs
+++ b/fearless_simd_tests/tests/harness/ops/shr.rs
@@ -279,6 +279,61 @@ fn shr_u8x32<S: Simd>(simd: S) {
 }
 
 #[simd_test]
+#[ignore] // Stress test: run with `cargo test --release shr_8bit_random -- --ignored`.
+fn shr_8bit_random<S: Simd>(simd: S) {
+    let mut rng = fastrand::Rng::with_seed(1337);
+
+    for iteration in 0..100_000 {
+        let mut unsigned = [0_u8; 64];
+        rng.fill(&mut unsigned);
+        let signed = unsigned.map(|value| value as i8);
+
+        let unsigned_16 = u8x16::from_slice(simd, &unsigned[..16]);
+        let unsigned_32 = u8x32::from_slice(simd, &unsigned[..32]);
+        let unsigned_64 = u8x64::from_slice(simd, &unsigned);
+        let signed_16 = i8x16::from_slice(simd, &signed[..16]);
+        let signed_32 = i8x32::from_slice(simd, &signed[..32]);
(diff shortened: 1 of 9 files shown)


Shnatsel:
This is what GPT-5.6 came up with for shifts using avx512gfni: https://gist.github.com/Shnatsel/091fa67291157abdeb581210579f2754

llvm-mca shows 3x better throughput for gfni on tigerlake, 2.5x better throughput on sapphire rapids. Zen4 is tied for left shift: it's much better at handling the current implementation for unsigned shifts and achieves the gfni level of throughput on the current implementation. For signed right shift gfni is worse on Zen4.

The latency for gfni is worse across the board:

CPU | Current u8 >> | GFNI u8 >> | Current i8 >> | GFNI i8 >>
-- | -- | -- | -- | --
Tiger Lake, 512-bit | 11 cycles | 13 | 11 | 14
Sapphire Rapids, 512-bit | 13 cycles | 17 | 13 | 19
Zen 4/5, 512-bit | 8 cycles | 9 | 8 | 10

GFNI is better for register pressure, with only 2 live registers instead of 3 or 4 on the other paths.

On the whole it seems like the better implementation depends on both the algorithm (throughput vs latency) and the CPU model. So I'm not seeing a strong case for switching to gfni.

danderson (approved):
Looks good. I agree that given the relative numbers, using GFNI isn't worth it. avx512 doubled the number of vector registers, so I don't think register pressure is something to worry about here. Code that really needs to use every last register with perfect optimization will have to write a custom kernel for the operation anyway to get consistent results.

fearless_simd_tests/tests/harness/ops/shr.rs
```diff
@@ -278,6 +278,61 @@ fn shr_u8x32<S: Simd>(simd: S) {
     assert_eq!(result.as_slice(), expected.as_slice());
 }
 
+#[simd_test]
+#[ignore] // Stress test: run with `cargo test --release shr_8bit_random -- --ignored`.
+fn shr_8bit_random<S: Simd>(simd: S) {
```
LaurenzV, line 334: Maybe we should move this to the very bottom so it's separate from the normal tests?

LaurenzV (approved):
Haven't thought this through, mostly relying on our tests and your experiments here.

PR #293, opened 2026-07-29
Allow safely creating SIMD level tokens from #[target_feature] functions

Shnatsel:
An updated version of #223

Also adds tests that verify that tokens can be safely made inside `kernel!` macros to verify that the target_feature annotations match.

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -23,6 +23,8 @@ pub struct Avx2 {
 impl Avx2 {
     #[doc = "Create a SIMD token proving that the x86-64-v3 features are available."]
     #[doc = r""]
+    #[doc = r" Most users should safely obtain the token from [`Level::new`] instead of calling this function."]
+    #[doc = r""]
     #[doc = r" This function can be called without an `unsafe` block from a function"]
     #[doc = r" with all the required target features enabled via the `#[target_feature]` annotation."]
     #[doc = r""]
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -28,9 +28,9 @@ impl Avx2 {
     #[doc = r""]
     #[doc = r" # Safety"]
     #[doc = r""]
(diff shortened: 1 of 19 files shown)


fearless_simd/src/generated/avx2.rs
```diff
@@ -21,15 +21,19 @@ pub struct Avx2 {
...
     #[doc = r" # Safety"]
     #[doc = r""]
-    #[doc = r" The `fxsr`, `avx2`, `bmi1`, `bmi2`, `cmpxchg16b`, `f16c`,"]
-    #[doc = r" `fma`, `lzcnt`, `movbe`, `popcnt`, and `xsave` CPU features"]
-    #[doc = r" must be available."]
+    #[doc = "When invoking this function through an `unsafe` block, the caller must ensure that the current CPU supports `avx2`, `bmi1`, `bmi2`, `cmpxchg16b`, `f16c`, `fma`, `fxsr`, `lzcnt`, `movbe`, `popcnt`, `xsave`."]
     #[inline]
-    pub const unsafe fn new_unchecked() -> Self {
+    #[target_feature(enable = "avx2,bmi1,bmi2,cmpxchg16b,f16c,fma,fxsr,lzcnt,movbe,popcnt,xsave")]
+    pub const fn new_unchecked() -> Self {
```
DJMcNab: Does the name still need to be `new_unchecked`?

DJMcNab (approved):
Looks good to me! It's good to see it taken over with better results - the tests are a fantastic idea. I don't think its possible to test the other way around though, right?

I'm not overly happy with the name `new_unchecked` for the methods, but it doesn't block.

fearless_simd/src/generated/avx2.rs
Shnatsel: Yeah I'm not thrilled about `new_unchecked()` either. Simply `new()` feels too low-hanging, doesn't communicate it's not very commonly used, but we have the docs pointing to `Level::new()` so it should be fine? We could also call it `assume_supported()`.

Shnatsel: I've renamed the function to `assume_supported()` after sleeping on it for a few days.

PR #296, opened 2026-07-31
Add BITS constant to SimdElement

danderson:
While writing bitpacking routines, this was a blocker for defining generic helpers: given just an `impl SimdInt<S>`, afaict there's no way to find out the bit width of elements, which is necessary in bitpacking for stuff like calculating carry shifts when a packed value spans words. This PR adds a `BITS` constant to `SimdElement`, mirroring the constant of the same name that's defined on primitive types.

### DIFF
--- a/fearless_simd/src/traits.rs
+++ b/fearless_simd/src/traits.rs
@@ -131,57 +131,47 @@ pub trait SimdElement: Copy + Seal {
     type Mask: SimdElement<Mask = Self::Mask>;
 
     /// The size of an element in bits.
-    const BITS: usize;
+    const BITS: usize = size_of::<Self>() * u8::BITS as usize;
 }
 
 impl SimdElement for f32 {
     type Mask = i32;
-    const BITS: usize = 32;
 }
 
 impl SimdElement for f64 {
     type Mask = i64;
-    const BITS: usize = 64;
 }
 
 impl SimdElement for u8 {
     type Mask = i8;
-    const BITS: usize = 8;
 }
 
 impl SimdElement for i8 {
     type Mask = Self;
-    const BITS: usize = 8;
 }
 
 impl SimdElement for u16 {
     type Mask = i16;
-    const BITS: usize = 16;
 }
 
 impl SimdElement for i16 {
     type Mask = Self;
-    const BITS: usize = 16;
 }
 
 impl SimdElement for u32 {
     type Mask = i32;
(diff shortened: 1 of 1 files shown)


Shnatsel:
Actually you can, that's `std::mem::size_of::<T>()`, but this certainly could be made more convenient and discoverable.

I've refactored the implementation to compute the width instead of duplicating it everywhere via an associated constant.

I'm curious to see the code where this crops up. This makes me wonder if we should add optional [num-traits](https://crates.io/crates/num-traits) support, or add a whole lot of generic bounds for ops like addition etc. But this looks like a small and uncontroversial improvement, and will be needed regardless since `f32::BITS` is nightly-only. Thanks!

Shnatsel (approved):

danderson:
Hmm, I thought I'd run into some issue with size_of in generic context, where it couldn't satisfy the Sized bound, but I've gone through a dozen iterations of this code so I might just be confused.

More bounds in places (especially more bounds on the associated types that exist) would be useful, but it's also a larger set of changes and I'm not yet sure if they benefit a lot of code or not.

The usage is more bit-packing code, I'm trying out different code layouts and helpers to make the generated code more compact and reusable for other more complex packing ops. https://gist.github.com/danderson/a26121c47f07a934a1cd4a491e6ef326 is the helper where this came up, which executes one step of packing: load a vector from the input, mask the low bits to pack, shift them into place and OR them into the output word. For steps where the value spans output words, it also returns a carry value (the start of the next output vector) containing the bits that were shifted out of the left of the output word. I needed the element bit width to compute the shift for the carry vector.

The helper is admittedly a bit convoluted, but it makes the generated code for each packing width very regular and easy to follow, and empirically the compiler has no problem eliminating the carry computation completely for steps where it's not needed, so the generated code is equivalent to a tight handwritten kernel.

Shnatsel:
Thank you for the context!

Now that I'm looking at [`BITS` in the standard library](https://doc.rust-lang.org/stable/std/primitive.i8.html#associatedconstant.BITS), it returns a `u32`, while this code adds it as `usize`. Is there a good reason for that or should we match std here?

FWIW `Sized` is an autotrait, so it doesn't need to be spelled out explicitly. This code compiles fine for me:

```rust
fn generic_element_size<S: Simd, V: SimdBase<S>>(_: V) -> usize {
    std::mem::size_of::<V::Element>()
}
```

Issue #297, opened 2026-08-03
ops with inconsistent type coverage


Shnatsel:
- widen is only implemented for u8 vectors, narrow only for u16 vectors. All other integer widths and floats are missing. The API also needs work as discussed [here](https://github.com/linebender/fearless_simd/pull/127#issuecomment-3539665349).
- `load/store_interleaved_128_*` are implemented for `f32` and unsigned integers, not for signed and f64 (weirdly, seems like there are no blockers by now)
- Conversions between `f64`  and `i64`/`u64` are missing because they need special handling before AVX-512

LaurenzV:
Only implemented the interleaved ones for the ones I needed in Vello, hence why some are missing.

Shnatsel:
All 3 PRs to fix this are now merged, closing as fixed.

PR #298, opened 2026-08-03
API improvements to 4-way interleaved load/store

Shnatsel:
Change `load_interleaved_128` that returns a 512-bit vector to `load_four_interleaved` that returns four 128-bit vectors. 

This maps to the real-world usage better (in Vello, all uses of this function save one go through manual split/combine first), eliminates pointless split/combine in x86 implementations, and most importantly allows us to add wrappers for vld3q later which would perform the same operation for RGB rather than RGBA.

Here's how this change affects vello: https://github.com/linebender/vello/pull/1796

Also supports interleaved load/store for all widths, addressing a part of #297

The implementations are untouched in this PR save for removing redundant split/combine. AVX2 leaves performance on the table because it uses the sse2 unpack tree for most types, but I don't want to write a bespoke implementation for what is just a swizzle. Instead let's merge #276, then I'll add a relaxed version for x86 that doesn't zero out-of-bounds indices (update: that's #304), and then we'll just lower interleaving into that on x86 since it would call literally all the same intrinsics.


danderson (approved):
Looks reasonable! I can't comment on Vello's use case obviously, but the linked draft looks like a nice reduction in conversion boilerplate.

fearless_simd/src/generated/neon.rs
```diff
@@ -566,6 +566,32 @@ impl Simd for Neon {
...
+    fn load_four_interleaved_f32x4(self, src: &[f32; 16usize]) -> [f32x4<Self>; 4usize] {
+        unsafe {
+            let native = vld4q_f32(src.as_ptr());
+            [
+                native.0.simd_into(self),
+                native.1.simd_into(self),
+                native.2.simd_into(self),
+                native.3.simd_into(self),
+            ]
+        }
```
LaurenzV: Can we limit the unsafe blocks to the actually unsafe parts? Same for the other locations.

LaurenzV (approved):
Looking forward to better AVX2 impls and RGB interleaving!

PR #304, opened 2026-08-04
Add relaxed swizzle_dyn

Shnatsel:
A variant of #276 that produces implementation-defined values instead of 0 on out-of-bounds indices. Only native-size vectors get the relaxed behavior; swizzle_dyn_precise is reused for 2x the native vector width because zeroing allows cheaply joining the halves.

Improves performance on x86 and fallback compared to swizzle_dyn_precise. Reduces register pressure, which is important on x86. WASM also improves for 512-bit vectors where it routes through fallback.

Part of #29

### Performance details

Most AVX-512 machines get -1 cycle of latency (icelake-server 4->3 cycles, zen4 6->5 cycles), except for Sapphire Rapids where this doubles throughput and halves the latency.

AVX2 ~~is unchanged~~ slightly improves throughput at the cost of worse latency (18->21 cycles on Haswell and Skylake, 20->21 on Zen3) and uses one less register.

SSE4.2 results:

Intel [Westmere](https://en.wikipedia.org/wiki/Westmere_(microarchitecture)) gets double the throughput and latency drops from 3 to 1 cycle.

AMD Jaguar (btver2) drops latency from 4 to 2 cycles, throughput improves by 50%.

SSE4.2-only Intel [Tremont](https://en.wikipedia.org/wiki/Tremont_(microarchitecture)) launched as recently as 2021 improves latency from 7 to 5 cycles, throughput is unchanged.

Fallback (also used by SSE4.2 and WASM for 512-bit vectors) improves 50% in both latency and throughput.


LaurenzV (approved):
Only took a cursory look + AI review!

Shnatsel:
The implementation is straightforward and covered by tests so I'm not too worried about correctness.

The design questions here are:

1. Is this desirable, considering uneven and hardware-dependent gains? My call is yes, since it doesn't add much complexity (no emulation etc).
2. Should this be the default (swizzle_dyn) or should the zero-on-out-of-bounds version be called `swizzle_dyn` and this be called something else? `swizzle_dyn_precise` follows the existing fearless_simd convention but `swizzle_dyn` would be a drop-in replacement for `std::simd`

We can easily change point 2 after the fact so I'm going to go ahead and merge this.

PR #306, opened 2026-08-04
Remove WithSimd trait

### LINKED ISSUE #161: Is `WithSimd` still necessary?
I'm starting to look at the API surface, and I'm now wondering if `WithSimd` gives us anything that the `dispatch!` macro doesn't. The one method that actually makes use of `WithSimd`, `Level.dispatch`, just defers to the `dispatch!` macro internally anyway:

```rs
    pub fn dispatch<W: WithSimd>(self, f: W) -> W::Output {
        dispatch!(self, simd => f.with_simd(simd))
    }
```

I thought it might be useful for adding extra type parameters to dispatched function calls or something, but it appears to be equivalent in power to `dispatch!` while being a lot more annoying to use. Should we just remove `WithSimd`?

Shnatsel:
This appears to be a remnant of pulp's design, and it just delegates to the `dispatch!` macro internally anyway.

In #161 it was agreed that it is useless all the way back in 2025, so let's just remove it before we ship 1.0.


DJMcNab (approved):

DJMcNab:
Fwiw, I do think that `VectorisedFnOnce` is a unique capability provided by that type. But if someone needs it, they can write their own version, which has the advantage of the argument types being easier to reason about.

PR #308, opened 2026-08-05
Move ops shared between SimdInt and SimdFloat to SimdBase

### LINKED ISSUE #301: `SimdBase` could have many ops from `SimdInt` and `SimdFloat`
In the past `SimdBase` was also used for masks so it was very restricted in the ops it could implement. However, after masks became opaque, only `SimdInt` and `SimdFloat` inherit from `SimdBase`, and they have many shared ops between them. Those shared ops could be moved to `SimdBase` to allow abstacting over int and float vectors in generic code.

Shnatsel:
Closes #301

Not part of this PR, but could be: min/max, min_precise/max_precise. Also we have no generic numeric conversions between ints and floats, only dedicated to_int/to_float, not sure if those are useful.

### DIFF
--- a/fearless_simd/src/generated/simd_trait.rs
+++ b/fearless_simd/src/generated/simd_trait.rs
@@ -3611,6 +3611,28 @@ pub trait SimdBase<S: Simd>:
     fn swizzle_dyn_within_blocks(self, indices: impl SimdInto<Self::Bytes, S>) -> Self;
     #[doc = "Dynamically swizzle this vector's bytes across the whole vector.\n\nThe `indices` operand is a same-width byte vector. For each output byte, index values within the vector's byte length select the corresponding byte from the input vector. Out-of-range indices produce zero bytes."]
     fn swizzle_dyn_precise(self, indices: impl SimdInto<Self::Bytes, S>) -> Self;
+    #[doc = "Compare two vectors element-wise for equality.\n\nReturns a mask where each logical lane is true if the corresponding elements are equal, and false if not."]
+    fn simd_eq(self, rhs: impl SimdInto<Self, S>) -> Self::Mask;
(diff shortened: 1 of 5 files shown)


fearless_simd/src/generated/simd_trait.rs
```diff
@@ -3611,24 +3623,34 @@ pub trait SimdBase<S: Simd>:
     fn swizzle_dyn_within_blocks(self, indices: impl SimdInto<Self::Bytes, S>) -> Self;
     #[doc = "Dynamically swizzle this vector's bytes across the whole vector.\n\nThe `indices` operand is a same-width byte vector. For each output byte, index values within the vector's byte length select the corresponding byte from the input vector. Out-of-range indices produce zero bytes."]
     fn swizzle_dyn_precise(self, indices: impl SimdInto<Self::Bytes, S>) -> Self;
+    #[doc = "Compare two vectors element-wise for equality.\n\nReturns a mask where each logical lane is true if the corresponding elements are equal, and false if not."]
+    fn simd_eq(self, rhs: impl SimdInto<Self, S>) -> Self::Mask;
+    #[doc = "Compare two vectors element-wise for less than.\n\nReturns a mask where each logical lane is true if `self` is less than `rhs`, and false if not."]
+    fn simd_lt(self, rhs: impl SimdInto<Self, S>) -> Self::Mask;
```
danderson, line 3629: Maybe note that the comparisons are not guaranteed to be a total order? Maybe that's too obvious but IME it's a common footgun when doing generic stuff that includes floats.

danderson (approved):

LaurenzV (approved):
Makes sense!

fearless_simd/src/generated/simd_trait.rs
LaurenzV: I think that's just the same comment as before though, no? If we want to change it, let's do it in a follow-up, so code moves and semantic changes are kept separate.

PR #310, opened 2026-08-05
Add f64 <-> i64/u64 conversions

Shnatsel:
Part of #297

The combined diff is large but it's not as scary as it looks. This is what this PR does:

- Add 64-bit type conversions, with lowering to hardware instructions where available, and scalar fallback otherwise (482faea2891da061fb336e1ec888b903e0ee164d)
    - Faster AVX2 formulations are [possible](https://chatgpt.com/share/6a735624-05dc-83ed-8549-dccae7955569) but that's for a follow-up PR to keep the size of this one manageable
- Optimize AVX-512 Float->Unsigned conversion for both 32 and 64 bits (39adc0e9870d2763637b48699b0a4c0f364d479d)
- The rest is generator refactoring for readability with no impact on the generated code, confined to its own separate commits


fearless_simd_gen/src/mk_x86.rs
```diff
@@ -25,6 +25,12 @@ pub(crate) enum X86 {
     Avx512,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
```
danderson, line 28: Maybe Debug too? Only matters for println debugging though.

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -2950,65 +2950,64 @@ impl X86 {
...
+                    vec_ty.n_bits(),
+                );
                 let blend = avx512_mask_blend_intrinsic(&target_ty);
                 let set1_float = set1_intrinsic(vec_ty);
                 let set1_int = set1_intrinsic(&target_ty);
                 let set0_int =
                     intrinsic_ident("setzero", coarse_type(&target_ty), target_ty.n_bits());
+                let upper_bound = match scalar_bits {
+                    32 => quote! { 2147483648.0 },
+                    64 => quote! { 9223372036854775808.0 },
```
danderson: Does this literal survive compilation unchanged? In rust playground, loading it into an f64 rounds it up to 9223372036854776000.0

danderson (approved):
Reviewed commit by commit, which makes the refactors easy to reason about because they don't alter codegen output or tests.

fearless_simd_gen/src/mk_x86.rs
Shnatsel: That's a formatting artifact, the bitpattern is the same. Try this:

```rust
let x = 9223372036854775808.0_f64;

assert_eq!(x.to_bits(), 0x43e0_0000_0000_0000);
assert_eq!(x, 2.0_f64.powi(63));
```

I've added a comment on that, and then looked at it and added comments on the rest of the code as well because if you didn't write it it's probably completely inscrutable

fearless_simd_gen/src/mk_x86.rs
Shnatsel: It's an internal generator struct, we'll add it when it comes up

fearless_simd_tests/tests/harness/ops/cvt_u64.rs
LaurenzV, line 1: I think having one more test with some more "normal numbers" (i.e. just numbers between -200 to 200) instead of only testing for those "edge cases" would be nice. Same for the other modules.

LaurenzV (approved):
Only took a cursory look.

fearless_simd_tests/tests/harness/ops/cvt_u64.rs
Shnatsel: To get the normal codepaths exercised across all lanes, right? Sure, I'll add some.

Issue #316, opened 2026-08-06
Fallback level is compiled on platforms where it's not needed


Shnatsel:
It is redundant both on x86 and Aarch64 since we have a better baseline available.

On x86 this costs 0.5s out of 3.4s total build time. On Aarch64 this costs 0.7 seconds out of 1.8s total build time, since it compiles only two levels (fallback and NEON).

The other levels have to exist in some form for use in `kernel!` (they can be aliases to higher levels) but `Fallback` can be just eliminated unless `force_fallback_support` feature is enabled because you generally don't need it for anything.

Issue #113, opened 2025-11-11
Make the generic split/combine operations into default trait methods?


valadaptive:
I assume there's some reason this isn't done already. The "generic" implementations for SIMD operations split the input in two, process the two halves, and then combine them.

Currently, the exact same split/combine function body is generated once for every SIMD implementation. For instance, if you grep the codebase for:

```rust
    #[inline(always)]
    fn div_f64x4(self, a: f64x4<Self>, b: f64x4<Self>) -> f64x4<Self> {
        let (a0, a1) = self.split_f64x4(a);
        let (b0, b1) = self.split_f64x4(b);
        self.combine_f64x2(self.div_f64x2(a0, b0), self.div_f64x2(a1, b1))
    }
```

This function body appears *four* times: in the AVX2, NEON, WebAssembly, and fallback implementations.

Instead, these generic operations could be default implementations on the `Simd`  trait.

Shnatsel:
Done in https://github.com/linebender/fearless_simd/pull/318

PR #318, opened 2026-08-06
Further build time reductions

Shnatsel:
~~Stacked on top of #317~~

In total this brings down x86 build time on my machine from 3s to 2.2s, on par with `wide`. This method has hit diminishing returns, but with #316 we can beat `wide` on build time.

Best reviewed commit by commit. Each one explains what it does and what the effects are.

The first commit fulfills #113


LaurenzV (approved):
Nice!

PR #314, opened 2026-08-06
Release 0.7.0

Shnatsel:
Update changelog and bump version

PR #320, opened 2026-08-07
Only compile Fallback level on targets without a better baseline

### LINKED ISSUE #316: Fallback level is compiled on platforms where it's not needed

Shnatsel:
It emitted no LLVM IR, but it was still being type- and borrow-checked despite never being called.

When applied on top of #318 reduces x86 clean build time by 0.2s (from 2.15s to 1.95s) and Aarch64 build time by 0.3s (from 1.25s to 0.95s).

Fixes #316

There are two instances of this cfg, which is as low as I could get it. Could be brought down to one with a build script but that would negate the build time gains.


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ You can find its changes [documented below](#060-2026-07-10).
 
 - The `new_unchecked()` function on SIMD level tokens such as `Avx2` has been renamed to `assume_supported()` and is now safe to call from contexts that already contain the appropriate `#[target_feature]` annotations. Functions without such annotations can still call `assume_supported()` with an `unsafe` block. ([#293][] by [@Shnatsel][])
 - On x86_64 targets with static SSE2 support, `Level::baseline()` now returns `Sse2` instead of `Fallback`. ([#270][] by [@Shnatsel][])
+- The scalar `Fallback` backend and `Level::Fallback` variant are no longer compiled when the target has a better ambient SIMD baseline (e.g. SSE2 on x86, NEON on Aarch64). The `force_support_fallback` feature continues to make them available for testing. `disable_dispatch_sse2` no longer disables SSE2 if it is the baseline level.
```
DJMcNab, line 26: This should have the pr number reference (by the way, I've been working on tooling to make our changelog management better - see https://github.com/DJMcNab/release_eng/blob/main/gazeto/README.md - which I'd appreciate if you had any feedback on).

fearless_simd/src/lib.rs
```diff
@@ -366,6 +368,19 @@ pub enum Level {
     /// Scalar fallback level, i.e. no supported SIMD features are to be used.
     ///
     /// This can be created with [`Level::fallback`].
```
DJMcNab, line 375: Let's document that matching against this directly is discouraged/fragile, and that using `is_fallback` is the correct way to achieve that.

(If this is already documented on the item directly, then that might be unnecessary)

DJMcNab (approved):
Thanks!

This does leave me with a niggling doubt about whether we should return to doing this more generally (e.g. separating the "proof types" used for `kernel` from the `Simd` impls to improve build time). But I suspect the cases where that would actually come up are so niche it's not worth worrying about in practise.

fearless_simd/src/lib.rs
Shnatsel: Good call! I meant to do that but forgot.

CHANGELOG.md
Shnatsel: I feel my current workflow of having an LLM go through the git history and write a draft, then editing the PR summaries manually works well. See https://github.com/linebender/fearless_simd/pull/314 for what the result looks like.

And frankly I don't want to be beholden to even more inflexible tools than this repository already is.

Shnatsel:
> This does leave me with a niggling doubt about whether we should return to doing this more generally (e.g. separating the "proof types" used for `kernel` from the `Simd` impls to improve build time). But I suspect the cases where that would actually come up are so niche it's not worth worrying about in practise.

I thought about this and came to the same conclusion.

There is however a related use case: when a `kernel!` macro accepts e.g. `AVX2`, this could in theory be satisfied with any level that is AVX2 **or higher.** So if the host actually has AVX-512 it can give `kernel!` calls e.g. better AVX-512 loads if there's `-C target-cpu=icelake-server` active. But I don't think this is very useful in practice, and LLVM should optimize trivial things like loads anyway, while more complex ops inside `kernel!` are not that common.

PR #321, opened 2026-08-07
Expose interleaved loads/stores on vector types they operate on

### LINKED ISSUE #315: Interleaved loads/stores not exposed on the vector types they operate on
Right now the only operation that is available through the `Simd` trait only is `load_four_interleaved`.

I'd like to be able to add it to a trait implemented for 128-bit vector types, and put a warning on the `Simd` trait that it's very verbose and people shouldn't be browsing its docs directly and instead looking at docs for `u32x4`, `i32x4`, `f32x4`.

Shnatsel:
Fixes #315

This was the last method missing from documentation listings for types such as `u32x4`, so add a notice to SIMD level tokens pointing people at doc pages for types like `u32x4` for browsing the documentation.


fearless_simd/src/traits.rs
```diff
@@ -261,6 +261,40 @@ pub trait SimdCvtFloat<T: Seal>: Seal {
...
+/// These operations load/store up to 512 bits in one go, and loading into wider vectors
+/// usually degrades performance by causing more register pressure.
+///
+/// If processing in wider vectors is desirable, combine the 128-bit vectors into larger ones
+/// and process them together as a single vector. This avoids issues with register pressure.
+pub trait SimdInterleaved<S: Simd>: SimdBase<S> {
+    /// The exact-sized flat scalar array exchanged by four-way interleaved operations.
+    ///
+    /// This contains four times as many elements as one vector.
+    type FourInterleavedArray: Copy + AsRef<[Self::Element]> + AsMut<[Self::Element]>;
```
LaurenzV: We need this because associated constants in Rust are pretty limited right now, right? I remember dealing with similar issues before...

fearless_simd/src/traits.rs
```diff
@@ -261,6 +261,40 @@ pub trait SimdCvtFloat<T: Seal>: Seal {
...
+
+    /// Load four 128-bit vectors from an array with 4-way interleaving
+    ///
+    /// This is useful e.g. in image processing to turn interleaved RGBA pixels into vectors of each color component.
+    ///
+    /// For example, with 32-bit lanes, memory laid out as
+    /// `[r0, g0, b0, a0, r1, g1, b1, a1, r2, g2, b2, a2, r3, g3, b3, a3]` loads as
+    /// `[[r0, r1, r2, r3], [g0, g1, g2, g3], [b0, b1, b2, b3], [a0, a1, a2, a3]]`.",
```
LaurenzV:
```suggestion
    /// `[[r0, r1, r2, r3], [g0, g1, g2, g3], [b0, b1, b2, b3], [a0, a1, a2, a3]]`."
```

fearless_simd/src/traits.rs
LaurenzV: Will we also add `ThreeInterleaved` and `TwoInterleaved` in the future? Have you considered whether it fits into this? Just wondering

LaurenzV (approved):

fearless_simd/src/traits.rs
Shnatsel: There isn't a way to say "this is an array of length N" for use in generic code, mostly because there's no "array" trait in std.

Besides, for these ops the size of the read in bytes is constant, but the number of elements varies depending on the vector type. So I'm not sure how to even express associated constants for a given scalar type.

Shnatsel: I have considered that and it should. I don't see why not. I can add TwoInterleaved to see how it slots in, that should have straightforward lowerings into the existing interleave/deinterleave.

The trickier question is can generic code even use this as-is without extra bounds such as `TryFrom<&[Self::Element]>` or some such. But that should be a non-breaking change to add in the future.

LaurenzV: Don’t have to try it now, just wanted to ask!

Shnatsel: I've reworked the API to remove the associated type by using slices as input and output.

It also allows end-to-end usage in generic contexts, which wasn't possible previously due to missing TryFrom (the associated type wasn't possible to create from a slice).

Please take another look.

Shnatsel: The updated slice-based API is definitely future-proof now.

LaurenzV: Hmm I'm a bit concerned that the `try_into` might not always be optimized away... To be clear I wasn't criticizing the associated type, it was just a drive-by comment! But if you can confirm that it doesn't seem to influence the ASM then it's also fine from my side, up to you.

Shnatsel: In generic code a `try_into()` needs to appear *somewhere*: you can't use `.as_chunks()` on an associated constant, so you need `.chunks_exact()` plus a `.try_into()` at *some* point. And this is `#[inline(always)]` so it has the best change of being optimized away we can give it.

I think this API is much better for use in generic code. Backend methods that load from arrays directly are still available, and if we find a generic array-based API we can add in a non-breaking change.

LaurenzV: Right, I forgot that this is for the generic code. Sorry, as you can see I haven’t thought about most things very deeply. 😅 Mostly just want to unblock you.

Shnatsel: You're still asking very good questions. It was your question that made me re-evaluate whether I can remove complex associated types, and turns out I can and it makes the API better!

Issue #18, opened 2025-06-25
Remove `msub` operation


ajakubowicz-canva:
Context: https://github.com/linebender/fearless_simd/pull/14#discussion_r2167634297

LaurenzV:
First we should verify that the negating thing indeed leads to efficient code generation.

awxkee:
FWIW LLVM can recognize all kinds of FMA variants starting from Rust 1.45 ( for x86_64 ):

https://godbolt.org/z/f3G4WEM6P

https://godbolt.org/z/61jYsxr83

raphlinus:
That matches my experiment with achieving [vfms](https://doc.rust-lang.org/stable/core/arch/aarch64/fn.vfmsq_f32.html) on Neon: https://rust.godbolt.org/z/YTe4bcz4n

Maybe there's more validation to be done, but from my perspective I think we can now confidently say that LLVM is able to optimize this particular class of intrinsics. Much less so for other things I'm concerned about, like round/convert combinations, so I think more investigation is needed on those.

Shnatsel:
I looked into this, and turns out LLVM 22 fails to recognize `mul_add(a, b, -c)` as `mul_sub(a, b, c)` for f32x4 on AVX2, even though it does it just fine for `f64x2` and `f32x8`.

And this was tested on isolated harnesses where the only op is FMA. When inlined into larger functions, the optimization may degrade further.

It seems this stuff is just fragile, and we don't really have a way around explicit ops if we want them to optimize well - and we do, since they're part of very hot kernels like FFT.

PR #323, opened 2026-08-09
Add mul_add_precise() which guarantees single-rounding precision

Shnatsel:
This is useful for building algorithms with guaranteed precision. For example, most of the literature on getting accurate SIMD trigonometry requires precise fused multiply-add.

Lowers into:

- accurate FMA on WASM even with relaxed SIMD
- Custom scalar FMA based on a formally verified algorithm from a [2008 paper]( https://guillaume.melquiond.fr/doc/08-tc.pdf) to work around [a bug](https://github.com/rust-lang/compiler-builtins/issues/1262) in Rust libm and musl libc
- SSE4.2 f32 emulates it on f64 SIMD vectors using the same algorithm from the same paper

The paper is "Emulation of FMA and correctly-rounded sums: proved algorithms using rounding to odd" by Sylvie Boldo and Guillaume Melquiond.

The SSE4.2 version is 5x faster than the libm scalar implementation, and 2.5x faster when the rare fixup branch always needs to be taken. On branches: Nehalem misprediction recovery is 17 cycles which amounts to a 40% throughput penalty, so this should never be worse than scalar even in the worst case for branch misprediction. Tremont timings are hard to find but it gets a very beefy predictor and its throughput is awful, so it shouldn't be any worse. Zen4 benchmarks confirm that worst case misprediction is roughly in line with scalar. A branchles formulation with more work in the hot path would be possible but with misprediction being rare, it isn't worth it in practice.

Also fixes `mul_sub()` on NEON which used to produce zero with an incorrect sign when all inputs are 1s with different signs. The fixed assembly has identical performance to the previous one on llvm-mca in both throughput and latency. It's always two instructions because NEON doesn't have `mul_sub` with the semantics we need, this PR only changes which two instructions are selected.

It is unfortunately not possible to remove the dedicated mul_sub operation and just write mul_add(a, b, -c) because LLVM does not optimize it reliably enough, see https://github.com/linebender/fearless_simd/issues/18

### DIFF
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -12,6 +12,30 @@ use crate::{
 };
 use core::arch::wasm32::*;
 use core::ops::*;
+#[inline(always)]
+#[allow(
+    dead_code,
+    reason = "Generated backends use different subsets of these helpers"
+)]
+fn scalar_mul_add_precise_f32(a: f32, b: f32, c: f32) -> f32 {
+    let product = (a as f64) * (b as f64);
+    let c = c as f64;
+    let mut sum = product + c;
+    if sum.is_finite() {
+        let virtual_sum = sum - product;
+        let residual = (product - (sum - virtual_sum)) + (c - virtual_sum);
+        let sum_bits = sum.to_bits();
+        if residual != 0.0 && sum_bits & 1 == 0 {
+            let corrected_bits = if sum.is_sign_negative() == residual.is_sign_negative() {
+                sum_bits.wrapping_add(1)
+            } else {
+                sum_bits.wrapping_sub(1)
+            };
+            sum = f64::from_bits(corrected_bits);
+        }
+    }
+    sum as f32
+}
 #[cfg(all(feature = "libm", not(feature = "std")))]
 #[allow(
     dead_code,
@@ -321,10 +345,10 @@ impl Simd for WasmSimd128 {
     #[inline(always)]
     fn mul_add_precise_f32x4(self, a: f32x4<Self>, b: f32x4<Self>, c: f32x4<Self>) -> f32x4<Self> {
         [
-            f32::mul_add(a[0usize], b[0usize], c[0usize]),
-            f32::mul_add(a[1usize], b[1usize], c[1usize]),
-            f32::mul_add(a[2usize], b[2usize], c[2usize]),
-            f32::mul_add(a[3usize], b[3usize], c[3usize]),
+            scalar_mul_add_precise_f32(a[0usize], b[0usize], c[0usize]),
+            scalar_mul_add_precise_f32(a[1usize], b[1usize], c[1usize]),
+            scalar_mul_add_precise_f32(a[2usize], b[2usize], c[2usize]),
+            scalar_mul_add_precise_f32(a[3usize], b[3usize], c[3usize]),
         ]
         .simd_into(self)
     }
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -340,6 +340,10 @@ impl Simd for WasmSimd128 {
         }
     }
     #[inline(always)]
+    fn mul_sub_precise_f32x4(self, a: f32x4<Self>, b: f32x4<Self>, c: f32x4<Self>) -> f32x4<Self> {
+        self.mul_add_precise_f32x4(a, b, -c)
+    }
+    #[inline(always)]
     fn floor_f32x4(self, a: f32x4<Self>) -> f32x4<Self> {
         f32x4_floor(a.into()).simd_into(self)
     }
@@ -2304,6 +2308,10 @@ impl Simd for WasmSimd128 {
         {
             self.sub_f64x2(self.mul_f64x2(a, b), c)
         }
+    }
+    #[inline(always)]
+    fn mul_sub_precise_f64x2(self, a: f64x2<Self>, b: f64x2<Self>, c: f64x2<Self>) -> f64x2<Self> {
+        self.mul_add_precise_f64x2(a, b, -c)
     }
     #[inline(always)]
     fn floor_f64x2(self, a: f64x2<Self>) -> f64x2<Self> {
--- a/fearless_simd/src/generated/wasm.rs
+++ b/fearless_simd/src/generated/wasm.rs
@@ -12,6 +12,82 @@ use crate::{
 };
 use core::arch::wasm32::*;
 use core::ops::*;
+#[cfg(all(feature = "libm", not(feature = "std")))]
+#[allow(
+    dead_code,
+    reason = "Generated backends use different subsets of these helpers"
+)]
+trait FloatExt {
+    fn floor(self) -> Self;
+    fn ceil(self) -> Self;
+    fn round_ties_even(self) -> Self;
+    fn fract(self) -> Self;
+    fn sqrt(self) -> Self;
+    fn trunc(self) -> Self;
+    fn mul_add(self, a: Self, b: Self) -> Self;
+}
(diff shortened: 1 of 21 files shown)


Shnatsel:
Pedantry for posterity:

Commit 33d9f98069a6ebee23e982ed56d915b2c40897f6 says Tremont (2021 Atom) improves in latency from 20 to 12, but that's a mistake: it improves in reciprocal throughput from 20 to 12, and in latency from 21 to 16. 

But this is already a very rarely taken branch so it doesn't really matter either way.

awxkee:
It seems emulation currently fails a bit https://gist.github.com/awxkee/afd3ee135602056a56806db9ddfcb9c9. For pair `(0x19ffe002, 0x1a001001)` it fails almost half of results

Shnatsel:
Oh! Good catch, thanks!

I've been wondering why the exhaustive test missed it, but in retrospect it wasn't exhaustive at all, I've been feeding the same value into all three lanes :facepalm: 

I'll rework this and also add far more in-depth random testing.

Shnatsel:
@awxkee I think you've just found a bug in libm :eyes: 

I'll investigate in more detail and report, but basically its software FMA uses the same ideas as this one and it seems to fail on your inputs in forced-software mode (hardware FMA disabled).

awxkee:
I strongly recommend checking glibc's or LLVM's libm, since the Rust implementation is ported from musl, and musl isn't the most robust one (and LLVM is the unreadable one). The typical FMA code seems to work great.

<details>
<summary>Code</summary>

```rust

#[inline]
pub fn soft_fma(x: f32, y: f32, z: f32) -> f32 {
    let xy = f64::from(x) * f64::from(y);
    let z = f64::from(z);
    let result = xy + z;
    let mut u = result.to_bits();
    if u & 0x0fff_ffff != 0 {
        return result as f32;
    }

    if u & 0x1000_0000 == 0 && (u >> 52) & 0x7ff > 1023 - 126 {
        return result as f32;
    }

    if result - xy == z && result - z == xy {
        return result as f32;
    }

    let neg = u >> 63 != 0;
    let err = if neg == (z > xy) {
        xy - result + z
    } else {
        z - result + xy
    };
    if neg == (err < 0.0) {
        u += 1;
    } else {
        u -= 1;
    }
    f64::from_bits(u) as f32
}
```
</details>

Shnatsel:
The musl implementation was also submitted to glibc: https://sourceware.org/pipermail/libc-alpha/2025-November/172449.html

So I don't even know what to use as the source of truth anymore.

WASM CI also fails so maybe our wasmtime is built with musl?

awxkee:
Rust have an unique 'rounding branch', it enters it on those values and fails. Therefore it seems it's a porting problem :)

WASM always use Rust standard libm as far as I understand. WASM itself doesn't have libm at all.

Shnatsel:
The wasmtime failure in CI is even more bizarre: 

> wasm trap: wasm `unreachable` instruction executed

Shnatsel:
> wasm trap: wasm `unreachable` instruction executed

Okay it's the same libm bug, it's just that the wasmtime runner doesn't report the assertion failure properly.

Shnatsel:
The same subnormals rounding bug is present in musl libc: https://git.musl-libc.org/cgit/musl/tree/src/math/fmaf.c?id=f21a96538f78fa8e2040831b4209b35f2fb581da

`if ((u.i & 0x1fffffff) != 0x10000000) /* not a halfway case */`

this only handles normal floats but not subnormals.

This paper has a fix - it provides the correct algorithm with a correctness proof in Coq: https://guillaume.melquiond.fr/doc/08-tc.pdf

@awxkee wanna report it upstream since it's your finding?

awxkee:
Nah, feel free to report

Shnatsel:
I've reported the bug to musl: https://www.openwall.com/lists/musl/2026/08/10/1

And opened a PR for Rust's libm.

And worked around it using the formally proven FMA formulation from the paper, using your implementation as a test oracle.

fearless_simd_tests/tests/harness/ops/mul_sub_precise.rs
```diff
@@ -0,0 +1,283 @@
+// Copyright 2026 the Fearless_SIMD Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use fearless_simd::*;
+use fearless_simd_dev_macros::simd_test;
+
+#[simd_test]
+#[ignore = "exhaustively checks all non-NaN f32 bit patterns"]
+fn mul_sub_precise_f32x4_exhaustive<S: Simd>(simd: S) {
```
LaurenzV, line 9: Can we move the exhaustive one to the bottom?

fearless_simd_tests/tests/harness/ops/mul_add_precise.rs
```diff
@@ -0,0 +1,606 @@
...
+        u += 1;
+    } else {
+        u -= 1;
+    }
+    f64::from_bits(u) as f32
+}
+
+#[simd_test]
+#[ignore = "exhaustively checks all non-NaN f32 bit patterns"]
+fn mul_add_precise_f32x4_exhaustive<S: Simd>(simd: S) {
```
LaurenzV, line 351: Also these two.

fearless_simd/src/generated/wasm.rs
```diff
@@ -12,6 +12,106 @@ use crate::{
 };
 use core::arch::wasm32::*;
 use core::ops::*;
+#[inline(always)]
+#[allow(
+    dead_code,
+    reason = "Generated backends use different subsets of these helpers"
+)]
+fn scalar_mul_add_precise_f32(a: f32, b: f32, c: f32) -> f32 {
```
LaurenzV, line 16: Not super urgent, but I think this (and the `FloatExt` trait) are duplicated in multiple files. Maybe there should be a util.rs file that contains those?

LaurenzV (approved):
Not gonna pretend I understand much of the implementation, but if you think this is important to have in the library, LGTM!

fearless_simd/src/generated/wasm.rs
Shnatsel: In this case not really because there are 2 instances of the plain function and 1 instance of it wrapped in `kernel!` which we need to keep separate. The `kernel!`-wrapped one is there for runtime detection of SSE2 on i586 targets which get x87 f32 arithmetic otherwise which is not IEEE 754 compliant and breaks the function's precision contract.

But beyond this helper - yes, there's probably something that could be moved into shared code.

fearless_simd_gen/src/mk_x86.rs
```diff
@@ -2401,9 +2406,10 @@ impl X86 {
                 // Outlining this into a #[cold] function regresses performance on both fast and slow paths.
                 // TODO: try using std::hint::cold_path() once MSRV is >= 1.95 and see if that does anything
                 if _mm_testz_si128(any_round_to_odd, any_round_to_odd) == 0 {
-                    // TwoSum recovers the exact residual of each widened addition. If a candidate
-                    // addition was inexact and its rounded f64 significand is even, shift it by one
-                    // ULP toward the residual. This produces a round-to-odd intermediate result.
+                    // Knuth's unconditional TwoSum establishes
+                    // `sum + residual == product + c` exactly. If a candidate addition was
```
awxkee: Nitpicking: Knuth's summation and double-double representation, triple-double etc don't track a residual of the operation, they track an error term or "rounding error". `s = a + b` has an error, and Knuth's 2Sum captures it exactly and unconditionally. While it can be considered as residual, an "error" is much more common in the literature.

Shnatsel: That's clearer. Thanks!

Shnatsel:
I've published a blog post about this work: https://shnatsel.github.io/implementing-fma-finding-bugs-in-std/

PR #324, opened 2026-08-10
Add SIMD emulation of `mul_add_precise` for f64 on SSE4.2

Shnatsel:
The implementation is based on a 2025 paper by Graillat and Muller, [Emulation of the FMA and the correctly-rounded sum of three numbers in rounding-to-nearest floating-point arithmetic](https://perso.lip6.fr/Stef.Graillat/papers/NM-2025.pdf). Integrating cutting edge research here!

To the best of my knowledge this doesn't fix any libm or musl bugs, so this isn't necessary for correctness, it's just an optimization. 

This is upwards of 3x faster than scalar on normal values; huge values and subnormals fall back to scalar and only get about 0.7x the usual scalar performance. It's still worth it because subnormals are rare and [slow even in hardware](https://gitlab.in2p3.fr/CTA-LAPP/COURS/GRAY_SCOTT_REVOLUTIONS/GrayScottRevolution/-/wikis/uploads/4-ComputingPrecision/Subnormal.pdf).

@awxkee I'd appreciate it if you could take a look

### DIFF
--- a/fearless_simd/src/generated/sse4_2.rs
+++ b/fearless_simd/src/generated/sse4_2.rs
@@ -4022,11 +4022,111 @@ impl Simd for Sse4_2 {
     }
     #[inline(always)]
     fn mul_add_precise_f64x2(self, a: f64x2<Self>, b: f64x2<Self>, c: f64x2<Self>) -> f64x2<Self> {
-        [
-            f64::mul_add(a[0usize], b[0usize], c[0usize]),
-            f64::mul_add(a[1usize], b[1usize], c[1usize]),
-        ]
-        .simd_into(self)
+        crate::kernel!(
+            #[inline(always)]
+            fn kernel(
+                token: Sse4_2,
+                a: f64x2<Sse4_2>,
+                b: f64x2<Sse4_2>,
+                c: f64x2<Sse4_2>,
+            ) -> f64x2<Sse4_2> {
+                let a_raw: __m128d = a.into();
+                let b_raw: __m128d = b.into();
+                let c_raw: __m128d = c.into();
+                let absolute_value_mask = _mm_set1_epi64x(i64::MAX);
+                let a_bits = _mm_and_si128(_mm_castpd_si128(a_raw), absolute_value_mask);
+                let b_bits = _mm_and_si128(_mm_castpd_si128(b_raw), absolute_value_mask);
+                let c_bits = _mm_and_si128(_mm_castpd_si128(c_raw), absolute_value_mask);
+                let lower_bound_minus_one = _mm_set1_epi64x((623_i64 << 52) - 1);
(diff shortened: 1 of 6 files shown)


awxkee (approved):
LGTM

Dekker product with Veltkamp's split is no doubts impossible for edge cases which the paper covers only with a blanket "provided that underflow and overflow do not occur", and instead it use fast2mult using FMA in annex to compute FMA, what is a little bit ironic. Since values that are impossible to split are gated here, everything else seems correct to me.

Shnatsel:
Thanks a lot!

Shnatsel:
Now that #323 is merged, I've rebased it on main and this is now ready for review

fearless_simd_tests/tests/harness/ops/mul_add_precise.rs
LaurenzV, line 1: Could we again move the ignored tests to the bottom?

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -4022,11 +4022,107 @@ impl Simd for Sse4_2 {
...
-        .simd_into(self)
+        crate::kernel!(
+            #[inline(always)]
+            fn kernel(
+                token: Sse4_2,
+                a: f64x2<Sse4_2>,
+                b: f64x2<Sse4_2>,
+                c: f64x2<Sse4_2>,
+            ) -> f64x2<Sse4_2> {
+                let a_raw: __m128d = a.into();
```
LaurenzV, line 4033: For such a long function, is it really good to inline everything? Should we maybe do an inline never and use simd.vectorize here somehow?

LaurenzV (approved):
I obviously haven't tried to verified this. 😄

fearless_simd/src/generated/sse4_2.rs
Shnatsel: Function calls are quite expensive with SIMD arguments so I don't think `#[inline(never)]` is a good idea. But I've been thinking about inlining and `vectorize` recently and I do believe there is a way to do it that'll be a good fit.

Shnatsel: Upon further consideration, `kernel!` already desugars into a function declaration with `#[inline]` and `#[target_feature]` on it, but not `#[inline(always)]`.

Outlining this sounds like a good idea because it could reduce register pressure and avoid spills. However, functions calling SIMD are quite slow because all the SIMD vectors need to be spilled to the stack, they cannot be passed in registers. A change to pass them in registers was made in Rust 1.87 but it broke inlining across functions with compatible but not identical `#[target_feature]` annotations and caused massive regressions.

I don't think it's a good idea to _force_ register spills here out of fear it _might_ cause register spills if we don't force them.

LaurenzV: Makes sense!

PR #325, opened 2026-08-11
Docs: Show where #[inline] goes in kernel!()

Shnatsel:
I wanted to look it up recently but it wasn't in the doc comment.


DJMcNab (approved):
I guess?

PR #331, opened 2026-08-15
Fix example link in readme

### LINKED ISSUE #326: Readme links to wrong example
The readme [says](https://github.com/linebender/fearless_simd/blob/54fbd6205657e70d0459d2abed02c04fd49926da/fearless_simd/README.md?plain=1#L209-L210)

> You can also [disable certain instruction sets for select functions](https://github.com/linebender/fearless_simd/blob/main/fearless_simd/examples/sigmoid.rs) without disabling them globally.

However, the linked example doesn't seem to show that show that ability, and I don't think that ability exists.

Without thinking much about it, maybe a syntax like:

```rust
let res: T = dispatch!(level, simd => use_simd(simd));

// Maybe the simd fallback and sse2 are worse than a manual fallback, let me do my own fallback
// but automatically try to simd-ize any other types, including future added types that
// the caller may not not about
let res: Option<T> = dispatch!(level, simd => use_simd(simd), !Fallback, !Sse2);

// This is basically already possible with manual `.as_neon()`, etc, but
// could be nice to be able to do it automatically
let res: Option<T> = dispatch!(level, simd => use_simd(simd), Neon, Avx512);
```

danderson:
Fixes #326

### DIFF
--- a/fearless_simd/README.md
+++ b/fearless_simd/README.md
@@ -206,7 +206,7 @@ Note that later extensions can be beneficial even if you are only using 128-bit 
 AVX2 and AVX-512 provide more efficient instructions for some operations,
 and AVX-512 also more than doubles the number of vector registers of all sizes.
 
-You can also [disable certain instruction sets for select functions](https://github.com/linebender/fearless_simd/blob/main/fearless_simd/examples/sigmoid.rs)
+You can also [disable certain instruction sets for select functions](https://github.com/linebender/fearless_simd/blob/main/fearless_simd/examples/disable_avx2_for_one_function.rs)
 without disabling them globally.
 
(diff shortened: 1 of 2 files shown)


Shnatsel:
README.md is generated through `cargo rdme` 1.x (not the nightly-only 2.x). It's a little backwards, the usual way is to use the README as the lib.rs doc comment instead, but that's what this repository does.

danderson:
Oops, TIL. Changed the module doc and ran cargo-rdme 1.5.1 to regenerate README.md.

Shnatsel:
Thanks!

Shnatsel (approved):

PR #333, opened 2026-08-16
Fix dangling `libm` link when building docs without the libm feature

danderson:
Checked that `cargo doc` with and without the libm feature builds without warnings.

The README.md didn't change when regenerated by `cargo rdme`, which surprised me slightly. But looking at the readme, it has a block of explicit links outside of the generated section that fixes up libm references already, so no change needed there.

(extracted from #330 after feedback from @DJMcNab, thank you!)


DJMcNab (approved):
Yeah - `cargo rmde` doesn't understand cfg flags (at least, the 1.x series)

PR #335, opened 2026-08-16
Add TryFrom bounds for all integer primitive types to SimdIntElement

danderson:
I skipped these in #302 because conditional conversion didn't seem great for typically performance sensitive simd code. However, it turns out that these conversions are very useful in tests especially, and even in regular code LLVM can often rule out failure and elide the branch.

Still no `From` impls other than `bool`, because that's the lowest common denominator across signed/unsigned and all the different sizes.

No change to `SimdFloatElement`: unfortunately they don't implement TryFrom for integers, so without num-traits's cast trait (which provides the equivalent of `as` conversion) we're stuck with the existing lowest common denominator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ You can find its changes [documented below](#070-2026-08-11).
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


danderson:
Also added extra bounds so that `TryFrom::Error` must be `Debug`, so that `unwrap` and `expect` work without further bounds in the caller.

danderson:
Hmm, the extra debug bounds cause failures that I don't get on my machine. Converting to draft while I figure it out.

danderson:
Okay, fixed. The TryFrom impls all use either a `core::num::TryFromIntError` or `Infallible` as their error type, so I made TryFrom::Error's the useful traits both of those implement (`Error` implies `Debug` and `Display`).

Shnatsel (approved):
Sorry it took a while to properly review. I was concerned about different concrete error types being problematic somehow, like messing with downstream `impl Trait` or being hard to store in a struct, but I cannot find any fault with this. Thanks!

Issue #342, opened 2026-08-22
mask8x64<Neon>::to_bitmask() is sub-optimal


Dr-Emann:
I haven't fully figured out how to contribute here, with the generated code and all, but I've found a better way to implement `mask8x64<Neon>::to_bitmask()`:

```rust
kernel! {
    #[inline(always)]
    fn aarch64_bitmask(simd: Neon, mask: mask8x64<fearless_simd::Neon>) -> u64 {
        use core::arch::aarch64::*;

        let lanes = <[i8; 64]>::from(mask);
        let lanes: u8x64<_> = i8x64::simd_from(simd, lanes).bitcast();
        let (l, r) = lanes.split();
        let ((a, b), (c, d)) = (l.split(), r.split());

        // The `sri` chain below packs lane `j` of `q0..q3` into bit `4j + k`, so it wants
        // the 64 lanes 4-way deinterleaved: `a` holds bytes 0..16, but `q0` has to hold
        // bytes 0, 4, 8, and so on.
        let (even_ab, odd_ab) = a.deinterleave(b);
        let (even_cd, odd_cd) = c.deinterleave(d);
        let (q0, q2) = even_ab.deinterleave(even_cd);
        let (q1, q3) = odd_ab.deinterleave(odd_cd);
        let (q0, q1, q2, q3) = (q0.into(), q1.into(), q2.into(), q3.into());

        // shift the second vector right by one, insert the top bit from the first vector
        // The top two bits each element of temp0 are from the first and second vector
        let temp0 = vsriq_n_u8::<1>(q1, q0);
        // shift the fourth vector right by one, insert the top bit from the third vector
        // The top two bits each element of temp1 are from the third and fourth vector
        let temp1 = vsriq_n_u8::<1>(q3, q2);
        // shift temp1 (the top two bits of which are from the third and fourth vector) right by 2,
        // insert the top two bits from temp0 (the top two bits of which are from the first and
        // second vector)
        // The top four bits of each element of temp2 are from the first, second, third, and fourth
        // vector
        let temp2 = vsriq_n_u8::<2>(temp1, temp0);
        // duplicate the top 4 bits into the bottom 4 bits of each element
        let temp3 = vsriq_n_u8::<4>(temp2, temp2);

        // Returns a value where the first 4 bits are the low 4 bits of the first element,
        // the next 4 bits are the high 4 bits of the second element, and so on.
[...]

PR #343, opened 2026-08-22
Fold wide mask reductions with a bitwise op instead of short-circuiting

Dr-Emann:
For masks wider than the native register, any/all_true/false now combine the split halves with `or`/`and` and do a single horizontal aggregation, rather than aggregating each half and joining with `||`/`&&`.

This trades away early-exit on the first half for one lane-wise vector op plus one scalar extraction, which is the cheaper path in practice.


Shnatsel:
Thank you! I'll try to take a look tomorrow.

Is this a fix for #342 or is this unrelated?

Dr-Emann:
Unrelated

Shnatsel (approved):
Looks good to me, thank you!

llvm-mca agrees this is beneficial on both AVX2 and NEON.

PR #344, opened 2026-08-23
Optimize to_bitmask() on NEON

Shnatsel:
A stab at #342 to see what it would take to optimize it.

Summary of changes:

- Add dedicated handwritten kernels for 8x32 (256-bit) and 8x64 (512-bit) masks
- Narrow 16/32/64-bit masks into 8-bit ones (or wider, if 128-bit width is reached before 8-bit elements) 

The narrowing reduction into base 8-bit kernels helps a lot; mask16x16 sees 1.5x to 2.x throughput improvement depending on the CPU.

The 256-bit kernel was discovered by having Codex generate a dozen candidates and select the best one according to llvm-mca. It saves 4-5 cycles of latency and provides 30% better throughput compared to the generic split-combine lowering into 128-bit ops. It is tested exhaustively on all possible valid mask patterns.

I'm not entirely convinced this is worth optimizing, so marking it as draft. If this really is an important operation for some algorithms, similar optimizations should also be extended to AVX2 and possibly SSE4.2; AVX-512 is already zero-cost.

### DIFF
--- a/fearless_simd/src/generated/neon.rs
+++ b/fearless_simd/src/generated/neon.rs
@@ -1197,9 +1197,9 @@ impl Simd for Neon {
                     1, 2, 4, 8, 16, 32, 64, 128, 1, 2, 4, 8, 16, 32, 64, 128,
                 ]);
                 let bits = vandq_u8(vreinterpretq_u8_s8(a.into()), weights);
-                let lo = vaddv_u8(vget_low_u8(bits)) as u64;
-                let hi = vaddv_u8(vget_high_u8(bits)) as u64;
-                lo | (hi << 8)
+                let rotated = vextq_u8::<8>(bits, bits);
+                let paired = vzip1q_u8(bits, rotated);
+                vaddvq_u16(vreinterpretq_u16_u8(paired)) as u64
             }
         );
         kernel(self, a)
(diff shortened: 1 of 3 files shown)


fearless_simd_gen/src/mk_neon.rs
```diff
@@ -645,6 +645,14 @@ impl Level for Neon {
...
         }
     }
+
+    fn should_use_generic_op(&self, op: &Op, vec_ty: &VecType) -> bool {
+        if matches!(op.sig, OpSig::MaskToBitmask) {
+            false
+        } else {
+            op.sig.should_use_generic_op(vec_ty, self.native_width())
+        }
+    }
```
Shnatsel, line 655: I don't like this, it should hopefully be gone once AVX2 and SSE are optimized.

Shnatsel:
x86 has dedicated `movemask` intrinsics for this that lower into a single instruction even in SSE2, and zip/unzip on AVX2 is expensive. LLVM is also smart enough to rewrite the recursive decomposition into packing trees. So none of the tricks from here really apply to x86.

fearless_simd_gen/src/mk_neon.rs
Shnatsel: x86 already does something similar so I guess we'll live with it

Shnatsel:
The complexity is isolated to NEON so it seems worthwhile to add this. Marking as ready for review.

fearless_simd/src/generated/neon.rs
```diff
@@ -1197,9 +1197,9 @@ impl Simd for Neon {
                     1, 2, 4, 8, 16, 32, 64, 128, 1, 2, 4, 8, 16, 32, 64, 128,
                 ]);
                 let bits = vandq_u8(vreinterpretq_u8_s8(a.into()), weights);
-                let lo = vaddv_u8(vget_low_u8(bits)) as u64;
-                let hi = vaddv_u8(vget_high_u8(bits)) as u64;
-                lo | (hi << 8)
+                let rotated = vextq_u8::<8>(bits, bits);
+                let paired = vzip1q_u8(bits, rotated);
+                vaddvq_u16(vreinterpretq_u16_u8(paired)) as u64
```
LaurenzV, line 1202: Will this work correctly on big-endian targets? Not sure if we care about those? Maybe we should run those in CI as well, as far as I can tell we currently don't.

LaurenzV: Probably doesn't matter, just wondering!

LaurenzV (approved):
Otherwise fine from my side, without having tried to understand the new approach!

fearless_simd/src/generated/neon.rs
Shnatsel: Big-endian Aarch64? No, we definitely don't care. Support for it was removed from Linux 7.4 because it is never used.

Shnatsel:
The old approach was splitting 256/512-bit vectors into 2/4 128-bit vectors, reducing their values to bits individually and combining the result.

The new approach instead narrows elements first (u32->u16, u16->u8, etc) to reduce the width of the vector in bits, until it reaches either 128 bits or u8 elements, and then goes through the handwritten u8 codepaths.

Does that make sense?

LaurenzV:
Oh yeah it does, I meant more like I didn't spend time trying to verify all the used intrinsics, haha. :smile:

Issue #338, opened 2026-08-18
vectorize() + #[inline(always)] is error-prone and hard to teach


Shnatsel:
After seeing an HPC expert who literally wrote a book on HPC and SIMD struggle with it (and entirely failing to get it to work when combined with criterion benchmarks), I'm convinced that we need to provide a more robust and convenient way to handle this.

Ideally I want to have some sort of annotation that you just put on every function generic over `Simd` and do not think about inlining at all, ever. The compiler just does the right thing automatically.

I see multiversion's `#[multiversion(targets=simd)]` ([PR adding it](https://github.com/calebzulawski/multiversion/pull/52)) and [thermite](https://github.com/raygon-renderer/thermite)'s `#[thermite::dispatch(V)]` as prior art that does this well. 

AFAIK (I haven't re-checked yet) `multiversion` has an indirect dispatch through a cached atomic pointer on every function call, but doesn't require functions to be generic over `Simd`. I haven't looked into how `thermite` works yet.

Shnatsel:
We could add an (optional) proc macro `#[simd]` in a new `fearless_simd_macros` crate that performs the following transformation. `fearless_simd_macros` would not be depended on by `fearless_simd` and is not subject to the v1.0 stability guarantees.

Let's say we have an input like this:

```rust
fn do_stuff<S: Simd>(simd: S, ...) {
    // do stuff here
}
```

This would get transformed into

```rust
fn do_stuff<S: Simd>(simd: S, ...) {
    simd.vectorize_inline(|| {
        // do stuff here
    }
}
```

where `vectorize_inline` is basically the current `vectorize()` but with an `#[inline]` on the inner function:

```rust
#[inline(always)]
fn vectorize_inline<F: FnOnce() -> R, R>(self, f: F) -> R {
    #[inline]
    #[target_feature(enable = "neon")]
    fn vectorize_neon<F: FnOnce() -> R, R>(f: F) -> R {
        f()
    }
    unsafe { vectorize_neon(f) }
}
```

There are drawbacks:

1. Proc macro infrastructure inflates build times
   - Kind of no way around this? Declarative macros are not as nice to use. Doing something like `simd!( /* do stuff */  )` will likely run into hygiene issues. We keep the manual way for those who don't want the proc macro.
2. "The variables that a [closure](https://doc.rust-lang.org/reference/types/closure.html) captures by move are dropped in an unspecified order." - [Rust reference](https://doc.rust-lang.org/reference/destructors.html)
[...]

PR #347, opened 2026-08-24
Add #[simd] macro

Shnatsel:
My attempt at #338

I am not well-versed in proc macros. While the design of what the macro should expand into is mine, the implementation of the macro and thinking through various edge cases and handling them was largely delegated to Codex (GPT-5.6 Sol Ultra). 

After the initial AI implementation I moved all soundness-critical parts into a declarative macro, which does all the interesting stuff, and which I find a lot easier to understand. The proc macro is just there as a non-soundness-critical convenience that forwards arguments to the declarative macro.

Before implementing the macro I investigated `vectorize()` and found that it already gets inlined most of the time anyway. It doesn't work well as an inlining barrier, since the function that isn't annotated `#[inline]` is trival enough for the optimizer to always inline it anyway, even with a large closure body.  So I reverted a separate `vectorize_inline()` which is recorded in commit history, and inlining can instead be controlled with annotation on the function calling `vectorize()`; the docs on it already show that usage in the example, the doc comment on vectorize() can probably be improved.

The proc macro crate is separate from fearless_simd so it doesn't fall under the v1.0 guarantees. I expect to publish it in lockstep with v1.0, although we might ship pre-releases just to make sure it all works. The README points to git but should work with most versions no problem, just wouldn't have as much inlining annotations in `vectorize()`.


LaurenzV:
I think let's wait until https://github.com/linebender/vello/pull/1853 has been merged and then we can try the ergonomics of using it there?

Dr-Emann:
Woudln't mind the `#[inline]` on the inside of vectorize (and/or eliminating the inner call when at baseline), I'm seeing some useful performance impact on aarch64 by doing it kinda manually with:

```rust
#[cfg(all(target_arch = "aarch64", target_feature = "neon"))]
if simd.level().as_neon().is_some() {
    return inner_code(a, b, c);
}
simd.vectorize(
    #[inline(always)]
    move || inner_code(a, b, c);
}
```

I believe it's somewhat exaggerated by taking enough arguments, in:
```rust
        #[target_feature(enable = "neon")]
        fn vectorize_neon<F: FnOnce() -> R, R>(f: F) -> R {
            f()
        }
        unsafe { vectorize_neon(f) }
```
`vectorize_neon` gets a single closure struct containing a, b, and c, which spills to the stack, rather than being passed as registers which passing as arguments can do.

Wonder if there's some macro magic we could do in the `#[simd]` impl that could use real arguments, rather than delegating to `.vectorize(closure)`

Dr-Emann:
And/or, maybe just special casing a few arguments, and doing an annoying `simd.vectorize_1(a, #[inline(always)] |a| { ... })`, `simd.vectorize_2(a, b, #[inline(always)] |a, b| { ... })`, ...

Shnatsel:
Ugh, turns out closure-captured arguments get put into a struct that then gets spilled to the stack.

I've rewritten the macro to avoid that with Astra's assistance. 

The proc macro now also needs to refer to fearless_simd crate explicitly, and to avoid making it unsound by spoofing fearless_simd crate I had to move all the `unsafe` into a declarative macro inside fearless_simd itself. 

I think it's a reasonable split going forward, especially since all the level matching is now happening inside fearless_simd where it can all be updated in lockstep if/when we add new levels. Alternatively we can make renaming fearless_simd harder by using `::fearless_simd` in the proc macro output and this will require renaming it in Cargo.toml; technically a soundness hole if you do that, but not a practical one.

Shnatsel:
tests fail because `trybuild` is too strict and the changes to the compiler's diagnostic output between versions trip it up. The current snapshots expect 1.97 output while CI runs on 1.89.

There is no fuzzy matching functionality in `trybuild` so we might have to use something else for testing the macro.

Shnatsel:
I've fixed the issues around `trybuild` being overly strict for UI tests by migrating to `ui_test`

The README.md is quite poor and needs to be rewritten, but I'll deal with that once the code settles down

Shnatsel:
I feel a lot more confident in this code now that I've moved all the safety-critical bits into a straightforward and comprehensible declarative macro. Well, comprehensible by macro standards, anyway.

Shnatsel:
I've updated to documentation, I think this is fully ready to go now.

fearless_simd/examples/srgb.rs
```diff
@@ -49,7 +49,7 @@ fearless_simd::kernel!(
 ///
 /// This helper shows how portable SIMD code can opportunistically call
 /// target-specific kernels while still providing a fallback for every backend.
-#[inline(always)]
+#[inline(always)] // or #[simd], either works
 fn copy_alpha<S: Simd>(a: f32x4<S>, b: f32x4<S>) -> f32x4<S> {
```
RunDevelopment, line 53: That's not true yet. #379

fearless_simd/README.md
```diff
@@ -132,18 +165,13 @@ You can also [mix and match](https://github.com/linebender/fearless_simd/blob/ma
...
+Fearless SIMD requires functions that use SIMD to be annotated with the `#[simd]` attribute from the `fearless_simd_macros` crate.
 
-- All SIMD functions need `#[inline(always)]`.
-- Use [`dispatch`] when calling SIMD code from non-SIMD code.
-- Use [`vectorize()`][Simd::vectorize] when calling SIMD from SIMD if you don't want to force inlining.
+Use [`dispatch`] when calling SIMD code from non-SIMD code.
 
-[The article describing the design](https://shnatsel.github.io/safe-simd-in-rust-even-on-the-inside/#the-abi-would-like-a-word) covers why this is the
-case. There's also Q&A on [Zulip](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/inlining/with/546913433).
+If you cannot use proc macros, [you can achieve the same effect manually](https://github.com/linebender/fearless_simd/blob/main/fearless_simd/SECURITY.md), but it requires some care. The use of `#[simd]` is recommended as the more robust and ergonomic option.
```
RunDevelopment: Wrong link
```suggestion
If you cannot use proc macros, [you can achieve the same effect manually](https://github.com/linebender/fearless_simd/blob/main/fearless_simd/MANUAL_INLINING.md), but it requires some care. The use of `#[simd]` is recommended as the more robust and ergonomic option.
```

fearless_simd/README.md
```diff
@@ -132,18 +165,13 @@ You can also [mix and match](https://github.com/linebender/fearless_simd/blob/ma
 intrinsics with the other approaches, using high-level code most of the time and dropping down to
 hardware-specific intrinsics only when necessary.
 
-## Inlining
-
-Fearless SIMD relies heavily on Rust's inlining support to create functions which have the given target features enabled.
+### The `#[simd]` annotation
 
-As a rule of thumb:
+Fearless SIMD requires functions that use SIMD to be annotated with the `#[simd]` attribute from the `fearless_simd_macros` crate.
```
RunDevelopment: > requires

Not really part of the review, but I want to give some feedback. 

I take issue with "requires" here and the general framing of `#[simd]`/`Simd::vectorize`/`#[inline(always)]` being necessary for SIMD. Technically, you don't need to use any of them for SIMD. Example:

```rs
#[inline(never)] // no inlining to make a point
fn add<S: Simd>(a: f32x4<S>, b: f32x4<S>) -> a: f32x4<S> {
    a + b
}
```

This function will work (=correct result), and the addition will indeed use whatever SIMD instruction `S` allows to perform the addition. The only issue is that it is going to be slow. Without any target features, the internal function that actually executes the SIMD instruction can't be inlined. So you pay for a whole function call + missing optimization opportunities to run (probably) a single instruction. Ouch.

This *really* tripped me when I first used `fearless_simd`. I read your docs, but it wasn't 100% clear to me why `#[inline(always)]` and the likes were strictly necessary. My code worked, tests passed. Took me some time to figure out why the fearless AVX2 version of the encoder was slower than the previous version. I only figured it out after looking at the assembly of my program.

So, please, don't just document how to use fearless correctly; also show what happens if you don't. Show some assembly and benchmarks to emphasize how bad it is. As a user, I can say that it's easy with this library to "hold it wrong." `#[simd]` is going to help a lot in that regard (because it's easy to use), but at least the section on manual inlining should drive home what the consequences for improper usage are.

fearless_simd_macros/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+
+    match &*argument.pat {
+        Pat::Ident(pattern) => {
+            reject_conditional_attributes(&pattern.attrs)?;
+            if let Some(by_ref) = &pattern.by_ref {
+                return Err(syn::Error::new(
+                    by_ref.span,
+                    "the SIMD token parameter must be bound by value, not `ref`",
+                ));
+            }
```
RunDevelopment: Any reason why `ref simd: S` is disallowed? Weird for sure, but seems sound to me.

fearless_simd_macros/README.md
```diff
@@ -0,0 +1,113 @@
...
+separately so that the macro can evolve without adding a procedural-macro
+dependency to `fearless_simd` itself.
+
+Add both packages from crates.io to your `Cargo.toml`:
+
+```toml
+[dependencies]
+fearless_simd = "1.0"
+fearless_simd_macros = "0.1"
+```
```
RunDevelopment, line 14: Considering #379 and further library evolution, `fearless_simd_macro` is only going to be compatible with certain versions of the main crate. I think this should be documented.

fearless_simd_macros/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+            .parse_args()
+            .expect("inline attribute has one identifier argument");
+        assert_eq!(inline_kind, "always");
+        assert_eq!(body.block.stmts.len(), 3);
+        assert!(
+            parsed
+                .attrs
+                .iter()
+                .all(|attr| !attr.path().is_ident("inline"))
+        );
```
RunDevelopment: God, this is hard to read. No idea whether any of that is correct.

fearless_simd/src/kernel_macros.rs
```diff
@@ -182,6 +184,111 @@ macro_rules! __fearless_simd_kernel_target_fn {
...
+        // block loses the expected FnOnce signature during closure input
+        // inference. Method lookup retains it without adding a machine argument.
+        struct __FearlessDispatch;
+        impl __FearlessDispatch {
+            // Do not force inlining: a large body may be shared by several
+            // callers. Even an out-of-line dispatcher keeps separate arguments.
+            #[inline]
+            fn call<S: $crate::Simd, $($argument_type,)* F, R>(
+                self, simd: S, $($argument: $argument_type,)* f: F,
+            ) -> R
```
RunDevelopment: I'm pretty sure this code gen (or any code gen really) isn't going to support trait methods that use associated types. Example:

```rs
impl Trait for Foo {
  type A = 32;
  #[simd]
  fn oh_no<S: Simd>(&self, _: S, a: Self::A) { ... }
}
```

I don't see this pattern in the tests anywhere.

fearless_simd_tests/tests/ui/simd/pass/functions.rs
```diff
@@ -0,0 +1,76 @@
...
+
+//@run
+
+use fearless_simd::{Fallback, Simd};
+use fearless_simd_macros::simd;
+
+#[simd]
+#[inline(never)]
+#[must_use]
+fn with_generics<'a, S, T>(simd: S, value: &'a mut T) -> Result<&'a mut T, ()>
```
RunDevelopment, line 12: Could you also add a test with const generics? Example:

```rs
fn with_const_generics<S: Simd, const N: usize>(simd: S, array: [f32; N]) -> [u8; N] {
  todo!()
}
```

fearless_simd/src/kernel_macros.rs
RunDevelopment, line 231: I'm pretty sure this code gen (or any code gen really) isn't going to support trait methods that use associated types. Example:

```rs
impl Trait for Foo {
  type A = 32;
  #[simd]
  fn oh_no<S: Simd>(&self, _: S, a: Self::A) { ... }
}
```

I don't see this pattern in the tests anywhere.

fearless_simd_macros/src/lib.rs
RunDevelopment: I hope the AI generated this. If you wrote this by hand, please talk to a loved one. 😟

Anyway, there got to be a better way to test a macro expansion. I'd say just `assert!(expanded.to_string(), "...");`, but unformatted code isn't nice (better than *this* though). I haven't used it myself, but [`macrotest`](https://docs.rs/macrotest/latest/macrotest/) might be the tool for the job here.

RunDevelopment (commented):
Since you wanted the PR reviewed, I spent some time reading through it. Far from a proc macro expert, but I have some experience with it.

fearless_simd/src/kernel_macros.rs
RunDevelopment, line 231: We should probably suppress clippy lints for this generated function (and maybe all code generated here).

Image

fearless_simd_macros/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+            // Keep `self` captured so its uses, including inside nested macros,
+            // retain their original meaning without rewriting the body.
+            continue;
+        };
+        if has_conditional_attributes(&argument.attrs) {
+            // FnOnce's argument types cannot carry cfg attributes. Preserve
+            // conditional parameters as captures, just as in the old expansion.
+            continue;
+        }
+        let name = Ident::new(&format!("__fearless_argument_{index}"), Span::mixed_site());
```
RunDevelopment: These names get shown to the user by Rust Analyzer and rustdoc. I think we should use the original names whenever possible.

Image

RunDevelopment (commented):
Found 2 minor issues after using the macro.

fearless_simd_macros/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+    if !args.is_empty() {
+        return Err(syn::Error::new_spanned(
+            args,
+            "`#[simd]` does not accept arguments",
+        ));
+    }
+
+    // ItemFn's signature grammar also accepts body-bearing inherent and trait
+    // methods. Parsing the item rejects bodyless and specialization methods.
+    let mut function: ItemFn = syn::parse2(item)?;
```
RunDevelopment: Could we get a custom compiler error here? The one currently emitted is a plain parser error with no details at all. Maybe something like "`#[simd]` can only be used on function and method definitions: {current_error}."

Here's an example of applying the macro to a struct. Errors are similarly bad for other non-fn items as well.

```rs
#[simd]
struct Foo;
```
```
error: expected `fn`
   --> src\encode\bc4.rs:783:1
    |
783 | struct Foo;
    | ^^^^^^
```

fearless_simd/README.md
Shnatsel: Great point, thank you!

It can be difficult to know what exactly needs explaining when you already have the whole system in your head, so this kind of feedback is very valuable!

Shnatsel: I've added a note on what happens if you omit `#[simd]`, how's this?

fearless_simd_macros/src/lib.rs
Shnatsel: Not anymore, as far as I can tell. Rejecting it is now more code than leaving it be, so I'll leave it be.

fearless_simd_macros/src/lib.rs
Shnatsel: I was hoping these wouldn't be visible from the outside, but alas.

There are ambiguities in some edge cases, see https://doc.rust-lang.org/reference/patterns.html#identifier-patterns where some disambiguation can only happen after name resolution, but we can mostly preserve the original argument names.

Shnatsel: The things you have to do to cause an ambiguity are really quite exotic, like constructing a new unit struct inline in the arguments, so we can probably just ignore them.

fearless_simd/src/kernel_macros.rs
Shnatsel: Associated types just work, actually. I added a test to keep them that way.

A trickier case is `impl <S: Simd>` but I don't want to deal with that just yet.

fearless_simd/README.md
RunDevelopment: This isn't resolved yet.

fearless_simd/README.md
RunDevelopment: Short and simple. It's good.

I would just go into my detail *why* it is slow in the section about manual inlining. But assuming that only experts will use manual inlining, maybe it's not necessary?

fearless_simd/README.md
Shnatsel: Oh, `cargo rdme` overwrote that after I applied the github suggestion. Good catch. That thing keeps tripping me up, the sync almost always goes in the other direction via `#![doc = include_str!("README.md")] `

Shnatsel: Actually resolved now. Thanks!

fearless_simd_macros/src/lib.rs
Shnatsel: Yeah that's from the LLM.

I just deleted this hideous test. It didn't look like it was needed. If you think we should bring back a more focused form of it please let me know.

fearless_simd/src/kernel_macros.rs
```diff
@@ -182,6 +184,137 @@ macro_rules! __fearless_simd_kernel_target_fn {
...
+                    // This public predicate works even when the library omits
+                    // Level::Fallback. Do not inspect downstream Cargo features.
+                    level if level.is_fallback() => f($($argument),*),
+                    // Level is non_exhaustive. Future backends still get their
+                    // own feature context until this macro learns about them.
+                    _ => $crate::Simd::vectorize(
+                        simd,
+                        #[inline(always)]
+                        move || f($($argument),*),
+                    ),
```
LaurenzV, line 275: So am I understanding correctly that in theory, this macro is unnecessary and we could just always call this instead? But what this macro does is that it passes the arguments as direct method parameters instead relying on Rust to capture the parameters of the closure?

fearless_simd/MANUAL_INLINING.md
```diff
@@ -0,0 +1,44 @@
+# Avoiding #[simd] proc macro with manual inlining
+
+The use of `#[simd]` is recommended as the more ergonomic option. This document describes how to achieve the same effect manually, if you have to.
```
LaurenzV, line 3: So, I checked again on your vello branch, and I saw that for gradients, you removed the simd macro and instead replaced it with inline always. And indeed, when I tried using the SIMD macro instead of inline always, I got worse performance.

The documentation makes it sound like if you are fine using `#[simd]` you can just always use this and don't have to worry about anything anymore, but clearly that doesn't seem to apply in some cases. But I'm wondering whether there is any rule to that and how we can convey that to users somehow? Or is it possible for the simd macro to take an additional `inline_always` attribute or something, so that the user still has some control over this behavior if we can't make this issue go away "magically"?

LaurenzV: I tried putting `inline(always)` on top of the methods and it recouped the performance loss for radial gradients, but not sweep gradients. But maybe that's enough. Should we document that it might be worth combining with inline always (or that it's supported in the first place)?

Do we need to document that, in case it doesn't work, users can still try not using the simd annotation and just marking the method as inline always>

fearless_simd/src/kernel_macros.rs
LaurenzV: Ah, I see now this is explained in the comment in the macro definition!

fearless_simd_macros/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2018 Raph Levien
```
LaurenzV: Is this really the correct copyright since this has been written from scratch? Maybe "2026 The Fearless SIMD authors"?

LaurenzV (commented):
Thanks a lot! I just have a single question/concern regarding the usage of the simd macro. I have not tried to validate all of the edge cases of the macro itself, but clearly you've already invested a lot of time into finding those!

I'm just going to rerun the vello benchmarks on AVX2 and if it all looks good happy to approve.

Also, three comments from codex, they seem to be accurate from that I can tell:

Image

fearless_simd/MANUAL_INLINING.md
Shnatsel: This is the usual messing around with inlining that I don't think the presence of `#[simd]` really changes. So I'm not sure how to document it succinctly.

It's not like the `#[simd]` macro didn't do its job in that case. It does, and the gradient functions still emit SIMD instructions. The compiler simply decided to keep it a separate function based on its heuristics, and that happened to reduce performance in this specific case.

LaurenzV: Okay, fair enough!

LaurenzV (approved):

Shnatsel:
On Codex critique:

1. This is valid, but would require a great deal of additional complexity to support. I don't really want to go down that road. 
2. Fixed in 0488084a2d231290b29bb8222d59e26703813350 and made the code simpler, nice!
3. Good catch, fixed!

Shnatsel:
I'm excited to have this finally merged, thank you both @RunDevelopment and @LaurenzV for the review!

PR #348, opened 2026-08-24
Optimize 64-bit int<->float conversions on AVX2, add more tests

### LINKED ISSUE #311: Optimize float<->integer conversions further
#310 which adds 64-bit conversions left some performance on the table

also, the 32-bit AVX2 conversions have a branch inside them, which might cause rampant misprediction; looking for branchless formulations may be worthwhile

there's also a mention of a "magical mystery algorithm taken from LLVM" for SSE and we might want to plumb LLVM for more of those

Shnatsel:
Closes #311

Summary:

1. Replace a branch with a xor in the 32-bit codepath, which is free on anything but haswell thanks to ILP
2. Route f64x2 -> i64x2 conversions through non-precise scalar conversion intrinsic `_mm_cvttsd_si64` which is faster than SIMD formulations thanks to ILP (for 2-element vectors only) and cheaper than `as` casts thanks to not handling saturation
3. Add AVX2 specializations for 64-bit element conversions, some lifted from LLVM and other kernels selected by Codex for performance on llvm-mca out of a large pool of candidates.

llvm-mca reciprocal throughput for 256-bit vectors on AVX2, lower is better:

Conversion | Haswell | Skylake | Alder Lake | Zen 3
-- | -- | -- | -- | --
i64 → f64, old → new | 10 → 2.0 | 10 → 1.5 | 8 → 1.3 | 4 → 2.0
u64 → f64, old → new | 2.5 → 2.0 | 1.7 → 1.5 | 1.7 → 1.3 | 2.0 → 2.0
f64 → i64, approximate | 12.3 → 4.5 | 12 → 3.0 | 12 → 3.0 | 8 → 4.0
f64 → i64, precise | 12.3 → 6.0 | 12 → 4.0 | 12 → 4.3 | 8 → 4.5
f64 → u64, approximate | 20 → 3.0 | 16 → 2.0 | 16 → 2.0 | 12.5 → 3.5
f64 → u64, precise | 20 → 3.8 | 16 → 2.5 | 16 → 2.5 | 12.5 → 4.0

Up to 8x speedup!

Also adds in-depth tests for various edge cases against a from-scratch integer-only reference implementation because I don't trust anything anymore after the libm and musl FMA fiasco, and testing an algorithm from LLVM against itself is silly.

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -4066,19 +4066,103 @@ impl Simd for Avx2 {
     }
     #[inline(always)]
     fn cvt_u64_f64x2(self, a: f64x2<Self>) -> u64x2<Self> {
-        [a[0usize] as u64, a[1usize] as u64].simd_into(self)
+        crate::kernel!(
+            #[inline(always)]
+            fn kernel(token: Avx2, a: f64x2<Avx2>) -> u64x2<Avx2> {
+                let a = a.into();
+                let bits = _mm_castpd_si128(a);
+                let exponent = _mm_srli_epi64::<52>(bits);
+                let significand = _mm_or_si128(
+                    _mm_and_si128(bits, _mm_set1_epi64x(0x000f_ffff_ffff_ffff)),
+                    _mm_set1_epi64x(0x0010_0000_0000_0000),
+                );
+                let shift_bias = _mm_set1_epi64x(1075);
+                let right_count = _mm_sub_epi64(shift_bias, exponent);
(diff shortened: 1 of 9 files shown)


LaurenzV (approved):
As you already know I'm not a huge fan of big tests with large loops and magic numbers, but if you think it makes sense then let's just keep it. 😄

Shnatsel:
Yeah, sorry, these are kind of necessary. 64 bits is a very large input space, so throwing a bunch of handwritten inputs at it isn't very useful, and you can't write a simple exhaustive test either (which works fine for 32 bits). I'd rather verify the edge cases like this than ship a buggy implementation accidentally.

PR #349, opened 2026-08-25
Add portable SIMD swizzle, multishift, compress, and expand

Novum:
I needed those ops in a project of mine so I gave it a go. 

This adds 128-, 256-, and 512-bit operations for concatenated dynamic swizzles, per-64-bit-lane multishift, masked compression/expansion, and masked expand-loads.

Concatenated swizzles use optimized existing primitives on SSE4.2, AVX2, Neon, and Wasm, while AVX-512 uses native VBMI/VBMI2 instructions.

As usual there are scalar fallbacks for portable code.


Shnatsel:
Thanks for the PR!

I really don't want to add ops that only exist on AVX-512 as "portable" ops. The proper way to deal with these is to write a `kernel!` for AVX-512 and route everything else to a fallback in user code, like this: https://github.com/linebender/fearless_simd/blob/main/fearless_simd/examples/srgb.rs

I can see us adding `concat_swizzle_dyn` along with the `_precise` variant because that can have meaningfully different lowerings than simply calling `swizzle_dyn_precise` and joining the halves, which a user could already do. NEON can directly use wider `tbl` for 128 and 256-bit vectors and has a reasonable lowering for 512-bit vectors by OR-ing two 512-bit swizzle_dyn_precise.

Novum:
Fair. It would be nice to just write code once though without even having to write the scalar path too. But I can see that usually you probably want to do things differently than the direct translation from AVX512 for the scalar path.

Shnatsel:
I'm going to go ahead and close this, but I'd welcome a PR for just concat_swizzle_dyn.

PR #350, opened 2026-08-26
LLVM-inspired optimizations for x86

Shnatsel:
Some lowerings are lifted from LLVM while others are original. Best reviewed commit by commit, one commit = one op. Each commit message includes llvm-mca data, and also benchmark results where possible.


Shnatsel:
@valadaptive commit 624a49f16614e1cd6d880942dd36b9312249e066 might be worth upstreaming back into LLVM. They use it for 128-bit vectors only, but llvm-mca and real-hardware benchmarks on AVX2 show it to be beneficial for wider vectors too.

LaurenzV (approved):

PR #354, opened 2026-08-27
Add concat_swizzle_dyn and concat_swizzle_dyn_precise

Shnatsel:
1. Add concat_swizzle_dyn and concat_swizzle_dyn_precise
2. Change swizzle_dyn_precise SSE4.2 128-bit formulation; it's one fewer instruction, slightly worse latency, but much better optimization when larger vectors are decomposed into it.
3. Use [`select_unpredictable`](https://doc.rust-lang.org/stable/std/hint/fn.select_unpredictable.html) stabilized in 1.88 in scalar fallbacks of concat_swizzle_dyn_precise and swizzle_dyn_precise.

**Rationale for the branchless formulation:**  `cmov` is far from a silver bullet and is [often a bad idea](https://yarchive.net/comp/linux/cmov.html). In my benchmarks this is 20% to 40% slower than branches for swizzle_dyn_precise if the indices are all out of range, but avoids a 5x (400%) performance degradation on unpredictable indices. If you know your indices are predictable, you don't use `_precise` in the first place, so tuning for unpredictability seems fine. Most importantly, removing branches helps autovectorizer a lot, so we get more reliable autovectorization on e.g. RISC-V, which is the primary use case.

Implementation-wise, NEON and AVX-512 are straightforward with native-ish instructions, SSE4.2 uses [the same decomposition idea](https://shnatsel.github.io/improving-std-simd-swizzle-dyn/#doubling-the-vector-width) as swizzle_dyn_precise, and AVX2 needs babysitting as always.

The first commit is squashed because I did a bunch of experiments with different formulations, e.g. a specialized one for 256-bit AVX2 that reduces shuffle port pressure on haswell and doesn't regress recent AVX2 much if at all, but in the end decided those are not worth the extra complexity.

I'm not thrilled by the `lhs.concat_swizzle_dyn(rhs, indices)` API but I don't see how to do better without it getting really awkward in other ways: `concat_swizzle_dyn(simd, lhs, rhs, indices)` isn't any better and breaks convention, and `indices.concat_swizzle_dyn(lhs, rhs)` runs into issues with matching vector lengths.

### DIFF
--- a/fearless_simd/src/generated/avx2.rs
+++ b/fearless_simd/src/generated/avx2.rs
@@ -1366,8 +1366,7 @@ impl Simd for Avx2 {
             #[inline(always)]
             fn kernel(token: Avx2, a: u8x16<Avx2>, indices: u8x16<Avx2>) -> u8x16<Avx2> {
                 let indices = indices.into();
-                let index_out_of_range = _mm_add_epi8(indices, _mm_set1_epi8(112));
-                let zeroing_indices = _mm_or_si128(indices, index_out_of_range);
+                let zeroing_indices = _mm_adds_epu8(indices, _mm_set1_epi8(0x70));
                 let result = _mm_shuffle_epi8(Bytes::to_bytes(a).val.0, zeroing_indices);
                 let result_bytes = u8x16 {
                     val: crate::support::Aligned128(result),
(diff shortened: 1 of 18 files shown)


Shnatsel:
@Novum please let me know if this looks good to you

Novum:
Looks good for me for those operations

Issue #340, opened 2026-08-19
Feature Request: Add cross-lane horizontal reduction methods (e.g., .reduce_sum(), .reduce_max()) to SimdBase / Simd vector types


Mnwa:
### Feature Description
With `fearless_simd` solidifying its generic traits and racing toward a stable v1.0 release, it would be extremely valuable to have native, cross-lane horizontal reduction operations built directly into the crate's vector APIs. 

Specifically, adding methods such as `.reduce_sum()`, `.reduce_min()`, `.reduce_max()`, and logical reductions like `.any()` / `.all()` on boolean masks.

### Motivation & Use Case
When performing numerical algorithms (like calculating dot products, vector averages, or image processing bounds), the standard optimization pattern relies on accumulating data vertically across independent vector lanes within the loop, followed by a single horizontal reduction to a scalar at the very end.

Currently, `std::simd` supports this directly via `vector.reduce_sum()`. For `fearless_simd` to serve as a complete, drop-in replacement on stable Rust, users shouldn't have to break abstraction or manually write platform-specific shuffles/hadds to extract a single scalar sum from a generic vector type. 

Providing this inside `SimdBase` or via a dedicated reduction trait would significantly improve ergonomics for generic SIMD code.

### Proposed API / Design
Ideally, these operations would be exposed through the recently unified trait architecture (like `SimdBase`).

```rust
pub trait SimdBase {
    type Element;
    // ... existing items

    /// Sums all lanes in the vector horizontally.
    fn reduce_sum(self) -> Self::Element;

    /// Finds the maximum value among all lanes.
    fn reduce_max(self) -> Self::Element;

    /// Finds the minimum value among all lanes.
    fn reduce_min(self) -> Self::Element;
}
```

### Alternatives Considered
1. **Manual Array Reinterpretation**: Extracting the vector to an array via `.to_array()` (or using upcoming `as_array` methods) and using `.iter().sum()`. However, this relies heavily on the compiler to optimize out memory operations, which can sometimes fail to lower cleanly to native hardware horizontal addition instructions (like `HADDPS` on x86 or `ADDV` on AArch64).
2. **Platform-Specific Intrinsics**: Writing custom matching logic per SIMD architecture level (`Sse2`, `Avx2`, `Neon`), which completely defeats the purpose of the beautiful generic abstractions `fearless_simd` provides.

Shnatsel:
> Extracting the vector to an array via `.to_array()` (or using upcoming `as_array` methods)

[...]

PR #357, opened 2026-08-28
Implement reduce_sum()

Shnatsel:
This implements the hardest part of https://github.com/linebender/fearless_simd/issues/340

With floats this is very tricky. Their addition is non-trivial due to accumulating precision loss, and there are many algorithms and many precision trade-offs. See https://orlp.net/blog/taming-float-sums/ for more info.

This PR's implementation splits vectors in half and performs lane-wise addition down to 128-bit vectors, then reduces the 128-bit vector down to a single value with pairwise summation. This is a fast and precise-ish algorithm with log2(N) roundings.

I feel this is a good default because it is portable (bit-exact output everywhere, except for NaN payloads) and pretty fast. This leaves us space to add a `_relaxed` variant if higher-performance options with fewer guarantees are ever discovered, and/or a `_precise` variant for Kahann summation with a single rounding instead of log2(N) roundings.

We handily beat `std::simd` which sums floats one by one left to right, which is awful for both precision and performance: it forces a long chain of scalar additions, and [this is the worst case for precision too](https://orlp.net/blog/taming-float-sums/), causing N roundings for st::simd as opposed to log2(N) in this PR.

Integers are not under any of the same constraints, so I tried looking for faster formulations for integers, but couldn't find anything. This reinforces my hunch that this is about as fast as this op can get, even for floats.

cc @Mnwa who requested this feature

### DIFF
--- a/fearless_simd_gen/src/mk_x86.rs
+++ b/fearless_simd_gen/src/mk_x86.rs
@@ -318,7 +318,13 @@ impl Level for X86 {
             OpSig::Splat => self.handle_splat(op, vec_ty),
             OpSig::Compare => self.handle_compare(op, method, vec_ty),
             OpSig::Unary => self.handle_unary(op, method_sig, method, vec_ty),
-            OpSig::Reduce { lane_op } => self.handle_reduce_min_max(op, vec_ty, lane_op),
+            OpSig::Reduce { lane_op } => {
+                if lane_op == "add" {
+                    self.handle_reduce_sum(op, vec_ty)
+                } else {
+                    self.handle_reduce_min_max(op, vec_ty, lane_op)
+                }
+            }
             OpSig::Widen { target_ty } => self.handle_widen(op, vec_ty, target_ty),
             OpSig::Narrow { target_ty, mode } => self.handle_narrow(op, vec_ty, target_ty, mode),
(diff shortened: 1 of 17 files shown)


fearless_simd_gen/src/mk_x86.rs
```diff
@@ -1127,6 +1128,56 @@ impl X86 {
...
+        assert_eq!(
+            vec_ty.n_bits(),
+            128,
+            "wide reductions must use the generic 128-bit-grained implementation"
+        );
+
+        match (vec_ty.scalar, vec_ty.scalar_bits) {
+            (ScalarType::Float, 32) => self.kernel_method(op, vec_ty, |_| {
+                quote! {
+                    // _mm_hadd_ps is slower than shuffle followed by add
```
LaurenzV, line 1301: So why does it exist in the first place? 😂

LaurenzV (approved):

fearless_simd_gen/src/mk_x86.rs
Shnatsel: Looks like historical reasons to me.

It was added in ssse3 while shuffles appeared in SSE4, and it doesn't need an extra register for a shuffle mask.

PR #358, opened 2026-08-28
Implement reduce_min()/reduce_max() and their precise variants

Shnatsel:
The simpler part of https://github.com/linebender/fearless_simd/issues/340

Implementation decisions largely mirror #357 but we don't have to worry about error accumulation in floats. API design follows the existing lane-wise min/max.

cc @Mnwa who requested this feature


fearless_simd_tests/tests/generics.rs
```diff
@@ -63,6 +63,23 @@ fn generic_first<S: Simd, V: SimdBase<S>>(vector: V) -> V::Element {
     vector[0]
 }
 
+// Ensure that horizontal extrema can be obtained from a generic numeric vector.
```
LaurenzV: Nit but can we replace "horizontal extrema" with: min/max or something else? Sounds a bit weird to me 😄

fearless_simd_tests/tests/harness/ops/reduce_min.rs
```diff
@@ -0,0 +1,238 @@
...
+fn reduce_min_f32x4<S: Simd>(simd: S) {
+    let value = f32x4::from_slice(simd, &[3.0, f32::NEG_INFINITY, -2.0, f32::INFINITY]);
+    assert_eq!(value.reduce_min(), f32::NEG_INFINITY);
+    assert_eq!(
+        f32x4::from_slice(simd, &[0.0, -0.0, 0.0, -0.0]).reduce_min(),
+        0.0
+    );
+}
+
+#[simd_test]
```
LaurenzV, line 19: I think for those, it would once again be better to just spell out the whole vector explicitly? And I think it would be good to also have coverage of other values then just `X::MIN` and `X::MAX`.

LaurenzV: Same for `max`.

LaurenzV (approved):
Why is everything so much easier on NEON compared to AVX2! 😄

Issue #280, opened 2026-07-24
Optimize AVX2 vector shifts once rustc codegen bugs are fixed


Shnatsel:
I've tried optimizing the 16-bit variable-vector shifts for AVX2 and got formulations that should be much better than autovectorization: https://github.com/Shnatsel/fearless_simd/commit/95666a57f5cd1f1f0dbf0c77f4a2e70b3e958ade

However, the actual codegen is poor due to a compiler bug, with intrinsics not lowering into the instructions they're supposed to: https://github.com/rust-lang/rust/issues/159831

We should revisit this once the upstream codegen bugs are fixed. Even if they were fixed today we likely won't get the upstream LLVM fixes till January.

Issue #163, opened 2025-12-12
Stop exposing "shift by vector" via operator overloading?


valadaptive:
The "shift left by vector" and "shift right by vector" operations don't lower to hardware operations on many platforms. On WebAssembly and SSE4.2, they all fall back to scalar operations. On AVX2, they're only vectorized for 32-bit and 64-bit operands.

It seems like a bit of a footgun to expose such operations in a way that looks identical to the much faster "shift by scalar". For example, [`clatter` uses a vectorized shift as part of an RNG](https://github.com/Ralith/clatter/blob/ea2f36ae7ddc32b7ed759d0dd1de2513b2575e52/src/hash.rs#L7), and [switching to a different algorithm that uses a non-vectorized shift is faster in practice](https://github.com/valadaptive/clatter/blob/30a7b47ebcba076a7c9e82221c128f2bb9532039/src/hash.rs#L5-L13).

It also makes it easier to accidentally use a vectorized shift when a scalar shift would do. You'd probably expect the two code snippets below to produce identical code:

```rust
let x: u32x4<S> = y >> 5;
```

```rust
let x: u32x4<S> = y >> u32x4::splat(simd, 5);
```

All the other operations take an `impl SimdInto<Self, S>` as the right-hand side and just call `splat` internally, so it's reasonable to assume that's what happens for the shifts as well. In this case, however, the latter snippet is slower.

Ralith:
Can we define an intuitive, consistent bar for operations which are widely supported enough to merit convenient syntax? I worry that this might become a bit mercurial.

> On AVX2, they're only vectorized for 32-bit and 64-bit operands.

I feel like that's the overwhelmingly common case, which is worth something.

PR #359, opened 2026-08-29
Improve shift documentation

Shnatsel:
See also: https://github.com/linebender/fearless_simd/issues/163

Doesn't document #280 since that's hopefully temporary

### DIFF
--- a/fearless_simd_gen/src/mk_simd_trait.rs
+++ b/fearless_simd_gen/src/mk_simd_trait.rs
@@ -390,8 +390,6 @@ fn mk_simd_base() -> TokenStream {
             /// Shift the vector elements to the left by `OFFSET`, filling in with `padding` from the right.
             ///
             /// If `OFFSET` is greater than or equal to `Self::N`, all lanes are filled with `padding`.
-            ///
-            /// If you don't need to configure the padding, prefer the `<<` operator instead.
             #[inline(always)]
             fn shift_elements_left<const OFFSET: usize>(self, padding: Self::Element) -> Self {
                 match OFFSET.min(Self::N) {
@@ -403,8 +401,6 @@ fn mk_simd_base() -> TokenStream {
             /// Shift the vector elements to the right by `OFFSET`, filling in with `padding` from the left.
             ///
             /// If `OFFSET` is greater than or equal to `Self::N`, all lanes are filled with `padding`.
-            ///
-            /// If you don't need to configure the padding, prefer the `>>` operator instead.
             #[inline(always)]
             fn shift_elements_right<const OFFSET: usize>(self, padding: Self::Element) -> Self {
                 let padding = Self::splat(self.witness(), padding);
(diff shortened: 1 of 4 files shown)


fearless_simd_gen/src/mk_simd_trait.rs
```diff
@@ -401,6 +403,8 @@ fn mk_simd_base() -> TokenStream {
             /// Shift the vector elements to the right by `OFFSET`, filling in with `padding` from the left.
             ///
             /// If `OFFSET` is greater than or equal to `Self::N`, all lanes are filled with `padding`.
+            ///
+            /// If you don't need to configure the padding, prefer the `>>` operator instead.
```
LaurenzV: I don't get this one, `shift` swizzles the elements within a vector, while `>>` shifts each element individually, no?

fearless_simd_gen/src/ops.rs
```diff
@@ -952,7 +952,10 @@ const INT_OPS: &[Op] = &[
         "Shift each element left by the given number of bits.\n\n\
         Bits shifted out of the left side are discarded, and zeros are shifted in on the right.\n\n\
         When shifting out of bounds (e.g. shifting a 32-bit value by 32 or more), the result is implementation-defined and may vary by platform.\n\n\
-        This operation is not implemented in hardware on all platforms. On WebAssembly, and on x86 platforms without AVX2, this will use a fallback scalar implementation.",
+        # Performance\n\
+        This per-lane shift is slower than shifting the entire vector by a scalar. \
+        It is not vectorized at all on WebAssembly, SSE4.2, and for `u8` elements on AVX2. \
```
LaurenzV: and `i8` right?

fearless_simd_gen/src/mk_simd_trait.rs
Shnatsel: Oh yeah my bad, I got confused. Thanks. I'll revert that one.

fearless_simd_gen/src/ops.rs
Shnatsel: Yes.

Okay wow I am not thinking straight today. I'll take a break from programming because this is way too many mistakes in a single simple PR.

LaurenzV (approved):
LGTM! Though it's still an open question whether we should keep the overload (which is what the issue was mainly about I think), any opinion that? I guess people might not discover the doc comment if they just use `<<` directly.

Shnatsel:
I genuinely don't know.

`std::simd` uses `<<` but it's not exactly a model of performance.

SSE4.2 is 15% of x86 in active use, only 8-bit AVX2 will be slow forever, so I guess it's not _that_ bad?

I can't come up with a good name for it either. `shift` is close to `shift_elements` and is easy to confuse. The `safe_arch` crate uses `shl_each_*` but `safe_arch` isn't exactly widely used, and even `wide` built on top doesn't expose those ops at all.

Shnatsel:
We might be able to do better for u8 than outright scalarizing with some clever emulation.

Sadly this optimization work is blocked by multiple compiler bugs:

https://github.com/rust-lang/rust/issues/159831 https://github.com/rust-lang/rust/issues/159464

PR #360, opened 2026-08-29
Implement rotate_elements_{left,right} on masks

Shnatsel:
Pretty straightforward: forward either to the vector `rotate_elements_*` for vector-backed masks, or to `rotate_*` on std scalars for AVX-512.

Since lowerings into `slide()` caused issues with compile time before, I measured how this impacts the build, and it's within measurement noise. I've experimented with implementing it as a default trait without re-emitting basically the same body into individual implementation methods, but that created more IR for LLVM dead code elimination to chew through, and slowed down compilation of code that actually uses and instantiates this without meaningfully changing the frontend costs. So I believe this is the best way to implement this.


fearless_simd_tests/tests/harness/ops/rotate_elements_right.rs
LaurenzV, line 1: Optional but I wonder if there should be 1 test for each with a mask with more than 1 bit set at a time?

LaurenzV (approved):

PR #361, opened 2026-08-30
Implement reduce_product()

Shnatsel:
Matches `std::simd` in precision (same number of roundings), but uses a more SIMD-friendly multiplication order. It uses the same order across all backends. The implementation is similar to `reduce_sum()`.

The result is not guaranteed to be equal to `v.to_slice().sum()`, but that is only achievable with fully scalar multiplication; `std::simd` scalarizes to uphold this behavior which it doesn't even document.

Unlike #357 we can't reduce precision loss to log2(N) rather than N, at least not without complex algorithms that compute the precise sum in one rounding by separately accounting for the accumulated error. A precise variant which does that could be added later.

This is the final part of #340.


fearless_simd_gen/src/ops.rs
```diff
@@ -645,9 +645,22 @@ const COMMON_BASE_OPS: &[Op] = &[
...
+        Because floating-point addition is not associative, separately reducing smaller vectors and then adding their results can differ from reducing their combined wider vector. See [Taming Floating-Point Sums](https://orlp.net/blog/taming-float-sums/) for more information and for other summation algorithms, including exact summation without accumulated rounding error. In that article's terms, our method has the precision properties of pairwise summation, although the exact pairing of values is different.",
+    ),
+    Op::new(
+        "reduce_product",
+        OpKind::BaseTraitMethod,
+        OpSig::Reduce { lane_op: "mul" },
+        "Return the product of all elements in the vector. Integer multiplication wraps.\n\n\
+        # Floating-point behavior\n\n\
+        For a vector with N elements, this operation performs N roundings.\n\n\
```
LaurenzV: `N - 1` roundings, right?

fearless_simd_gen/src/ops.rs
```diff
@@ -645,9 +645,22 @@ const COMMON_BASE_OPS: &[Op] = &[
...
+        For a given vector type and lane count, this operation produces the same result on all platforms and backends down to the bit pattern, except that when the result is NaN, its exact bit pattern is unspecified. \
+        This fixed-width guarantee does not make code using native-width associated types such as `S::f32s` independent of the selected SIMD level, because their lane counts can differ.\n\n\
+        The result of this operation is **not** bit-exact to scalar sum of the elements because it multiplies elements in a different (but fixed) order.\n\n\
```
LaurenzV:
```suggestion
        The result of this operation is **not** bit-exact to scalar product of the elements because it multiplies elements in a different (but fixed) order.\n\n\
```

fearless_simd/src/generated/sse4_2.rs
```diff
@@ -3548,6 +3628,15 @@ impl Simd for Sse4_2 {
...
     }
     #[inline(always)]
+    fn reduce_product_i32x4(self, a: i32x4<Self>) -> i32 {
+        let mul_level_0: [i32; 2usize] = [
+            a[0usize].wrapping_mul(a[1usize]),
+            a[2usize].wrapping_mul(a[3usize]),
+        ];
+        let mul_level_1: [i32; 1usize] = [mul_level_0[0usize].wrapping_mul(mul_level_0[1usize])];
+        mul_level_1[0]
+    }
```
LaurenzV, line 3818: I'm wondering whether the compiler will actually compile this just into a sequence of multiplications or whether there will be loads/stores?

LaurenzV (approved):

fearless_simd/src/generated/sse4_2.rs
Shnatsel: Since all these are local stack values with very clearly no aliasing, the intermediate stores should in theory be eliminated.

The reason it's emitted like this with consecutive arrays is to make autovectorization easier on platforms we don't already support via intrinsics.

fearless_simd_gen/src/ops.rs
Shnatsel: Yes! Good catch. Fixed.

PR #363, opened 2026-08-31
Do not allow subverting `kernel!` soundness by using #[doc(hidden)] items

Shnatsel:
The threat model for Rust is that the code written by the programmer is non-adversarial. So having some `#[doc(hidden)]` items is fine.

However, LLMs still flag this in an audit, and it's easier to tighten things up once and for all than to explain to users that this is actually fine every time.

### DIFF
--- a/fearless_simd/src/kernel_macros.rs
+++ b/fearless_simd/src/kernel_macros.rs
@@ -109,9 +109,7 @@ macro_rules! __fearless_simd_kernel_dispatch {
         $($body:tt)*
     ) => {
         $crate::__fearless_simd_kernel_impl! {
-            @cfg target_arch = "aarch64";
-            @token_ty $crate::Neon;
-            @kernel_attrs #[target_feature(enable = "neon")];
+            @level Neon;
             $($body)*
         }
     };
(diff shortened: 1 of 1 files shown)


DJMcNab:
Could you expand on what hole this is closing? I'm definitely hesitant to have a soundness break unless it's unavoidable, so I'd rather have a fix.

I presume that the type being `doc(hidden)` doesn't have any impact there - I'm not aware of any semantics impact of that attribute.

Shnatsel:
Prior to this PR, `__fearless_simd_kernel_impl!` accepted both the level token and the corresponding target feature string as arguments and trusts them. So one could theoretically use this private-ish, low-level and doc-hidden API (which can't be fully private due to use in macros which instantiate outside this crate) to construct a function that accepts e.g. an SSE2 token but emits an AVX-512 feature string.

Rust is not a sandbox to defend against adversarial code, unlike e.g. JavaScript running in the browser, so if someone went through _this_ much trouble to get a soundness issue, they probably deserve it. But this PR fixes this loophole regardless.

DJMcNab (approved):
Thanks!

PR #367, opened 2026-09-04
Add security policy, post-v1.0 MSRV bump policy

Shnatsel:
Based on the discussion in the relevant Zulip thread: [#simd > v1.0.0](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/v1.2E0.2E0/with/609522501)


fearless_simd/README.md
```diff
@@ -233,8 +233,11 @@ It benefited from conversations with Luca Versari, though he is not responsible
 
 This version of Fearless SIMD has been verified to compile with **Rust 1.89** and later.
 
-Future versions of Fearless SIMD might increase the Rust version requirement.
-It will not be treated as a breaking change and as such can even happen with small patch releases.
+Future versions of Fearless SIMD might increase the Rust version requirement. This will be accompanied by a minor version bump.
+
+MSRV increases do not impact existing users:
+Cargo supports [an MSRV-aware resolver](https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html) and will not select a version that fails to build,
+and we provide [security backports](SECURITY.md) to older Rust versions released within the last 3 years.
```
DJMcNab:
```suggestion
If you require a fixed MSRV, we recommend using Cargo's [MSRV-aware resolver](https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html) which will not select a version that fails to build.
We will provide [security backports](SECURITY.md) to older Rust versions released within the last 3 years.
```

Is my proposal to reword "do not impact existing users".

fearless_simd/SECURITY.md
```diff
@@ -0,0 +1,20 @@
...
+Earlier versions may or may not receive security backports at the discretion of Fearless SIMD maintainers.
+
+You can look up when each Rust version was released in the [Rust Release Notes](https://doc.rust-lang.org/stable/releases.html).
+
+## Reporting a Vulnerability
+
+This repository follows a full disclosure policy. Please report security issues publicly on the Github issue tracker.
+
+Entirely AI-generated reports are not permitted. While the use of LLMs for auditing code is welcome,
+the issue must be reproduced and understood by a human before reporting.
```
DJMcNab: I think this needs a final newline.

DJMcNab (approved):
One question I do have is what counts as a security backport. E.g. if one of our operations doesn't meet the specified semantics, that can be a security issue in a downstream crate (e.g. one doing cryptographic hashing, say). But then silently updating the implementation to have different semantics is a Hyrum's law nightmare.

I'd be tempted to specify "soundness" as the baseline boundary, and leave everything else unspecified. In a scenario where we do discover a method doesn't work as documented, I'd expect us to *probably* backport the deprecation and providing a new fixed method, rather than updating semantics. (Although I wonder how that works for `core::ops` impls!)

Aside from that concern, this looks great. And being overly vague with what counts as a security issue is *fine* - I don't expect this policy to ever come up in practice.

Shnatsel:
@LaurenzV I'd like to hear from you as well before we commit to this.

fearless_simd/README.md
```diff
@@ -233,8 +233,10 @@ It benefited from conversations with Luca Versari, though he is not responsible
 
 This version of Fearless SIMD has been verified to compile with **Rust 1.89** and later.
 
-Future versions of Fearless SIMD might increase the Rust version requirement.
-It will not be treated as a breaking change and as such can even happen with small patch releases.
+Future versions of Fearless SIMD might increase the Rust version requirement. This will be accompanied by a minor version bump.
+
+If you require a fixed MSRV, we recommend using Cargo's [MSRV-aware resolver](https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html) which will not select a version that fails to build.
+We will provide [security backports](SECURITY.md) to older Rust versions released within the last 3 years.
```
LaurenzV, line 239: What's the usual procedure for handling this? Do we create a new branch for each backported version?

LaurenzV (commented):
Overall fine from my side, trusting your judgement on this. 😄

fearless_simd/README.md
Shnatsel: Yes, branch off the git tag for that version. Even if we don't do tags, cargo now includes the git commit state into archive published to crates.io, so we always know where to branch off.

PR #369, opened 2026-09-06
Put SECURITY.md at the top level so that Github recognizes it

Shnatsel:
Turns out Github doesn't pick up on it if it's in a subfolder


Shnatsel:
Hang on, that breaks the link in the README. Let me think about this.

PR #371, opened 2026-09-06
Move abs() from SimdFloat to SimdBase, implement it on integers

Shnatsel:
This is technically API-breaking so I'd like to get it done before v1.0

Implements wrapping semantics, same as `i32::abs()`. We can add a saturating variant later, since it doesn't require API changes.


fearless_simd_gen/src/ops.rs
```diff
@@ -557,6 +557,16 @@ fn splat_arg_ty(vec_ty: &VecType) -> TokenStream {
 }
 
 const BASE_OPS: &[Op] = &[
+    Op::new(
+        "abs",
+        OpKind::BaseTraitMethod,
+        OpSig::Unary,
+        "Compute the absolute value of each element.\n\n\
+        Unsigned integers are unchanged. Signed integers use wrapping absolute value: \
+        the minimum representable value remains unchanged. This matches `i32::abs()`.\n\n\
```
LaurenzV, line 566: Probably `wrapping_abs` would be the better comparison, right?

LaurenzV (approved):

fearless_simd_gen/src/ops.rs
Shnatsel: Since we don't panic in debug mode, yes, that's more accurate. I think it's less clear though.

Shnatsel: I'm not married to it, so I'll change it if you insist.

LaurenzV: Not insisting 😄 as you want

PR #372, opened 2026-09-07
add cargo-semver-checks to CI

Shnatsel:


LaurenzV (approved):
I suppose let's see how it goes!

PR #373, opened 2026-09-08
Ligthen CI load, test on latest stable

Shnatsel:
We've gone over the limit of 20 concurrent jobs, so CI takes much longer than it has to. This gets us back under that limit.

Prior to this PR all jobs ran on v1.89, this uses latest stable for some jobs instead.

Clippy already doubles as 'cargo check', so remove separate 'cargo check' jobs and only run Clippy on MSRV, where we want to keep it anyway to avoid spurious new lints breaking CI.

With Clippy pinned to MSRV, we can safely use the latest toolchain to discover any newly introduced issues. While previously we had all the jobs running on 1.89, now we can cargo check on MSRV and actually test on the latest stable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,4 @@
 env:
-  # We aim to always test with the latest stable Rust toolchain, however we pin to a specific
-  # version like 1.70. Note that we only specify MAJOR.MINOR and not PATCH so that bugfixes still
-  # come automatically. If the version specified here is no longer the latest stable version,
-  # then please feel free to submit a PR that adjusts it along with the potential clippy fixes.
-  RUST_STABLE_VER: "1.89" # In quotes because otherwise (e.g.) 1.70 would be interpreted as 1.7
   # The purpose of checking with the minimum supported Rust toolchain is to detect its staleness.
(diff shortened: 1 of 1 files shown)


Shnatsel:
Since I've removed several jobs from CI, this needs privileged access to merge.

LaurenzV:
@DJMcNab

DJMcNab:
There's two or three things being mixed together here:
- That we should only use clippy on MSRV. I'm not a massive fan of this - new clippy versions do give useful lints. But I guess you're claiming that you will be running `cargo clippy` on a reasonably recent stable anyway, so that meets the same requirement? I'd be more in favour of e.g. combining the cargo test and cargo clippy jobs to get us under the limit. Alternatively, we could drop the non-address sanitizer test runs, as well as combine typos, check generated code and formatting into a single formatting job.
- Instead of testing on a pinned stable, we should use latest stable. It's a defensible proposition; if we do take the first item. Clippy is the main reason to want to be more intentional here, and so the risk of a.
- Updating the pinned stable version is not being done, so we're not testing on stable. The intended response there is to update the pinned stable version, not just drop the pinned version entirely.

I'm not likely to approve this myself, but it wouldn't be disastrous to apply. Once it's approved and you want to merge it, let me know and I'll update the protection rules.

PR #378, opened 2026-09-15
Add a way to select the native SIMD vector width for a given scalar type

Shnatsel:
Adds a way to process `&[f32]` and `&[f64]` in the same generic function using native-width vectors. Also adds an example demonstrating it.

This came up as a missing piece during experimentation with using fearless_simd for HPC and reported to me privately.

This is not semver-breaking, so we don't have to include it in v1.0, even though it's nice to have.


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ This release has an [MSRV][] of 1.89.
 
 ### Added
 
+- Added `Native<S>` to `SimdFloatElement` and `SimdIntElement` to select native-width vectors in scalar-generic code. The new `sigmoid_generic` example demonstrates dispatch and partial-vector handling for both `f32` and `f64`.
```
LaurenzV:
```suggestion
- Added `Native<S>` to `SimdFloatElement` and `SimdIntElement` to select native-width vectors in scalar-generic code. 
```

Last sentence isn't changelog-worthy I think.

LaurenzV (approved):

PR #383, opened 2026-09-19
Allow the first argument of #[simd] functions to be `f32x4<S>` and such

### LINKED ISSUE #379: `#[simd]` should work on functions without an explicit SIMD token parameter
Right now (#347), `#[simd]` requires the first non-receiver argument to a SIMD token. However, it's not uncommon for SIMD functions to not have an explicit SIMD token parameter, because any SIMD type (e.g. `f32xN`, etc) already carries a SIMD token with them. For example, this is what the `sRGB` example does:

https://github.com/linebender/fearless_simd/blob/bd9f7c7be6b4a2c009558fffd5db67c54ff744f5/fearless_simd/examples/srgb.rs#L52-L67

Such functions should be supported to make the `#[simd]` macro easier to use.

## Suggested solution

Instead of requiring that the first parameter ***is*** a SIMD token, require that the first parameter ***carries*** a SIMD token.

This could be done by adding a new trait like so:

```rs
pub trait ExtractToken { // bikeshed names
    type S: Simd;
    fn simd(&self) -> Self::S;
}

// all SIMD token types return self
impl ExtractToken for Avx2 {
    type S = Self;
    fn simd(&self) -> Self::S {
        *self
    }
}

// all data type return their SIMD token
impl<S: Simd> ExtractToken for f32x8<S> {
    type S = S;
    fn simd(&self) -> Self::S {
        self.simd
    }
}
```

and change the code gen of the `#[simd]` macro to use this trait to get a SIMD token from the first parameter.

By adding [`#[diagnostic::on_unimplemented]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute) on the trait, we can also get nice compiler errors when users use `#[simd]` incorrectly. E.g.

```rs
#[simd]
fn foo<S: Simd>(a: i32);
```
```
error[E0277]: `i32` does not carry a SIMD token
   --> src\encode\bc4.rs:961:24
    |
961 |     ExtractToken::simd(&a)
    |     ------------------ ^^ the trait `fearless_simd::ExtractToken` is not implemented for `i32`
    |     |
    |     required by a bound introduced by this call
    |
    = note: If you are using the #[simd] macro, the first parameter must carry a SIMD token. See the docs of #[simd] for more information.
help: the following other types implement trait `fearless_simd::ExtractToken`
   --> ...
```

---

Note: I think the `ExtractToken` trait should be part of the public API of the main crate. I want to implement it on the newtype wrappers around `fearless_simd` types I use.

Shnatsel:
Allow the first argument in `#[simd]` functions to be anything simd-token-carrying, not just a literal `_: Simd`

Heavily based on the ideas @RunDevelopment outlined in #379

Closes #379

I want to get it in before v1.0 because this is a small semver-breaking change (method moved from one trait to another)

@RunDevelopment please take a look


Dr-Emann:
I know it would be quite the huge change, but the naming feels inconsistent throughout between Simd vs token vs witness. Is there a distinction I'm missing, or could we call them all the same thing? (Simd, ExtractSimd::simd)

Shnatsel:
I think `Simd` would have been `SimdLevel` if you didn't have to keep repeating it on every function. 

I'm open to `ExtractToken::token` to make the naming more consistent.

Dr-Emann:
> I think `Simd` would have been `SimdLevel` if you didn't have to keep repeating it on every function.

Or maybe SimdToken, based on this pr.

I still think vector.simd() better matches, especially with vector.simd already existing. vector.token() is less obvious to me (but better than vector.witness())

Shnatsel:
Renamed. Any concerns other than naming?

Dr-Emann (approved):
I still say .simd() would be better

Dr-Emann:
Either way, definitely a breaking change to note for 1.0

Shnatsel:
This PR adds a mention to the changelog: https://github.com/linebender/fearless_simd/pull/383/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

Dr-Emann:
Saw the added section, missed the changed section somehow, awesome!

LaurenzV (approved):

PR #387, opened 2026-09-20
Reword the advice on the proc macro usage

Shnatsel:
following the feedback at [#simd > v1.0.0 @ 💬](https://xi.zulipchat.com/#narrow/channel/514230-simd/topic/v1.2E0.2E0/near/625490380)


fearless_simd/src/lib.rs
```diff
@@ -110,9 +110,7 @@
 //!
 //! ## The `#[simd]` annotation
 //!
-//! Fearless SIMD requires functions that use SIMD to be annotated with the `#[simd]` attribute from the `fearless_simd_macros` crate. Without it the code will still compile, but run far slower, defeating the purpose of SIMD.
-//!
-//! Use [`dispatch!`] when calling SIMD code from non-SIMD code.
+//! Fearless SIMD requires functions that use SIMD to be annotated with the `#[simd]` attribute from the `fearless_simd_macros` crate. Without it the code will still compile, but may run much slower.
```
Ralith: I don't think "requires" is justified, and the following paragraph immediately contradicts it. Perhaps "recommends"? Similarly, "may run much slower" could be reworded to something like "requires special care to achieve full performance"?

LaurenzV (approved):
LGTM once @Ralith's suggestions are applied.

